PR #1730, opened 2014-05-07
WebSearch: date after, date before bug fix

fpoli:
- Fixes unit tests;
- Fixes method '_convert_spires_date_after_to_invenio_span_query';
- Fixes method '_convert_spires_date_before_to_invenio_span_query';
- Fixes regexp used by '_convert_all_dates';
- Fixes indentation for kwalitee.

Signed-off-by: Federico Poli federico.poli@cern.ch


jirikuncar:
@jalavik it might worth moving to `invenio-query-parser`.

jalavik:
@jirikuncar Yeah, I guess this is addressing the inclusive issue by adding `AND NOT value`, so we could add similar technique to spires contrib unless its done on the Invenio query. 

As you know, this particular PR only has practical effect on legacy, so we could merge it there for `1.3` and then take the concepts with us to the `invenio-query-parser`.

PR #1735, opened 2014-05-13
bibtask: Makes AUTH optional for a list of valid processes in bibshced

frecar:
- promote CFG_BIBSCHED_VALID_PROCESSES_NO_AUTH_NEEDED to global config
- promote CFG_BIBSCHED_TASK_IS_NOT_A_DEAMON to global config

Signed-off-by: Fredrik Nygård Carlsen me@frecar.no


frecar:
Should we close this @tiborsimko ?

tiborsimko:
This patch would create simple string variables in `config.py` which would not work... One would have to amend `inveniocfg.py` in order to create Pythonic lists, not strings, for these two variables.

CC @blixhavn who signed this PR as a reviewer.  Can you please make the necessary changes if you are using this in production at TIND?

PR #1751, opened 2014-05-28
Websearch: fix results when applying search limits

### LINKED ISSUE #1750: Search limits are not correctly applied in the search engine
Function prs_apply_search_limits in the search engine should be updating the results_final dictionary it receives from function prs_intersect_with_colls_and_apply_search_limits. Instead, it re-defines it locally, and consequently, although correctly calculated, the search limits are not correctly applied do the final results.

kasioumis:
* Fixes prs_apply_search_limits so that the results_final dictionary is
  updated, rather than re-defined locally.

(closes #1750)

Signed-off-by: Nikolaos Kasioumis <nikolaos.kasioumis@cern.ch>

### DIFF
--- a/modules/websearch/lib/search_engine.py
+++ b/modules/websearch/lib/search_engine.py
@@ -6184,13 +6184,18 @@ def prs_apply_search_limits(results_final, kwargs=None, req=None, of=None, cc=No
         if verbose and of.startswith("h"):
             write_warning("Search stage 5: applying time etc limits, from %s until %s..." % (datetext1, datetext2), req=req)
         try:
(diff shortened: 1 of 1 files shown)


tiborsimko:
Looks good; I'll add an extra regression test case to your branch in order to make sure that this error does not reappear.

(BTW please note it is "Co-authored-by", not "Co-Authored-By".)

tiborsimko:
Thanks, amended as mentioned, merged as 13d3840146bdb9be52885ec91af357a017ba17be, and added 5cdb435a90bdbfcddc1be7f504528dee5614a898.

PR #1874, opened 2014-06-27
WebMessage: alert users by email

### LINKED ISSUE #1049: WebMessage: alert users also by email
Originally on 2012-05-09

It would be useful if WebMessage could also send email notifications to users email addresses when they receive a new web message.

makistsantekidis:
- Added the feature to receive webmessage notification via email,
  when a new webmessage is sent, both on global level (with the
  configuration variable CFG_WEBMESSAGE_ENABLE_MAIL_NOTIFICATION)
  and on a per user level with the user preference key
  webmessage_mail_notification. (closes #1049)

Signed-off-by: Avraam Tsantekidis makistsantekidis@gmail.com

### DIFF
--- a/config/invenio.conf
+++ b/config/invenio.conf
@@ -1603,6 +1603,10 @@ CFG_WEBMESSAGE_MAX_NB_OF_MESSAGES = 30
 ## we delete orphaned messages?
 CFG_WEBMESSAGE_DAYS_BEFORE_DELETE_ORPHANS = 60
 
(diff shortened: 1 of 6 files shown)


tiborsimko:
@kasioumis @egabancho Is this in works for CDS already?  Wanted milestone?

kasioumis:
@tiborsimko This is not deployed yet and not in works either. It should be trivial however so we should push for it. I think v1.3 milestone is fair for now.

PR #1952, opened 2014-07-15
WIP BibIndex: bibauthorid performance improvements

kaplun:
* Re-implements get_author_canonical_ids_for_recid() to bypass
  complex BibAuthorID API usage and directly retrieve data
  from the DB for a gain of speed of ~800x.

Co-authored-by: Alessio Deiana <alessio.deiana@cern.ch>
Co-authored-by: Samuele Carli <samuele.carli@cern.ch>

### DIFF
--- a/modules/bibauthorid/lib/bibauthorid_dbinterface.py
+++ b/modules/bibauthorid/lib/bibauthorid_dbinterface.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2011, 2012, 2013 CERN.
+## Copyright (C) 2011, 2012, 2013, 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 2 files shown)


tiborsimko:
Note that this change makes the module to peek into another module's internal tables, which is not very API friendly.  Can the performance improvement be done inside the other module, e.g. define a new function there if needed, and import it in the first module?  Considering the BibAuthorID's future, do we care?

kaplun:
> Note that this change makes the module to peek into another module's internal tables, which is not very API friendly. Can the performance improvement be done inside the other module, e.g. define a new function there if needed, and import it in the first module? 

Yes, of course. Albeit it would anyway make bibindex to hardcode a dependency on bibauthorid anyway, so no solution on that side...

>Considering the BibAuthorID's future, do we care?

The performance improvement is of the order of several order of magnitude, so for bibauthorid users (i.e. only INSPIRE) we do care. In general, I don't know.

This is better implemented in #2021 and in particular in https://github.com/fpoli/invenio/commit/0bb30382eddee9d90b5890c43f080e2e3ccf5c4c#diff-8fb9426fbd7e9a5a6fe688d49dab7957R288 

So probably this commit is superseeded.

tiborsimko:
> Albeit it would anyway make bibindex to hardcode a dependency on bibauthorid anyway, so no solution on that side...

Not really; if BibIndex imports and uses some API function that BibAuthorID exposes, then we are good. Kind of like now:

```console
$ git grep bibauthorid modules/bibindex/lib/
modules/bibindex/lib/bibindex_engine.py:    from invenio.bibauthorid_dbinterface import get_data_of_papers
modules/bibindex/lib/bibindex_engine.py:        from invenio.bibauthorid_personid_maintenance import get_recids_affected_since
```

If these API functions and/or their usage is slow, then we'd better either speed them up, or import another, faster API alternative that does the job.  Both would be preferable than accessing BibAuthorID tables directly in BibIndex.

(Basically, I'm just calling for moving this PR's SQL snippet code from BibIndex into some BibAuthorID API and we should be all set.)

> This is better implemented in #2021

That PR is for v1.x only, so (1) if you'd like to merge this for v1.2 already, please move the SQL code as suggested above; if not, (2) please close this PR.

kaplun:
OK, since it would be a pity to miss this optimization I have refactored as requested :)

tiborsimko:
Thanks, commit log message amended and merged as fa482cd58794a90f5ea7dcafbddb28bebf5a7bf2.

Issue #1990, opened 2014-07-23
records: expose past revisions


tiborsimko:
It would be useful to expose past revisions of records over the web.

Upon update of records, the past versions are all stored in the record
history table (`hstRECORD`).  The `bibedit` CLI offers functions to
inspect past versions, for example:

```
$ sudo -u www-data /opt/invenio/bin/bibedit --list-revisions 10
10.20140709180731
10.20140709180059
```

The record revision ID is of the form `RECID.YYYYMMHHhhmmss` where
`YYYY...ss` corresponds to the timestamp when the record was updated.
(Note that this equals to MARC tag `005`.)

Currently, the past record versions are exposed via the
above-mentioned `bibedit` CLI to the administrators, or in the record
editor UI panel to the cataloguers.  They are not easily exposed to
the regular end users.  It would be useful to expose past record
versions also via regular web interface, for example:

```
/record/18/export/xm?revision=REVID
/record/18?of=xm&revision=REVID
```

Task 1: implement the above `revision` argument to expose past master
formats in various record-accessing URLs and APIs, notably
`/record/NNN?of=XXX`.  This will provide 100% complete support for
exposing past historical versions of master formats.

There are several challenges here beyond implementing basic
functionality, notably regarding master-vs-derived formats and
regarding non-MARC master formats.

For example, let us say that BibTeX is an important output format for
a given Invenio installation.  Exposing past record versions in BibTeX
can be done in basically two ways: (a) take an old MARCXML master
format version and run it through the "latest greatest" BibTeX
conversion procedure using current conversion rules, as for the
current MARCXML master format; (b) take old MARCXML version and old
conversion rules, IOW take the old BibTeX format as it had been in
that past moment, and expose precisely that.  The former method has an
advantage that the output would be always "up-to-date" with current
BibTeX rules of the given installation, even in case the
MARCXML->BibTeX processing rules change, say when a new field is
added.  However, with the latter method the output would be "truer" to
what it has been at the given point in time of that old revision.

To use an analogy, one could say that:

```
incoming master format + transformation templates -> outgoing output format
```

where:

```
  marcxml + xslt -> dublincore
  marcxml + bibformat elements -> bibtex
  [...]
```

The former approach would call for using historical MARCXML version
with latest XSLT templates or bibformat elements, while the latter
approach would call for using even historical versions of templates
and bibformat elements.  (Which is not practical to do, since these
could use KB lookups or SQL lookups that are not preserved; hence it
[...]

PR #2042, opened 2014-08-05
records: exposing record metadata revisions

bogdan-kulynych:
(addresses #1990)

* Extends `get_record` with optional `revision` parameter
* Exposes revision for record information (metadata) tab
* Adds a config variable CFG_RECORD_REVISION_URL_PARAMETER
  with default value of "revision"

Signed-off-by: Bogdan Kulynych <bogdan.kulynych@gmail.com>


lnielsen:
@tiborsimko Invenio code base have moved quite far since this PR was opened, and given the new rewrite of data model and search I think it will be likely hard to reintegrate this, so would it be ok to close the PR as someday so that we can still look up the code if we need it in the future.

egabancho:
Related with #3129

tiborsimko:
Another issue is the relation to forthcoming DIP store that will allow to archive more historical formats than just JSON or master ones.  (E.g. slave BibTeX format, see discussion in #1990). Hence some parts of this PR can be recycled now in @egabancho's branch, and wider user-serving parts could be revived once we start with the DIP store.

jirikuncar:
Implementation should be adapted in new *Invenio-Records* module (https://github.com/inveniosoftware/invenio-records/issues/10).

PR #2047, opened 2014-08-06
general: invenio.css and invenio-ie7.css removal

### LINKED ISSUE #2040: general: old, probably not used invenio.css
Beside invenio.css bundle, there is invenio.css stylesheet. I have no idea if there is anything using it right now, but it doesn't look like needed.

The path is `invenio/base/static/css/invenio.css`

jirikuncar:
* Removes unused legacy cascade styles.  (closes #2040)

Reported-by: Kamil Neczaj <kamil.neczaj@cern.ch>


lnielsen:
Time to merge?

jirikuncar:
I'm up for. Sometimes it's better to break things and then fix them if needed. @jmartinm are you using any global styles in *BibEdit*? Can you check what styles needs to be moved to `editor/record.css`?

tiborsimko:
> Time to merge?

Now that we have started separating components out, we can either (1) start removing these module-by-module, or (2) merge this indeed and let people handle missing styles afterwards.  The latter approach seems more promising provided we have buy-in to act rapidly from everyone using some legacy modules still...

> @jmartinm are you using any global styles in BibEdit? Can you check what styles needs to be moved to editor/record.css?

Assigning to @jmartinm.

jmartinm:
The styles for the editor are in `editor/base.css`

Should be safe to remove `css/invenio.css`. That file does not seem to be used at the moment.

PR #2085, opened 2014-08-14
BibUpload: linting

Osso:
Pep8 and other linting in bibupload.py


jirikuncar:
@tiborsimko can we integrate it?

tiborsimko:
If no other branches are built on top of this, then I'd say let's "wonfix" as per Invenio 3.

PR #2096, opened 2014-08-15
OAIHarvest: Fix for call_authorlist_extract

switowski:
* Fixes the error in string formatting cause by the fact that
  the ticketid returned from create_authorlist_ticket() is not int.


tiborsimko:
Thanks, amended commit log message, added Jan and yours truly as reviewers, and merged as 3744d6df47902b660ba0863d056cbc836b76d034.

Issue #1604, opened 2014-04-30
BibUpload: adjust creation_date based on incoming 005


tiborsimko:
Originally on 2013-10-15

When uploading records with known 005 tag, it would be good if the system set bibrec.creation_date accordingly.  This feature request stemmed from #1595.

tiborsimko:
> I believe in these context the creation date shouldn't be considered. Rather only the 005 timestamp upon inserting of brand new record.

Yes, that was my main motivation as well, so that we are able to set `creation_date` to some point in the past, if we are uploading some old records.

PR #2123, opened 2014-08-19
WIP BibUpload: bibrec.earliest_date

kaplun:
Introduces support for `bibrec.earliest_date` consequently allowing to independently sorting by earlieast date (i.e. the time when a publication appeared) and creation date (i.e. when the corresponding record was created in the system).

### DIFF
--- a/modules/bibsort/etc/bibsort.cfg
+++ b/modules/bibsort/etc/bibsort.cfg
@@ -17,25 +17,30 @@
 
 [sort_field_1]
 name = latest first
-definition = BIBREC: creation_date
+definition = BIBREC: id
 washer = NOOP
 
 [sort_field_2]
+name = earliestdate
+definition = BIBREC: earliest_date
+washer = NOOP
+
+[sort_field_3]
 name = title
 definition = FIELD: title
 washer = sort_alphanumerically_remove_leading_articles:fr
 
-[sort_field_3]
+[sort_field_4]
 name = author
 definition = FIELD: firstauthor
 washer = sort_case_insensitive_strip_accents
 
-[sort_field_4]
+[sort_field_5]
 name = report number
 definition = MARC: 037__a,909C0r,088__a
 washer = sort_case_insensitive
 
-[sort_field_5]
+[sort_field_6]
 name = most cited
 definition = RNK: citation
 washer = NOOP
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -261,6 +261,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
             write_message("   -Check if the xml marc file is already in the database: DONE" , verbose=2)
 
     record_deleted_p = False
+    original_005 = None
     if opt_mode == 'insert' or \
     (opt_mode == 'replace_or_insert') and rec_id is None:
         insert_mode_p = True
@@ -288,6 +289,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
                 error = None
         else:
             write_message("   Note: 005 already existing upon inserting of new record. Keeping it.", verbose=2)
+        original_005 = time.strftime("%Y-%m-%d %H:%M:%S", time.strptime(record['005'][0][3].split('.')[0], "%Y%m%d%H%M%S"))
 
     elif opt_mode != 'insert':
         insert_mode_p = False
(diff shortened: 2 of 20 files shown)


modules/bibsort/etc/bibsort.cfg
```diff
@@ -17,25 +17,30 @@
 
 [sort_field_1]
 name = latest first
-definition = BIBREC: creation_date
+definition = BIBREC: id
 washer = NOOP
```
tiborsimko: This change does not seem good.  One of the primary use cases of BibSort was to enable "real" sorting by latest records as first.  Hence we should sort by the time records were created, not by their IDs that may be forced.  E.g. imagine filling new records into record ID hole; we still want to show them as new, depending on their real date.

modules/webstyle/lib/webstyle_templates.py
```diff
@@ -939,7 +945,7 @@ def detailed_record_container_bottom(self, recid, tabs, ln=CFG_SITE_LANG,
     <br/>
     """ % {'similar' : similar,
            'dates' : creationdate and '<div class="recordlastmodifiedbox" style="position:relative;margin-left:1px">&nbsp;%(dates)s</div>' % {
-                'dates': _("Record created %(x_date_creation)s, last modified %(x_date_modification)s") % \
+                'dates': _("Record added %(x_date_creation)s, last modified %(x_date_modification)s") % \
                 {'x_date_creation': creationdate,
```
tiborsimko: Some nomenclature musings.  If we have "datecreated", "datemodified", and "earliestdate" dates, then I'm afraid things may not be fully clear what the latter means, based on its sole name.  Partially perhaps because one can say "earliest modification date" which is not the same as the creation date, or "earliest citation date" which would be the date of the first citation, or "earliest publication date" in case a work was re-published, or "earliest X date" in general.  Seems like the qualifier X is missing, as it were.

It may be worth pondering whether we shouldn't find better name.  And, perhaps, whether we actually *need* a new DB column for storing it?  Wouldn't it be sufficient to have the above BibCheck rule running periodically to force some designed new metadata field, say 260/269 or something, to be equal to the earliest publication date?  For example, do we want to expose this information to cataloguers, or is it just a "derived field" living in the database for convenience?  Are we getting more derived fields of this kind, such as "latest refextract date", "latest keywording date", as we mused elsewhere could be convenient for internal processing?

modules/bibupload/lib/bibupload.py
```diff
@@ -2274,11 +2276,11 @@ def _process_document_moreinfos(more_infos, docname, version, docformat, mode):
...
 
-def update_bibrec_date(now, bibrec_id, insert_mode_p, pretend=False):
+def update_bibrec_date(now, bibrec_id, insert_mode_p, original_005=None, pretend=False):
     """Update the date of the record in bibrec table """
     if insert_mode_p:
-        query = """UPDATE bibrec SET creation_date=%s, modification_date=%s WHERE id=%s"""
-        params = (now, now, bibrec_id)
+        query = """UPDATE bibrec SET creation_date=%s, modification_date=%s, earliest_date=%s WHERE id=%s"""
+        params = (original_005, now, now, bibrec_id)
     else:
```
tiborsimko: Great, #1604 is solved in this way.  I fully agree on doing it only in the insert mode.

modules/bibsort/etc/bibsort.cfg
kaplun: Correct. Indeed mine was only a micro-optimization that we can very well skip (especially in the situation you describe and also in the context of the enforced 005).

modules/webstyle/lib/webstyle_templates.py
kaplun: Indeed nope, the proposed functionality is better implemented using a calculated field (in bibfield) or, for master, using a MARC field populated via BibCheck.

So after all I am not convinced this whole branch should really go into master. We could keep it as an OPS branch within INSPIRE as we are going to probably migrate to pu rather than upgrade to any master version in the near future.

Some of the commits (the 005 ones) though might be nice to preserve.

tiborsimko: > So after all I am not convinced this whole branch should really go into master. We could keep it as an OPS branch within INSPIRE as we are going to probably migrate to pu rather than upgrade to any master version in the near future.

:+1: 

> Some of the commits (the 005 ones) though might be nice to preserve.

:+1: 

Can you please amend the branch to remove the "earliest_date" parts and keep only 005 parts, so that #1604 would be closed?  If people can populate `bibrec.creation_date` for batch uploads in this way, then the extra earliest date column indeed does not seem necessary...

tiborsimko: > amend the branch

(To clarify, I rather meant to create a new branch/PR, keeping only 005 parts out of this branch/PR; then we could close this PR.)

tiborsimko:
Closing for someday, but 005 related parts could be extracted and merged to v1.2/v1.3.

Issue #1511, opened 2014-04-30
OAIRepository: restricted records support


kaplun:
Originally on 2013-05-15

So far, the OAI-PMH server in Invenio automatically removes records belonging to any restricted collection from the resulting set.

It is however desirable to have a smarter implementation that treats OAI-PMH harvesters has real users (e.g. provided with a real IP address that could checked against FireRole rules, or better authenticated as real users, using web api keys).

aw-bib:
@jirikuncar does https://github.com/inveniosoftware/invenio/commit/6aec1df09dbd971350ac28daf5d83abc29672c4a also fix this issue?

jirikuncar:
Yes

lnielsen:
We haven't tested if it actually fixes the problem. Also I don't think it's useful as it is now, since you'll need the OAI harvester to log into Invenio and maintain the session (i.e. like a browser would do).

kaplun:
I think this has become for `master`?

lnielsen:
Moved to https://github.com/inveniosoftware/flask-oaiserver/issues/1

PR #2145, opened 2014-08-25
WebSearch: restricted records in latest additions

### LINKED ISSUE #1315: Bug in latest addition WRT restricted records
Originally

Apparently, records that belongs to both a public and restricted collection are listed in the latest addition view of the public collection also to guests users, while they are not searchable.

kaplun:
- When a record belongs to both a public and a restricted collection,
  no longer display it in the list of latest additions of the
  public collection (in accordance to the outcome of an empty search query
  over the public collection).
  (closes #1315)
- Amend documentation of CFG_WEBSEARCH_VIEWRESTRCOLL_POLICY
  WRT default policy value.

### DIFF
--- a/modules/websearch/lib/websearch_webcoll.py
+++ b/modules/websearch/lib/websearch_webcoll.py
@@ -327,17 +327,20 @@ class Collection:
         # close file:
         f.close()
 
-    def update_webpage_cache(self, lang):
+    def update_webpage_cache(self, lang, all_restricted_recids=None):
         """Create collection page header, navtrail, body (including left and right stripes) and footer, and
            call write_cache_file() afterwards to update the collection webpage cache."""
+
+        if all_restricted_recids is None:
+            all_restricted_recids = intbitset()
 
         ## precalculate latest additions for non-aggregate
         ## collections (the info is ln and as independent)
         if self.dbquery:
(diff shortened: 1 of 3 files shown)


modules/websearch/lib/websearch_webcoll.py
```diff
@@ -1166,29 +1174,43 @@ def task_run_core():
...
                     for lang in CFG_SITE_LANGS:
-                        coll.update_webpage_cache(lang)
+                        if coll.get_name() in restricted_collection_cache.cache:
+                            ## We don't strip away restricted records as the collection
+                            ## is anyway restricted and hence the user must be
+                            ## already authenticated to see records.
+                            ## Additionally, by default, since CFG_WEBSEARCH_VIEWRESTRCOLL_POLICY
+                            ## is set to ANY, any record belonging to a restricted
+                            ## collection is entitled to be displayed in the
+                            ## latest additions.
```
tiborsimko: @kaplun Have you checked this case since?

kaplun: Actually nope, but I can give it one more look

aw-bib: @kaplun it might be that the fix/handling this correctly is tightly related to issue #1511 which actually would be quite important to solve.

PR #2148, opened 2014-08-25
WIP general: lxml none value fix for websearch external collections parser

kaplun:


modules/bibconvert/lib/bibconvert_xslt_engine.py
```diff
@@ -66,6 +67,10 @@
 try:
     # lxml
     from lxml import etree
+    lxml_version = etree.LXML_VERSION
+    if lxml_version < (2, 0, 11, 0):
+        raise ImportError("Minimum lxml version supported is 2.0.11.0, available is %s" % \
+                          ('.'.join([str(v) for v in lxml_version]),))
```
tiborsimko: @kaplun @kasioumis Any progress with moving requirement check?  Shouldn't we close #2275 since we started to discuss here?

kasioumis: @tiborsimko @kaplun No progress has been made on my side. Let's indeed close https://github.com/inveniosoftware/invenio/pull/2275. As far as the version check goes, It seems more logical to leave the version check for `configure-tests.py`.

tiborsimko: > It seems more logical to leave the version check for configure-tests.py.

Yes, that's what I had [suggested previously](https://github.com/inveniosoftware/invenio/pull/2148#discussion_r16703261).  Can you or @kaplun amend one of the two PRs accordingly?

PR #2150, opened 2014-08-26
BibRank: selfcites fix

Osso:
hash() are not stable across runs so I am using a different way to associate ids.

### DIFF
--- a/modules/bibrank/lib/bibrank_selfcites_indexer.py
+++ b/modules/bibrank/lib/bibrank_selfcites_indexer.py
@@ -102,11 +102,21 @@ def get_authors_from_record(recID, tags,
     if use_bibauthorid:
         authors = get_personids_from_record(recID)
     else:
-        authors_list = chain(
-             get_fieldvalues(recID, tags['first_author']),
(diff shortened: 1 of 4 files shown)


kaplun:
All the bibrank citation infrastructure as it is in legacy is working basically only within INSPIRE, which is anyway having its production living on its prod branch.

egabancho:
It would be nice if somebody tries to contribute back from your production branch :wink:

kaplun:
Does it really? We have to reimplement most of the code for the new master and it will very likely go into the overlay as it is way too specific for the domain.

In CDS the citation machinery shouldn't in principle run as it makes sense only on a subject-based digital library, not in an institutional repository.

egabancho:
I wasn't talking specifically about this branch but in general, I can imagine commits things like https://github.com/inspirehep/invenio/commit/d1308a5e26c3363cc10ad1efdfd26b3722a416c0 or https://github.com/inspirehep/invenio/commit/de7620abea267bf255d3e41bdcfaf1cb215dd47d would be welcomed by everyone :wink:

kaplun:
Mmh.. make sense... :-D

PR #2207, opened 2014-09-03
WebSearch: blacklist formats from search

### LINKED ISSUE #1712: of=recstruct not handled in search engine
Originally by *[adeiana (@Osso)](https://github.com/Osso)* on 2014-02-24

Currently situation:
We try to add a dictionary to a string and we throw a server error when of=recstruct

I think pretty much anything else is better.
I propose this solution:
1) We should modify the search engine to return the json representation of the recstruct objects.
2) We should modify the print_record() to return a filtered recstruct. That is needed because some fields are protected and should not exposed to the user.

kaplun:
* Introduces a new configuration variable CFG_WEBSEARCH_BLACKLISTED_FORMATS
  It lists all the invenio internal formats that should not be exposed by
  perform_request_search.
  (closes #1712)

Reviewed-by: Samuele Kaplun <samuele.kaplun@cern.ch>
Signed-off-by: Alessio Deiana <alessio.deiana@cern.ch>

### DIFF
--- a/.kwalitee.yml
+++ b/.kwalitee.yml
@@ -47,6 +47,7 @@ components:
 - BibSword
 - DocExtract
 - ElmSubmit
+- I18N
 - OAIHarvest
 - OAIRepositoy
 - PdfChecker
--- a/INSTALL
+++ b/INSTALL
@@ -850,7 +850,7 @@ Contents
            you may want to create the above symlink in order to
            configure completion for the inveniocfg command.
 
-Good luck and thanks for choosing Invenio.
+Happy hacking and thanks for flying Invenio.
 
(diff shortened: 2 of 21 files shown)


kaplun:
I repushed with the inclusion of the internal config variable. However the PR was made against `master`, so it looks like it doesn't merge. Also I don't have a way to test it at the moment. @tiborsimko you can have a look to see if it doesn't break. :smile:

tiborsimko:
Thanks, I've amended the commit log message so that the item will appear in release notes and I've cherry-picked into `legacy` as 88a5bcb38c4c5cbf5e71af0cd3ed8380cebcbe91.

PR #2259, opened 2014-09-09
BibIndex: tag.recjsonvalue NOT NULL

### LINKED ISSUE #1947: BibIndex: safer NOT NULL `tag.recjson_value`
Commit e0f2ea91 introduces the new `tag.recjson_value` column, with default value `NULL`.

However, according to Python code, default value should be `''`, and, additionally an upgrade recipe should be introduced.

kaplun:
* Complements upgrade recipe invenio_2014_08_13_tag_recjsonvalue by
  adding the further property NOT NULL to the tag.recjsonvalue column
  definition.
  (closes #1947)

### DIFF
--- a/modules/miscutil/lib/upgrades/invenio_2014_09_09_tag_recjsonvalue_not_null.py
+++ b/modules/miscutil/lib/upgrades/invenio_2014_09_09_tag_recjsonvalue_not_null.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2014 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


tiborsimko:
FYI I'm changing milestone to v1.2.0, since most people upgrading from v1.1.5 would be bitten by this...

modules/miscutil/lib/upgrades/invenio_2014_09_09_tag_recjsonvalue_not_null.py
```diff
@@ -0,0 +1,40 @@
...
+
+def info():
+    """Upgrade recipe information."""
+    return "tag.recjson_value NOT NULL"
+
+
+def do_upgrade():
+    """Upgrade recipe procedure."""
+    run_sql(
+        "ALTER TABLE tag CHANGE COLUMN recjson_value recjson_value text NOT NULL;")
```
tiborsimko: I'll amend this recipe during merge to also do:

```
update tag set recjson_value='' where recjson_value is null;
```

otherwise people upgrading from v1.1.5 might be bitten by this...

tiborsimko: It is not necessary to import `warnings` and `wait_for_user`, since they are not used.  I'll remove them during merge to keep the code kwalitee.

tiborsimko:
Thanks, amended as discussed, and merged as e128e971e980a4e98637953909a27415701f00db.

PR #2269, opened 2014-09-10
general: no INSERT delayed usage

### LINKED ISSUE #2268: general: no INSERT delayed usage
As @tsgit reports:
```
Inspire was completely unreachable for ~17minutes this afternoon US
time, shortly after midnight CERN time
[...]
there are no errors in the logs on wns, but mysql processlist showed a
large number of "INSERT DELAYED" connections, e.g.
[...]
so it looks like we create a bottleneck here with spikes in search activity?

This appears quite similar to another recent outage I observed in real
time, when I didn't get a chance to look at mysql processlist in time.
[...]
the "DELAYED" is not doing what you think it is, because:

http://dev.mysql.com/doc/refman/5.5/en/insert-delayed.html
[...]
Beginning with MySQL 5.5.7, INSERT DELAYED is always handled as a
simple INSERT (that is, without the DELAYED option) whenever the value
of binlog_format is STATEMENT or MIXED. (In the latter case, the
statement no longer triggers a switch to row-based logging, and so is
logged using the statement-based format.)
```

So it looks like in the common case `INSERT DELAYED` is producing more harm than benefits. And this is more critical the bigger the site. Given `INSERT DELAYED` is not even SQL standard. I propose here to no longer use it.

kaplun:
* No longer uses INSERT DELAYED, since, beginning with MySQL 5.5.7,
  INSERT DELAYED is always handled as a simple INSERT (that is,
  without the DELAYED option) whenever the value of binlog_format is
  STATEMENT or MIXED. (In the latter case, the statement no longer
  triggers a switch to row-based logging, and so is logged using the
  statement-based format.)
  (closes #2268)

Reported-by: Thorsten Schwander <thorsten.schwander@gmail.com>

### DIFF
--- a/modules/bibdocfile/lib/bibdocfile.py
+++ b/modules/bibdocfile/lib/bibdocfile.py
@@ -2150,7 +2150,7 @@ class BibDoc(object):
         run_sql('DELETE FROM bibrec_bibdoc WHERE id_bibdoc=%s', (self.id, ))
         run_sql('DELETE FROM bibdoc_bibdoc WHERE id_bibdoc1=%s OR id_bibdoc2=%s', (self.id, self.id))
         run_sql('DELETE FROM bibdoc WHERE id=%s', (self.id, ))
-        run_sql('INSERT DELAYED INTO hstDOCUMENT(action, id_bibdoc, doctimestamp) VALUES("EXPUNGE", %s, NOW())', (self.id, ))
+        run_sql('INSERT INTO hstDOCUMENT(action, id_bibdoc, doctimestamp) VALUES("EXPUNGE", %s, NOW())', (self.id, ))
         run_sql('DELETE FROM bibdocfsinfo WHERE id_bibdoc=%s', (self.id, ))
         del self._docfiles
         del self.id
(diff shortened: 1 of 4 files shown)


PXke:
Any ETA ?

kaplun:
IMHO, if Travis happen to confirm (which I doubt this days, as it fails for master), the branch was good to go... I have just re-rebased it.

PXke:
Okay, great ! Let's wait the Travis answer so.

PXke:
It failed. Do we know a possible reason of why ?

kaplun:
Travis fails for *all* master branches :cry: There is some trouble with pip that I have not figured out yet how to solve. @tiborsimko, @jirikuncar Any idea? In tiborsimko/invenio-devscripts#28 I was trying to solve it by not upgrading pip, but I don't seem to be able to test it...

tiborsimko:
@kaplun I'll try to have a look on Sunday.  Regarding https://github.com/tiborsimko/invenio-devscripts/pull/28, please don't sign your PRs if they are not fully tested :smile:

tiborsimko:
@PXke @kaplun BTW the Travis build issue has been fixed.  I restarted the build.

tiborsimko:
>  I restarted the build.

... and it fails with the following interesting finding:

```
FAIL: bibupload - replace with different records does indeed touch history
  File "/opt/invenio/lib/python/invenio/bibupload_regression_tests.py", line 694, in test_replace_different_record
    self.assertNotEqual(self.timestamp, run_sql("SELECT modification_date FROM bibrec WHERE id=%s", (self.recid,)))
AssertionError: ((datetime.datetime(2015, 1, 14, 16, 26, 54),),) == ((datetime.datetime(2015, 1, 14, 16, 26, 54),),)
```

Note identical values.  Anyway, I reviewed the patch and it looks good and works well for me.

tiborsimko:

Thanks, amended commit log message and merged as 386ac9d5ef1041f1ac98725e0c3aea178d811f96.

PR #2275, opened 2014-09-12
WIP lxml: fixes and improvements

kasioumis:


tiborsimko:
How is this different from PR #2148 that is in WIP?

kasioumis:
@tiborsimko This seems to be indeed identical to https://github.com/inveniosoftware/invenio/pull/2148/ . The only explanation is that I did this PR without having seen the previous one by @kaplun . If @kaplun confirms, then we should go ahead and close this one.

PR #2373, opened 2014-10-06
bibcirculation: fix bug blocking library creation

PXke:
* Fixes a bug due to a wrong argument washing name, indeed
  req argument name is lib_type, argument name waited was
  type.

* Reactivates the creation of library in bibcirculation.


tiborsimko:
Thanks, amended commit log message and merged as 2080fdb795d94d1106cd09e158901b7b539d7ece.

PR #2375, opened 2014-10-06
WebStat: Enable custom summary

PXke:
* Adds the dependency matplotlib.

* Fixes the fact that path was always printed.

* Removes annoying try except.

### DIFF
--- a/modules/webstat/lib/webstat_engine.py
+++ b/modules/webstat/lib/webstat_engine.py
@@ -1886,15 +1886,19 @@ def create_custom_summary_graph(data, path, title):
     """
     # If no input, we don't bother about anything
     if len(data) == 0:
-        return
+        return False
     os.environ['HOME'] = CFG_TMPDIR
 
     try:
         import matplotlib
         matplotlib.use('Agg')
         import matplotlib.pyplot as plt
-    except ImportError:
-        return
+    except:
+        from invenio.errorlib import register_exception
+        register_exception()
+        return False
+
+
     # make a square figure and axes
     matplotlib.rcParams['font.size'] = 8
     labels = [x[0] for x in data]
@@ -1940,7 +1944,7 @@ def create_custom_summary_graph(data, path, title):
     plt.figlegend(patches, labels, 'lower right', **legend_keywords)
     plt.savefig(path)
     plt.close(gfile)
-
+    return True
 # GRAPHER
 
 def create_graph_trend(trend, path, settings):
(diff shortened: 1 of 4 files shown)


modules/webstat/lib/webstat_engine.py
```diff
@@ -1886,15 +1886,14 @@ def create_custom_summary_graph(data, path, title):
...
-        matplotlib.use('Agg')
-        import matplotlib.pyplot as plt
-    except ImportError:
-        return
+
+    import matplotlib
+    matplotlib.use('Agg')
+    import matplotlib.pyplot as plt
+
     # make a square figure and axes
```
tiborsimko: matplotlib is only recommended (in the INSTALL file), not mandatory.  Since it is only use for this minor functionality, I think it is good to keep try/except.  But you can surely put it to requirements for those installations that use pip.  This would make graph OK, but would not break other installations, or ask them to install it.  (Kind of defensive programming.)  Can you please amend your PR in this respect?

requirements-extras.txt
```diff
@@ -17,3 +17,4 @@ nose
 nosexcover
 flake8
 raven==4.2.1
+matplotlib
```
tiborsimko: We could even move it to the main requirements file...

tiborsimko:
Changing title to WIP until amended.

modules/webstat/lib/webstat_engine.py
```diff
@@ -1886,15 +1886,18 @@ def create_custom_summary_graph(data, path, title):
...
     try:
         import matplotlib
         matplotlib.use('Agg')
         import matplotlib.pyplot as plt
-    except ImportError:
-        return
+    except:
+        from invenio.errorlib import register_exception
+        register_exception()
+
```
tiborsimko: It is OK to raise an exception, but let us keep `return` here, so that the page won't lead to Internal Server Error on installations that do not have matplotlib.  This was my thinking behind https://github.com/inveniosoftware/invenio/pull/2375#discussion_r22529417

If OK for you, I'll amend this when merging the PR.

PXke: Sorry you are right.... It was what I was thinking to do... spread attention is not always a gain of productivity it seems. 

I have already fix it I push it in 5 secs :) !

tiborsimko:

Thanks for the changes, amended commit log message and merged as df2468da7a90731779f74a7589d1c55716bf979c.

PXke:
Thank you, for you help in the process. 

Cheers.

PR #2388, opened 2014-10-08
WebStat: Set default query for the custom summary

kasioumis:
* According to RQF0382332 for CDS: Set a default query for the
  "Custom query summary" under "Administration" / "Statistics".


tiborsimko:
Thanks, amended commit log message and merged as 864c8ce24643b48ff1178ae5904d531dc99f2346.

PR #2393, opened 2014-10-09
authorids: removal of legacy code

MSusik:
* Removes bibauthorid and webauthorprofile from Invenio.
* Moves external dependencies to other modules.

Signed-off-by: Mateusz Susik <mateusz.susik@cern.ch>


tiborsimko:
@MSusik The author pages (`/author/...`) were active even when a site decided not to use BibAuthorID module for author disambiguation.  E.g. imagine a site that has all its authors uniquely identified already via a phone book, for example. In this case, there is no need to run the disambiguator, but it is still useful to display `/author/...` summary pages for its authors.  Hence the two facilities were independent. The demo site contains a simple working example: 

- http://invenio-demo.cern.ch/author/Ellis,%20N  

It would be good to retain these pages even when BibAuthorID module is removed.

kaplun:
@tiborsimko but we are planning to transform author profiles into native Invenio records (a.k.a. authority records) so there shouldn't be any need for further /author/ support in this case. Maybe just preserving a redirect.

tiborsimko:
> Maybe just preserving a redirect.

Since this branch would remove author pages completely, there is nowhere to redirect to, isn't it? Imagine a site updating from Invenio 1 to Invenio 2; the `/author` URL would be gone.  Hence @jirikuncar's comment above asking to keep at least a stub message.   

Note that my comment went even further, explaining that BibAuthorID and WebAuthorProfile were really covering two different use cases, and even though we decided to remove the disambiguator, the author page displayer is an independent thing in the name-based mode; which could ideally be retained.  Nextifying profile pages in name-based mode would go further than keeping basic stub pages, so depending on the manpower and the time plan, and other related name-based mode use cases such as `/keyword/foo`, it may or may not be advantageous to attack this now.

kaplun:
> Note that my comment went even further, explaining that BibAuthorID and WebAuthorProfile were really covering two different use cases, and even though we decided to remove the disambiguator [...]

Yes but, as I understand it, we are reimplemeting webauthorprofile using native records and fancy formatting (with appropriate widgets etc.).

Could the name-based URLs be implemented as simple redirectors to given records by plugging some special few-lines Flask route?

jirikuncar:
@kaplun IMHO you don't have to implement it now. Just add a blueprint that handles `/authors/<name>` urls and displays a WIP message (see also https://github.com/inveniosoftware/invenio/pull/2393#issuecomment-66971410).

tiborsimko:
> Yes but, as I understand it, we are reimplemeting webauthorprofile using native records

Not in the case of name-based mode, when there will be no ID records *per se*.  E.g. see my hypothetical example of `/keyword/foo` or `/subject/foo` etc, when one could display "summary pages" for a various entities, always basing on name-value comparisons, without having to have any formal authority record.  (Kind of like http://invenio-demo.cern.ch/author/Ellis,%20N used to do.)

jirikuncar:
@MSusik I think the block should be called `body` in https://github.com/MSusik/invenio/blob/pu_baiclean/invenio/modules/authors/templates/author/author_stub.html#L23

MSusik:
Now it's called `body`.

PR #2395, opened 2014-10-09
WIP BibAuthorID: improvements for master

kaplun:
This superseeds #2260


tiborsimko:
What do we do with this PR?  Shall we keep given later v1.3 milestone even though the priority is critical?

kaplun:
Have to double check. I'll do it this week. I think with the last bibindex fix I solved all failing regression test. However it needs to be rebased against latest master which now contains all BibEdit improvements

jirikuncar:
Invenio 1.x is in a feature freeze state. Please continue in https://github.com/inveniosoftware/invenio-authorities.

kaplun:
More specifically at INSPIRE we have reimplemented the author disambiguation project into a suite of components:

* beard: https://github.com/inspirehep/beard (Bibliographic Entity Automatic Recognition and Disambiguation), a ML library focused on bibliographic metadata
* beard-server: https://github.com/inspirehep/beard-server, a Celery+REST wrapper around beard
* a specific sub-module in our overlay: https://github.com/inspirehep/inspire-next/pull/1353 that is responsible for the middleware, i.e. understanding which literature records have been modified, send the needed data to beard-server, and apply the response back.

Issue #2197, opened 2014-09-03
formatter/records: recjson output format hidden fields (don't expose sensitive data)


lnielsen:
With the MARC/MARCXML export you can hide specific fields using ``CFG_BIBFORMAT_HIDDEN_TAGS`` to not expose sensitive data like e.g. email addresses.

The newly added recjson output format does not have this feature and thus will expose sensitive data if included in the recjson configuration. As far as I understand @egabancho is already working on a fix for this, however I wasn't able to find the ticket so adding one now.

tiborsimko:
Fixed in master for Invenio 1, so keeping only next labels for Invenio 2.

tiborsimko:
Since the issue is labelled as critical for INSPIRE, and since nobody has time to work on `@hidden` or `@hidden_if` decorators, putting this for triaging discussion. Let's assign at least semi-passive, semi-active ticket owner.

kaplun:
Yep this *was* critical for INSPIRE as long as master was concerned. At the moment, INSPIRE Labs (i.e. our pu-based instance) is not serving records, so it's no longer critical for us (still major)

jirikuncar:
The `@hidden` decorator has been implemented and merged in #2666. Do we need the `@hidden_if` version and if yes how it should work in respect to `request`, `current_user` and application context?

tiborsimko:
> The @hidden decorator has been implemented and merged in #2666.

Great that you made the time!

> Do we need the @hidden_if  version

I'd say nope. In theory, we did muse about finer granularity, this-field-seen-only-by-this-role, that-field-seen-only-by-that-role.  In practice, we had only one use case so far, where all the fields decorated as hidden can be seen by cataloguers. Hence I'd say let's close this issue now, and if needed, we'll open a new one for a more fine-grained conditional behaviour.

PR #2396, opened 2014-10-10
BibField: CFG_BIBFORMAT_HIDDEN_RECJSON_FIELDS

kaplun:
* Introduces ad-interim new config variable
  CFG_BIBFORMAT_HIDDEN_RECJSON_FIELDS to list which recjson fields
  need to be not exposed via public interfaces.
  (addresses #2197)


tiborsimko:
> WDYT?

As agreed, let's commit to v1.2.0 and let's amend it in next/pu.  However, it seems `@hidden` and/or `@hidden_if` functionality will not come anywhere soon due to lack of man time...

tiborsimko:
Thanks, amended commit log message and merged as 8076dc3f2899b5840d12920cf8a45d59d5984dda.

PR #2397, opened 2014-10-13
WIP BibCheck: new ticket creation policies

### LINKED ISSUE #2383: RFC bibcheck: optionally create one ticket per rule per record
bibcheck currently submits a single ticket for all the problems of each record. This can be problematic for a number of reasons:
- The title carries almost no valuable information (`"Bibcheck rule failed in record %s"`)
- It is hard to assign specific issues to specific members of a team (eg. have Michael find and fix issues that stem from the `mandatory` check)

This problem can be tackled by introducing a new configuration option (`splits_tickets`) to allow for creating one ticket per record per rule.
- `modules/bibcheck/lib/bibcheck_task.py:submit_ticket` would read this configuration and submit as many tickets as necessary.
- Adding tags to the subject might also be nice to have (eg `[Error] [ID:3] [rule:mandatory] Missing mandatory field 005%%_`). This would provide a way to search for specific rules or issues. (Despite not being ideal, this way is generic and should work with any existing interface)

@jmartinm please let me know if there are any issues or omissions in this RFC.

dset0x:
* Adds the `--ticket-creation-policy` option which can be either
  `per-rule`, `per-record`, `per-rule-per-record` or `no-tickets`. These
  control how tickets are dispatched, if at all.  `--no-tickets` is
  preserved for backwards compatibility.
  (closes #2383)

* Turns `errors`, `amendments` and `warnings` of records into instances of
  the new `Issue` class. This allows for a more coherent way of storing and
  extracting this information.

* Changes the call to `BIBCATALOG_SYSTEM.ticket_submit` to pass recordid
  only if the ticket concerns a single record.

* Fixes a typo in a comment.

### DIFF
--- a/modules/bibcheck/lib/bibcheck_task.py
+++ b/modules/bibcheck/lib/bibcheck_task.py
@@ -892,7 +892,14 @@ def main():
                                separated), ignoring all other filters
   -q, --queue=queue        Create tickets in the specified RT Queue (Default
                                Bibcheck)
-  -t, --no-tickets         Don't create any ticket in RT. Useful for debugging
+  -t, --no-tickets         Don't create any ticket in RT. Useful for debugging.
+                           To be deprecated by --ticket-creation-policy.
+  -p, --ticket-creation-policy Policy for ticket creation in RT. May be
+                               'per-rule', to create one ticket for each rule,
+                               'per-record' to create a ticket for each record,
+                               `per-rule-per-record` to create a separate ticket
+                               for every issue, or `no-tickets` to disable
+                               ticket creation.
   -b, --no-upload          Don't upload changes to the database
   -n, --dry-run            Like --no-tickets and --no-upload
   -c, --config             By default bibcheck reads the file rules.cfg. This
--- a/modules/bibcheck/lib/bibcheck_task.py
+++ b/modules/bibcheck/lib/bibcheck_task.py
@@ -132,12 +132,23 @@ class Tickets(object):
     def _generate_subject(self, issue_type, record_id, rule_name):
         """Generate a fitting subject based on what information is given."""
         assert any((i is not None for i in (issue_type, record_id, rule_name)))
+        if issue_type:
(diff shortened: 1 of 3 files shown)


tiborsimko:
@dset0x Can you please add appropriate `Signed-off-by` and `Reviewed-by` directives to the commit message?  It seems that at least @jmartinm reviewed your branch, so you could add him.

@jmartinm @kaplun Was this deployed on INSPIRE already? Can you set appropriate milestone?

kaplun:
Nope not yet deployed. I believe there is still some pending discussion as well with @ksachs . To be checked.

tiborsimko:
> I believe there is still some pending discussion as well with @ksachs

Maybe you are referring to #2474 (that I merged already), or was there something else?

ksachs:
I have no idea what you are talking about ... 


----- Original Message -----

dset0x:
@ksachs This is the patch that allows for submitting tickets in bundles (eg: One ticket in the HoldingPen per check).

@kaplun I can't recall any unaddressed issue, other than that we have not ran this through some usability test.

modules/bibcheck/lib/bibcheck_task.py
```diff
@@ -72,19 +75,168 @@ def __init__(self, rule_name, error):
...
+    def resolve_ticket_creation_policy(self):
+        """Resolve the policy for creating tickets."""
+        ticket_creation_policy = \
+            task_get_option('ticket_creation_policy', 'per-record')
+
+        known_policies = ('per-rule',
+                          'per-record',
+                          'per-rule-per-record',
+                          'no-tickets')
+        if ticket_creation_policy not in known_policies:
```
kaplun: Can you add proper documentation in `main()`?

dset0x: @kaplun Sure, done and rebased.

kaplun:
This is superseeded by #2841

PR #2404, opened 2014-10-14
WIP BibCheck: collection-based filters and timestamps

### LINKED ISSUE #2387: BibCheck: collection-based filters and timestamps
`BibCheck` currently allows for applying rules only within a specific collections. This is based on collection information produced in background by `webcoll`. Everytime `BibCheck` run for a given rule, it will operate over all the new and modified records since the previous run for the specific rule. However this set of record is intersected with the collection information (when specified) which is computed e.g. every 5 minutes (depending on task configuration), therefore some records might end up never being considered by a given rule.

To solve this, when a collection is being specified for a rule, at the end of the run, instead of storing the timestamp of the begin of bibcheck run, what has to be stored is the _earliest date among all the collections involved and the beginning of the execution of the corresponding rule_.

dset0x:
* Store the earliest date among last time webcoll was run and the last
  bibindex `last_update` as the `last_run`, instead of the last time the
  rule was initiated. This prevents a race condition where records
  inserted between a webcoll run and the beginning of the following
  bibcheck ran were never considered by bibcheck.
  (resolves #2387)

Reported-by: Samuele Kaplun <samuele.kaplun@cern.ch>

### DIFF
--- a/modules/bibcheck/lib/bibcheck_task.py
+++ b/modules/bibcheck/lib/bibcheck_task.py
@@ -62,6 +62,7 @@ from invenio.dbquery import run_sql
 from invenio.bibcatalog import BIBCATALOG_SYSTEM
 from invenio.shellutils import split_cli_ids_arg
 from invenio.jsonutils import json
+from invenio.websearch_webcoll import get_cache_last_updated_timestamp
 
 CFG_BATCH_SIZE = 1000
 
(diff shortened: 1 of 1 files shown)


dset0x:
@kaplun Has this been tested since?

kaplun:
Nope :flushed:

kaplun:
Superseeded by #2841.

PR #2418, opened 2014-10-16
installation: uninstall-jquery-plugins amendments

makistsantekidis:
* Amended deletion of files and folders that were
  not deleted during uninstall-jquery-plugins even
  though they were added by install-jquery-plugins

* Minor format fix


tiborsimko:
Thanks, enriched the patch by adding `rm -rf jqueryFileTree` as well, amened commit log message and merged as 3fac58af3a390ef78c2d145b87a0748767fca8a8.

PR #2434, opened 2014-10-20
WebComment: more prominent subscription link

kasioumis:
* Makes the subscription link more prominent by placing it both at
  the top and at the bottom of the comments, when there are comments.


tiborsimko:
Thanks, amended commit log message to add PR number, and merged as 0c6c4cebe24432c948d364e7b5cac5591a5ee611.

(Note that the multi-line statements use wrong hanging indentation, you may want to run kwalitee checker tool and fix the warnings.)

PR #2455, opened 2014-10-27
WebComment: fixes unnecessary page reload

### LINKED ISSUE #2454: WebComment: toggling comment visibility causes page reload
In [modules/webcomment/lib/webcomment_templates.py:441](https://github.com/inveniosoftware/invenio/blob/2c4d90d805b8b2783c96eb28f8314d8a812e8e60/modules/webcomment/lib/webcomment_templates.py#L441), in the anchor tag for the toggling of comments there is an unnecessary href tag (`href="%(toggle_url)s"`) pointing to the toggling url, which refreshes the page with the comments toggled to be visible/hidden 

```
toggle_visibility_block = """<div class="webcomment_toggle_visibility"><a id="collapsible_ctr_%(comid)s" class="%(collapse_ctr_class)s" href="%(toggle_url)s" onclick="return toggle_visibility(this, %(comid)i);" title="%(collapse_label)s"><span style="display:none">%(collapse_label)s</span></a></div>""" % \
                                    {'comid': com_id,
                                     'toggle_url': create_url(CFG_SITE_URL + '/' + CFG_SITE_RECORD + '/' + str(recID) + '/comments/toggle', {'comid': com_id, 'ln': ln, 'collapse': collapsed_p and '0' or '1', 'referer': user_info['uri']}),
                                     'collapse_ctr_class': collapsed_p and 'webcomment_collapse_ctr_right' or 'webcomment_collapse_ctr_down',
                                     'collapse_label': collapsed_p and _("Open") or _("Close")}
```

There is an already implemented ajax call to hide the comment and is already in place for the existing comments (see [here](https://github.com/inveniosoftware/invenio/blob/2c4d90d805b8b2783c96eb28f8314d8a812e8e60/modules/webcomment/lib/webcomment_templates.py#L749-L787)). The only thing needed is to remove the href attribute from the anchor tag.

makistsantekidis:
fixes #2454 

* removes href tag from the toggle anchor so that when toggling comments
  only an ajax request is fired and not a page reload.

### DIFF
--- a/modules/webcomment/lib/webcomment_templates.py
+++ b/modules/webcomment/lib/webcomment_templates.py
@@ -2,7 +2,7 @@
 ## Comments and reviews for records.
 
 ## This file is part of Invenio.
-## Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013 CERN.
+## Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


tiborsimko:
@kasioumis Is this deployed on CDS?  Which milestone do you want this to go to?

egabancho:
I think this change does not affect the way the toggle works for comments.

`href` is only used in case javascript is disabled on the browser, in which case reloads the hole page. Otherwise it does the AJAX query.

If we say that javascript is something every one is using nowadays, then we can apply this commit, otherwise we can close this PR.

BTW @tiborsimko it is not deploy on CDS and there is no hurry for integration

tiborsimko:
@egabancho Please set milestone accordingly...

egabancho:
See https://github.com/inveniosoftware/invenio/pull/2455#issuecomment-69178857

Issue #2462, opened 2014-10-28
WebComment: comment attachment file size checked after upload


makistsantekidis:
When adding the attachment to a comment, there is a configurable size limit to it (`CFG_WEBCOMMENT_MAX_ATTACHMENT_SIZE`).

This size is checked after the attachment has been uploaded and saved on disk. Testing it with very large files ( > 1Gb when default limit is < 5Mb ) made the server freeze for a couple of minutes because the file was too large to keep in memory and then save it on disk, before checking its size and deleting short after.

There is some commented code inside [webcomment_webinterface.py](https://github.com/inveniosoftware/invenio/blob/2c4d90d805b8b2783c96eb28f8314d8a812e8e60/modules/webcomment/lib/webcomment_webinterface.py#L411-L414) that was probably supposed to be used to read the file chunk by chunk, but after uncommenting and modifying it to make the upload stop when it surpassed the file size limit, I got errors from invenio webinterface_handler.

PR #2463, opened 2014-10-28
WebComment: adds file check for comment attachment

makistsantekidis:
temporary solution for #2462 

Adds a check for attaching files to a comment to make sure they are
less than the configured maximum size and deletes them before 
submitting them.

### DIFF
--- a/modules/webcomment/lib/webcomment_templates.py
+++ b/modules/webcomment/lib/webcomment_templates.py
@@ -1238,6 +1238,60 @@ class Template:
             <noscript>
             <input type="file" name="commentattachment[]" /><br />
             </noscript>
+            <script>
+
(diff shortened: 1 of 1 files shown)


tiborsimko:
@kasioumis Is this deployed on CDS?  Shall we aim for non-temporary solution? Which milestone do we give this?

kasioumis:
@tiborsimko This is not deployed on CDS yet. I tend to think we should aim for non-temporary solution, therefore giving it a v1.x milestone for now.

tiborsimko:
If this was deployed on CDS, let's merge into v1.3 already.

If this was not deployed yet, let's close this and address the problem in the new `master` codebase.

PR #2512, opened 2014-11-11
WebSearch: fix number of printed records

egabancho:
* Fixes the number of records printed when a ranking methond is
  used.

Co-authored-by: Nikolaos Kasioumis <nikolaos.kasioumis@cern.ch>


tiborsimko:
Thanks, amended commit log message and merged as 
176953e5db0451cde764a20f7da73380c123db13.

P.S. It would be good to add a test case for fixes of this kind, to make sure the error does not reappear.

PR #2525, opened 2014-11-14
OAIRepository: lower priority to updating uploads

kaplun:
* Attributes lower priority to updating uploads (generated
  by oai_repository_updater). Also, assigns a specific
  -N argument so that these bibuplods can run concurrently
  to other.

* Checks whether other uploads from previous execution of
  oai_repository_updater need still to be elaborated by
  the queue. In this case skip the execution of
  oai_repository_update.


tiborsimko:
Thanks, amended commit log message and merged as 903d435069a2e0cc1ea9cc0a16332434742abb60.

PR #2568, opened 2014-11-25
WIP checker: port to 2.x

dset0x:
Work so far done is regarding a stable configuration format interface to work towards an API, using manager.py and utilizing workflows.


invenio/modules/checker/errors.py
```diff
@@ -0,0 +1,27 @@
...
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Exceptions for checker module."""
+
+class PluginMissing(Exception):
+    def __init__(self, pluginspec, rule_name):
+        message = "Could not find plugin `{0}` as defined in `{1}`"\
+            .format(pluginspec, rule_name)
+        super(PluginMissing, self).__init__(message)
+
```
jirikuncar: - empty line at the end of the file
- missing class and `__init__` docstring

invenio/modules/checker/registry.py
```diff
@@ -0,0 +1,93 @@
+# -*- coding: utf-8 -*-
+#
```
jirikuncar: `##`

invenio/modules/checker/rules.py
```diff
@@ -0,0 +1,263 @@
...
+from pykwalify.errors import SchemaError
+from importlib import import_module
+from intbitset import intbitset
+
+from .registry import config_files, schema_files
+from .common import ALL
+from invenio.legacy.search_engine import perform_request_search
+
+
+class DuplicateRuleError(Exception):
```
jirikuncar: move to `errors.py`

invenio/modules/checker/rules.py
```diff
@@ -0,0 +1,293 @@
...
+            force_finiteness = True
+        if force_finiteness:
+            yield ('p', '* AND *')
+
+    def _query_options(self):
+        """Convert the options section of a rule into query arguments."""
+        # HACK: Force `search_pattern` to return deleted records.
+        try:
+            if self['options']['consider_deleted_records']:
+                return {'ap': -9, 'req': None}
```
dset0x: Don't return None here.

invenio/modules/checker/rules.py
```diff
@@ -0,0 +1,293 @@
...
+        except KeyError:
+            force_finiteness = True
+        if force_finiteness:
+            yield ('p', '* AND *')
+
+    def _query_options(self):
+        """Convert the options section of a rule into query arguments."""
+        # HACK: Force `search_pattern` to return deleted records.
+        try:
+            if self['options']['consider_deleted_records']:
```
dset0x: Don't return None if this is False.

invenio/modules/checker/rules.py
```diff
@@ -0,0 +1,293 @@
...
+
+"""Rule handlers for checker module."""
+
+import json
+import yaml
+from argparse import ArgumentTypeError
+from collections import defaultdict, MutableSequence
+from importlib import import_module
+from intbitset import intbitset
+from pykwalify.core import Core  # TODO: Remove when cfg is moved to db
```
jirikuncar: can we use `cerberus

invenio/modules/checker/tasks/tasks.py
```diff
@@ -0,0 +1,210 @@
...
+    #     return
+
+    records = (_load_record(extra_data, recid, local_storage_only=True)
+               for recid in recids)
+    records_xml = ""
+    for record in records:
+        records_xml += record.legacy_export_as_marc()
+    if not records_xml:
+        return
+    records_xml = (
```
jirikuncar: it would be better to use `formatter` here.

invenio/modules/checker/rules.py
dset0x: This configuration is moving to the database as per the comment. It will be then converted to a dict to interface with modules like this, at which point we can use cerberus, although I believe there should be no need for that if the converted is thoroughly tested.

invenio/modules/checker/tasks/tasks.py
dset0x: Formatter seems to only provide functions for printing records by the record ID, but I need to print records from their `Record` here. I ended up extending [this](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/workflows/definitions.py#L118) method yesterday (being oblivious to the fact that `formatter` is a module), to support multiple records. It still uses the legacy formatter though (and does a lot of hacking around with the XML).
Do we have another method that does what I need to do or should I move the code that I've extended from above into the formatter?

dset0x: @jirikuncar Formatter seems to only provide functions for printing records by the record ID, but I need to print records from their Record here. I ended up extending this method yesterday (being oblivious to the fact that formatter is a module), to support multiple records. It still uses the legacy formatter though (and does a lot of hacking around with the XML).
Do we have another method that does what I need to do or should I move the code that I've extended from above into the formatter?

jirikuncar: @dset0x that's why there is a RFC #2662, where we should agree on the correct way to go (cc @tiborsimko).

jirikuncar:
@dset0x I have noticed 2 commits that are enhancing `invenio.ext.sqlalchemy.utils`.  Could you move them to separate PR? Could you also provide more detailed example/test-case for the enhanced version of `to_dict()` method?

--
PS: Do you think it could worth it to implement custom JSON serialized instead?

dset0x:
@jirikuncar Sure, I will move the new `todict`, `CompositeItems` and `CompositeChecker` (renamed) to a new PR once I have the chance. It is definitely worth explaining further why I wrote the changes and how they are to be used, so I will add comments and discuss in that PR.

--
A pair of `dump`/`load` methods was my first approach, but I realized that I was writing non-reusable code for something that others could benefit from.

Issue #2570, opened 2014-11-25
bibknowledge: kb_export searchtype is substring match by default


pedrogaudencio:
**Issue:**
I'm trying to map the arXiv categories to the Inspire ones form a knowledge base, since I'm doing it in the client-side, I make an ajax request to `/kb/export?kbname=<kbname>&format=json`. The workflow is [`request`](https://github.com/inveniosoftware/invenio/blob/pu/invenio/legacy/bibknowledge/webinterface.py#L119-L124) > [`kb_export()`](https://github.com/inveniosoftware/invenio/blob/pu/invenio/legacy/bibknowledge/admin.py#L624-L628) > [`get_kb_mappings_json()`](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/knowledge/api.py#L201) > [`get_kb_mappings()`](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/knowledge/api.py#L104-L106).

**Problem:**
Since the `searchtype` in [`kb_export`](https://github.com/inveniosoftware/invenio/blob/pu/invenio/legacy/bibknowledge/admin.py#L584) is by default `"s"`, it always queries [substring key/value matches](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/knowledge/api.py#L104-L106).

**Proposal:**
Adding the `searchtype` parameter would make it empty by default in the export requests:
```diff
diff --git a/invenio/legacy/bibknowledge/webinterface.py b/invenio/legacy/bibknowledge/webinterface.py
index 03e9772..faf7256 100644
--- a/invenio/legacy/bibknowledge/webinterface.py
+++ b/invenio/legacy/bibknowledge/webinterface.py
@@ -76,6 +76,7 @@ class WebInterfaceBibKnowledgePages(WebInterfaceDirectory):
                                    'expression': (str, ''),  # for dynamic kbs
                                    'collection': (str, ''),  # for dynamic kbs
                                    'kbname': (str, ''),  # for exporting
+                                   'searchtype': (str, ''),  # for exporting
                                    'format': (str, ''),  # for exporting
                                    # for exporting to JQuery UI
                                    'term': (str, ''),
@@ -108,6 +109,7 @@ class WebInterfaceBibKnowledgePages(WebInterfaceDirectory):
         delete_mapping = argd['delete_mapping']
         kbname = argd['kbname']
         format = argd['format']
+        searchtype = argd['searchtype']
         limit = argd['limit']
 
[...]

PR #2581, opened 2014-11-26
BibKnowledge: searchtype parameter addition

pedrogaudencio:
* The searchtype was "s" by default, it always queried substring  key/value matches. Adding the searchtype parameter makes it empty by default in the export requests.

Fixes inspirehep/inspire-next#128.

Addresses #2570.


tiborsimko:
Thanks, reviewed, amended commit log message and merged as eac2b3ce46cd0d1ba4ae37f60f59ac676ca6a454.

PR #2587, opened 2014-11-26
search: engine based on query parser

jirikuncar:
This is initial work on separating search engine from formatter, ranker, sorter and it brings new architecture using *Flask-Registry* for basic search unit searchers. 

**Please note that naming is not yet finished and especially `invenio.modules.search.{api,engine}` will change in near future.**


lnielsen:
OAI server could perhaps use https://pypi.python.org/pypi/pyoai

tiborsimko:
> OAI server could perhaps use https://pypi.python.org/pypi/pyoai

Have you looked at pros/cons?  

The current OAI server uses search queries to look for information in the record metadata; so if we change these search calls, then it should still work OK for some time to come, even with new searhc engine and even with Elasticseach . (Until somebody has time to "nextify" the OAI repository server component properly.)

kaplun:
Hi! You are accumulating here an great amount of new stuff since some months. To be future compatible it would be great if all this can be merged into a mainstream Invenio branch (e.g. pu2 :smile: ) Somehow this is needs to be the future maint-2.1 branch... Shall we discuss this in triage?

jmartinm:
In ``invenio/legacy/websearch/webinterface.py``

```python
     output_formats = [output_format['attrs']['code'].lower() for output_format in \
-                      get_output_formats(with_attributes=True).values()]
+                      get_output_formats(with_attributes=False).values()]

```

jmartinm:
Just trying this branch with the INSPIRE overlay. First problem I see is if I search, e.g ``Supermoduli`` I find one record.

facet-search

But clicking on the facet, the search that can be seen in the sceenshot does not yield any result.

In the debug console I have:

```python
DEBUG in native [/home/jmartinm/.virtualenvs/pu-inspire2/src/invenio/invenio/modules/search/searchext/engines/native.py:134]:
search_unit: f=author, p=Witten, Edward, m=e
```

hachreak:
I have a question....

screenshot - 03122015 - 01 27 24 pm

The facets works with this PR?
Because if I search `author:"Girardello, L"` (as if you clink on facet) the results are empty, but if I search `author:"Girardello"`, it found the record.

hachreak:
I have an other problem:

it seems that the facets are not update after search something...
I tried to: 
 - search `of` 
 - click on collection `Multimedia & Arts`
 - I can see in author list `Ellis, J` but she should not be there.. :P

If you try: 
 - click `search` 
 - put on search engine directly of collection: `Multimedia & Arts` and click "search"
 - you should have differents facets (and maybe more correct)

PR #2588, opened 2014-11-27
MiscUtil: Added CERN LDAP plugin.

switowski:
* Creates ldap_cern plugin that can be used to retrieve information from
  LDAP at CERN (it's faster than the LDAP module in bibcirculation due
  to the modified queries and more generic - it returns lists with all
  users that match the query, so the filtering can be done outside of
  this plugin).

### DIFF
--- a/modules/miscutil/lib/ldap_cern_unit_tests.py
+++ b/modules/miscutil/lib/ldap_cern_unit_tests.py
@@ -1,0 +1,57 @@
+# -*- coding: utf-8 -*-
+##
+## This file is part of Invenio.
+## Copyright (C) 2008, 2009, 2010, 2011, 2013, 2014 CERN.
+##
+## Invenio is free software; you can redistribute it and/or
+## modify it under the terms of the GNU General Public License as
+## published by the Free Software Foundation; either version 2 of the
+## License, or (at your option) any later version.
+##
+## Invenio is distributed in the hope that it will be useful, but
+## WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Unit tests for the ldap_cern library."""
+
+from invenio.testutils import InvenioTestCase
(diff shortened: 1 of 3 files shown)


modules/miscutil/lib/ldap_cern_unit_tests.py
```diff
@@ -0,0 +1,57 @@
...
+## Invenio is distributed in the hope that it will be useful, but
+## WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Unit tests for the solrutils library."""
```
tiborsimko: s/solrutils/ldap_cern

tiborsimko:
@switowski Is this used in production on CDS?  Shall we merge for v1.2?

modules/miscutil/lib/ldap_cern_unit_tests.py
switowski: Good spot, sorry for that. I will fix it.

switowski:
@tiborsimko yes, it's in production, please merge it. It's also recommended that someone with the knowledge of bibcirculation module took a look at this since it's a more generic version of `bibcirculation_cern_ldap.py`

tiborsimko:
> yes, it's in production, please merge it. 

OK, two more things:

- please add test file to Makefile.am as well;

- IIRC anonymous xldap.cern.ch access is restricted to inside CERN only, so please either put client IP check, or else wrap call in `if CFG_CERN_SITE`; (at least for the test case)

- optionally, please consider renaming the module to `ldaputils` which would be more coherent with the other redisutils, htmlutils,  textutils and friends.  (We could also take configuration variables out, but this can wait until  there are non-CERN users of the module.)

> It's also recommended that someone with the knowledge of bibcirculation module took a look at this since it's a more generic version of bibcirculation_cern_ldap.py

@rsripada left, so BibCirculation is in a kind of hanging state, until @mvesper takes over in a couple of months.

switowski:
@jochenklein if I remember correctly, you were updating the LDAP plugin for CERN. Have you by any chance changed the modules/miscutil/lib/ldap_cern.py file ? If yes, maybe we can close this PR and just merge yours PR to invenio (since it contains updated version of my code).

jirikuncar:
Closing since Invenio 1.x is in feature freeze state.

PR #2600, opened 2014-12-01
oauthclient: local account discovery improvement

jmartinm:
* Uses userEXT table to do a lookup of the external identifier
  in order to find the local account corresponding to the user
  that logs in using oauth.

Signed-off-by: Javier Martin Montull <javier.martin.montull>

### DIFF
--- a/invenio/modules/oauthclient/contrib/orcid.py
+++ b/invenio/modules/oauthclient/contrib/orcid.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2014 CERN.
+## Copyright (C) 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
@@ -17,7 +17,7 @@
 ## along with Invenio; if not, write to the Free Software Foundation, Inc.,
 ## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
-""" Pre-configured remote application for enabling sign in/up with ORCID.
+"""Pre-configured remote application for enabling sign in/up with ORCID.
 
 **Usage:**
 
@@ -114,25 +114,20 @@ REMOTE_SANDBOX_APP['params'].update(dict(
 
 
 def account_info(remote, resp):
-    """ Retrieve remote account information used to find local user. """
+    """Retrieve remote account information used to find local user."""
+    account_info = dict(external_id=resp.get("orcid"), external_method="orcid")
 
-    from ..handlers import token_session_key
-
-    # Store ORCID of the user in session for later retrieval
-    session[token_session_key(remote.name) + "_orcid"] = resp.get("orcid")
-    return dict(email=None, nickname=None)
+    return account_info
 
 
 def account_setup(remote, token):
-    """ Perform additional setup after user have been logged in. """
-
+    """Perform additional setup after user have been logged in."""
     from invenio.ext.sqlalchemy import db
     from ..handlers import token_session_key
 
     extra_data = {
-        "orcid": session.get(token_session_key(remote.name) + "_orcid")
+        "orcid": session.get(token_session_key(remote.name) +
+                             "_account_info").get("external_id")
         }
     token.remote_account.extra_data = extra_data
     db.session.commit()
-    # Remove ORCID from session
-    del session[token_session_key(remote.name) + "_orcid"]
--- a/invenio/modules/oauthclient/handlers.py
+++ b/invenio/modules/oauthclient/handlers.py
@@ -206,6 +206,8 @@ def authorized_signup_handler(resp, remote, *args, **kwargs):
             if user is None:
                 # Auto sign-up requires extra information
                 session[token_session_key(remote.name)+'_autoregister'] = True
+                session[token_session_key(remote.name) +
+                        "_account_info"] = account_info
                 return redirect(url_for(
                     ".signup",
                     remote_app=remote.name,
@@ -276,8 +278,10 @@ def signup_handler(remote, *args, **kwargs):
     form = EmailSignUpForm(request.form)
 
     if form.validate_on_submit():
+        account_info = session.get(token_session_key(remote.name) +
+                                   "_account_info")
         # Register user
-        user = oauth_register(form.data)
+        user = oauth_register(account_info, form.data)
 
         if user is None:
             raise OAuthError("Could not create user.", remote)
@@ -300,6 +304,9 @@ def signup_handler(remote, *args, **kwargs):
 
         if not token.remote_account.extra_data:
             handlers['setup'](token)
+
+        # Remove account info from session
+        session.pop(token_session_key(remote.name)+'_account_info', None)
 
         # Redirect to next
         if request.args.get('next', None):
--- a/invenio/modules/accounts/models.py
+++ b/invenio/modules/accounts/models.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 ## This file is part of Invenio.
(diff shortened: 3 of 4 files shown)


jmartinm:
Comments addressed and rebased, please have a look @jirikuncar @lnielsen

jirikuncar:
@jmartinm looks good to me. :shipit:

invenio/modules/oauthclient/handlers.py
```diff
@@ -209,7 +209,9 @@ def authorized_signup_handler(resp, remote, *args, **kwargs):
                 return redirect(url_for(
                     ".signup",
                     remote_app=remote.name,
-                    next=request.args.get('next', '/')
+                    next=request.args.get('next', '/'),
+                    external_id=account_info.get("external_id"),
```
lnielsen: I think there's a security issue here. External id/method are passed in the request args to the form, and the code is then later relying on those values, meaning that the user can modified in between and thus fake them.

jmartinm:
@lnielsen good point. Thanks, I have updated the PR storing account_info in the session.

Can you have a look?

And I guess removing the account info from session here https://github.com/inveniosoftware/invenio/pull/2600/files#diff-e35f9879d99b458ee57ff1a2343f249cR140 is the right thing to do? We don't need it afterwards.

I will test this now on a test VM, have a look at the code and let me know if its looks ok.

jmartinm:
To follow up, just tried it in a VM and it works as expected.

invenio/modules/oauthclient/contrib/orcid.py
```diff
@@ -115,24 +115,26 @@
...
     from ..handlers import token_session_key
 
-    # Store ORCID of the user in session for later retrieval
-    session[token_session_key(remote.name) + "_orcid"] = resp.get("orcid")
-    return dict(email=None, nickname=None)
+    account_info = dict(email=None, nickname=None, orcid=resp.get("orcid"),
+                        external_id=resp.get("orcid"), external_method="orcid")
+
+    # Store account info of the user in session for later retrieval
+    session[token_session_key(remote.name) + "_accountinfo"] = account_info
```
jirikuncar: @jmartinm I would go for the same suffix like the variable name, hence WDYT about replacing `_accountinfo` with `_account_info`.

jmartinm:
I have updated the session variable to ``_account_info``

jirikuncar:
@jmartinm thanks! Do you think you could fix the docstrings at least in `orcid.py`?

see http://kwalitee.invenio-software.org/inveniosoftware/invenio/branches/d5c31da0c5e8f6947b5402c3519f6d68a1be0f81/jmartinm:local-account-discovery

I guess you can add @lnielsen and me as reviewes.

lnielsen:
First apologies for the many iterations on this PR. One last thing I've spotted.

Instead of setting/deleting ``account_info`` in the session in ``contrib/orcid.py``, it would better to do set ``account_info`` in ``authorized_signup_handler`` ([here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L208)) and retrieve/delete it in ``signup_handler`` ([here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L278) and [here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L285)) .

Additionally, modify ``oauth_register()`` ([here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/utils.py#L63)) to accept ``form.data`` and ``account_info`` (passed in [here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L279)).

This way every service (orcid, github, google, twitter), that uses the signup handler  doesn't need to set and retrieve the account info in the session.

jmartinm:
No problem @lnielsen , good suggestion. I will have a look and update the PR.

jmartinm:
Updated the PR.

> Instead of setting/deleting account_info in the session in contrib/orcid.py, it would better to do set account_info in authorized_signup_handler (here) and retrieve/delete it in signup_handler (here and here) .

I have put the deletion of the session here https://github.com/inveniosoftware/invenio/pull/2600/files#diff-8faf6f61294e068746c7eef9b029a0d1R308 as in some cases (like our orcid handler) one might want to access the account info in the ``setup`` handler.

> Additionally, modify oauth_register() (here) to accept form.data and account_info (passed in here).

Do you see any advantage on this? Currently the dict gets updated so if we have the same key, e.g. ``email`` both from ``acount_info`` and from ``form.data`` the one from the form will be taken.

Seems to me a bit simpler than modifying the signature of https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/utils.py#L64 and having to extract keys from 2 different places, but maybe I am missing some advantage.

invenio/modules/oauthclient/contrib/orcid.py
```diff
@@ -114,25 +114,21 @@
...
-    """ Perform additional setup after user have been logged in. """
-
+    """Perform additional setup after user have been logged in."""
     from invenio.ext.sqlalchemy import db
     from ..handlers import token_session_key
 
     extra_data = {
-        "orcid": session.get(token_session_key(remote.name) + "_orcid")
+        "orcid": session.get(token_session_key(remote.name) +
+                             "_account_info").get("orcid")
```
jirikuncar: If you use `"_account_info").get("external_id")` then you can remove `, orcid=resp.get("orcid")` from https://github.com/inveniosoftware/invenio/pull/2600/files#diff-e35f9879d99b458ee57ff1a2343f249cR118

lnielsen:
> Do you see any advantage on this? Currently the dict gets updated so if we have the same key, e.g. email both from acount_info and from form.data the one from the form will be taken.

Mainly for security. As it is now it's safe, but it's very easy to introduce unwanted security issues. E.g someone adds a poorly validated form field that overwrites a trusted account info value allowing the end-user to pass in any value. Or, an oauth app sets a value in account_info, and uses it later in account setup, however in-between account info data gets overwritten by form data (since you run ``account_info.update()``).

I would rather be careful in this case, since we're dealing with user authentication and sign up. Hence, being explicit about origin of values makes more sense to me.

jirikuncar:
Question: what about updating `form.data` with `user_info` on https://github.com/inveniosoftware/invenio/pull/2600/files#diff-8faf6f61294e068746c7eef9b029a0d1R284 ? Is there something what can get wrong?

Example:
```python
data = dict(form.data)
data.update(account_info)
```

jmartinm:
> Example

```python
data = dict(form.data)
data.update(account_info)
```

Problem with this is that, for example, right now [here](https://github.com/inveniosoftware/invenio/pull/2600/files#diff-e35f9879d99b458ee57ff1a2343f249cR118) we are returning ``email=None`` so the email that the user introduces in the signup form would be updated to ``None``.

If there is no other suggestion I will update later today the PR and modify the function to accept the two parameters.

jirikuncar:
> we are returning `email=None`

IMHO you don't have to add `None` values to `account_info`. The [usage](https://github.com/inveniosoftware/invenio/pull/2600/files#diff-6ffd71577ca22eab7dfaa7ad0bcc2704R51) looks safe.

jmartinm:
> IMHO you don't have to add None values to account_info. The usage looks safe.

Yep, in my case I agree, it is not needed and I will better remove it.

As the ``email=None`` was there before I am wondering if in the future somebody writes its own ``account_info`` it might happen that they pass ``email=None`` or ``email=""`` and be surprised by the signup form email not being used (if we go for the   ```data.update(account_info)```)

lnielsen:
I think the simplest is just adding the two parameters to ``oauth_register``, then there's no doubts.

jmartinm:
Updated and tested one more time ;)

PR #2601, opened 2014-12-01
OAIHarvest: fix oai harvesting for name space qualified xml

### LINKED ISSUE #2300: OAI Harvest breaks with valid XML if the tags do not match
Scenario: Harvesting projects from 'OpenAire'.
Harvestingsetup:
* Source Name: CORDIS
* Base URL: http://api.openaire.eu/oai_pmh
* Set: 'projects' (far down)
* Metadata prefix: oaf
* Postprocess convert + upload

For conversion we developed an XSL stylesheet (`cordisapi2marc.xsl`) that is working. Using 'Test' for the repo works as well from the web frontend.

`oaiharvest -v 9 -r CORDIS -d2014-05-01:2014-08-31` should give a bunch of records as can be verified via http://api.openaire.eu/oai_pmh?set=projects&verb=ListRecords&from=2014-05-01&metadataPrefix=oaf&until=2014-08-31 (there are submissions in this time frame). Similar `oaiharvest` calls against cds or one of our other repos work well without issues.

__Issue__ Invenio (tested with 1.1.3)  believes there are no updates, nothing is done.

Source of this problem seems to be in `OAI_Session()`:

    if harvested_data.lower().find('<'+http_param_dict['verb'].lower()) > -1:

This checks if the return data contains something like `<ListRecords>`. However, OpenAire returns the identifier in question as `<oai:ListRecords>`. Obviously, this it is valid to give the explicit namespace here, thus the `find()` falls short.

To check if there are no records in the set:

    if harvested_data.lower().find('<error code="noRecordsMatch">'.lower()) > -1:
 
seems more appropriate, as defined in section 3.6 of http://www.openarchives.org/OAI/openarchivesprotocol.html.

This at least solves the harvesting part and on ends up with proper content in the harvesters output files.

A similar issue, however, comes in from `remove_duplicates()`. Here again regexps are used that search only for `<ListRecords>` and not the version with an explict namespace. Thus after successful harvesting the conversoin may still fall short as the `remove_duplicates()` effectively removes everything.

Additionally, in the case of _OpenAire_ the `remove_duplicates()` even results in invalid XML:

    <?xml versio
    </ListRecords>
    </OAI-PMH>

 thus `bibconvert` called some where down the road is quite unhappy.

Still working on fiddling out what's going on in `remove_duplicates()` and how to fix it in general.

aw-bib:
OpenAIRE exposes the XML files with full name spaces. This breaks invenios harvesting routines which rely on regexp matching of the returned answers. 

Two issues arise:
1. `oai_harvest_getter.py` checks, if any records are returned by searching for a `<ListRecords>` line. This fails, if this line reads `<oai:ListRecords>` instead. The _fix_ uses the `noRecordsMatch` error return in OAI in case no records match the harvesting criteria. (cf. http://www.openarchives.org/OAI/openarchivesprotocol.html#ErrorConditions)
2. `oai_harvest_utils.py` uses a function to dedupe returned records by means of their OAI-ID. This also lives on regexping. In the OpenAIRE usecase this results in completely invalid XML, as only the first `len(<ListRecords>)` chars of the header get written at all (`header_index_end = data.find("<ListRecords>") + len("<ListRecords>")`. Secondly, all subsequent regexps fail, as they never match due to the namespace qualifier. The latter can in principle be arbitrary. The _fix_ introduces `lxml.etree` parsing of the document to extract the OAI-IDs by means of `XPath` statements. Additionally, only thos records get written to the result file that were not already done. Searching for records is also implemented by `XPath`, building of the final element uses `lxml` factories.

Combined they should address and fix #2300 also for harvesting from OpenAIRE.

@jalavik and @kaplun for review. Backported code 1.1 maint is available as well.

### DIFF
--- a/modules/oaiharvest/lib/oai_harvest_utils.py
+++ b/modules/oaiharvest/lib/oai_harvest_utils.py
@@ -118,60 +118,48 @@ def remove_duplicates(harvested_file_list):
     Usually happens when records are cross-listed across OAI sets.
 
     Saves a backup of original harvested file in: filename~
+
+    @param harvested_file_list: list of files to process and dedupe
     """
+    from copy import deepcopy
+    import lxml.etree as etree
+
     harvested_identifiers = []
     for harvested_file in harvested_file_list:
+
         # Firstly, rename original file to temporary name
         try:
             os.rename(harvested_file, "%s~" % (harvested_file,))
+        except OSError:
+            write_message("Error renaming harvested file '%s'. Skipping..." \
+                          % (harvested_file,))
+            continue
+
+        # Secondly, open target file for writing
+        try:
+            updated_harvested_file = open(harvested_file, 'w')
         except IOError:
-            write_message("Error renaming harvested file '%s': %s" % \
-                         (harvested_file, traceback.print_exc()))
-            continue
-        # Secondly, open files for writing and reading
-        original_harvested_file = None
-        try:
-            try:
-                original_harvested_file = open("%s~" % (harvested_file,))
-                data = original_harvested_file.read()
-            except IOError:
-                write_message("Error opening harvested file '%s': %s" % \
-                             (harvested_file, traceback.print_exc()))
-                continue
-        finally:
-            if original_harvested_file:
-                original_harvested_file.close()
-
-        if '<ListRecords>' not in data:
-            # We do not need to de-duplicate in non-ListRecords requests
+            write_message("Error opening harvested file '%s'. Skipping.."   \
+                          % (harvested_file,))
             continue
 
-        updated_file_content = []
-        # Get and write OAI-PMH XML header data to updated file
-        header_index_end = data.find("<ListRecords>") + len("<ListRecords>")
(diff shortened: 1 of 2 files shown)


modules/oaiharvest/lib/oai_harvest_utils.py
```diff
@@ -118,60 +118,48 @@ def remove_duplicates(harvested_file_list):
     Usually happens when records are cross-listed across OAI sets.
 
     Saves a backup of original harvested file in: filename~
+
+    @param harvested_file_list: list of files to process and dedupe
     """
+    from copy import deepcopy
+    import lxml.etree as etree
```
tiborsimko: The same comment about lxml could apply here as well as in https://github.com/inveniosoftware/invenio/pull/2608#discussion_r22522110.  However for master we can raise the lxml requirement to mandatory very easily, if wanted.  (And we would not need to play with gentle fallbacks.)  What would you prefer?

tiborsimko:
Thanks, merged as part of 565931f9a4117499c039692c96189f50a8e02b5d.

PR #2602, opened 2014-12-02
RefExtract: avoid double encoding

jalavik:
* Avoids encoding string twice when calling get_recids_matching_query()
  from BibRank in RefExtract.


tiborsimko:
Thanks, amended commit log message and merged as a9a2a44b10be3e7eb45b1fb1a7203ca47452ef6d.

jirikuncar:
NOTE check #2603 while merging master into next.

--
*PS: only to see nice reference in the other PR.*

PR #2608, opened 2014-12-03
OAIHarvest: various fixes for regex and remove_duplicates

aw-bib:
@jalavik, @jirikuncar, @kaplun: Backports fixes for OAI related to issue #2300 from invenio/master to maint-1.1. cf. pr #2601

### DIFF
--- a/modules/oaiharvest/lib/oai_harvest_daemon.py
+++ b/modules/oaiharvest/lib/oai_harvest_daemon.py
@@ -572,36 +572,47 @@ def remove_duplicates(harvested_file_list):
     """
     Go through a list of harvested files and remove any duplicate records.
     """
+    from copy import deepcopy
+    import lxml.etree as etree
+
     harvested_identifiers = []
     for harvested_file in harvested_file_list:
+
         # Firstly, rename original file to temporary name
         try:
             os.rename(harvested_file, "%s~" % (harvested_file,))
         except OSError:
-            write_message("Error renaming harvested file '%s'. Skipping.." % (harvested_file,))
+            write_message("Error renaming harvested file '%s'. Skipping..." \
+                          % (harvested_file,))
             continue
-        # Secondly, open files for writing and reading
+
+        # Secondly, open target file for writing
         try:
             updated_harvested_file = open(harvested_file, 'w')
-            original_harvested_file = open("%s~" % (harvested_file,))
         except IOError:
-            write_message("Error opening harvested file '%s'. Skipping.." % (harvested_file,))
+            write_message("Error opening harvested file '%s'. Skipping.."   \
+                          % (harvested_file,))
             continue
-        data = original_harvested_file.read()
-        original_harvested_file.close()
 
-        # Get and write OAI-PMH XML header data to updated file
-        header_index_end = data.find("<ListRecords>") + len("<ListRecords>")
-        updated_harvested_file.write("%s\n" % (data[:header_index_end],))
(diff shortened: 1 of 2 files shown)


modules/oaiharvest/lib/oai_harvest_daemon.py
```diff
@@ -572,36 +572,47 @@ def remove_duplicates(harvested_file_list):
     """
     Go through a list of harvested files and remove any duplicate records.
     """
+    from copy import deepcopy
+    import lxml.etree as etree
```
tiborsimko: This makes hard requirement on lxml which was only recommended, not mandatory.  Especially on maint-1.1 instances this could create problems.  Can you "fail gracefully" for people that do not have lxml installed?

aw-bib: Understood. Thats why I asked in chat if lxml is a good way to go.

To solve it without parsing xml would require to go with regexping again. I'm not sure it's easy to come up with a general regexp here. Is an other way of xml parsing preferable? Or do you think one should just raise an error notifying admins to get lxml as a solution?

tiborsimko: We can require lxml, but better only for master, as I mentioned in https://github.com/inveniosoftware/invenio/pull/2601#discussion_r22522187 .  Do you need this fixed for maint-1.1?  If yes, then throwing a warning "please consider installing lxml" may be acceptable, if we can mostly keep the old behaviour for those maint-1.1 that are not affected by this due to the OAI sources they use...

aw-bib: Indeed we'd need it fixed in 1.1 till 1.2 is released and we can upgrade. In fact we badly need it and applied our fix locally, via our fork.

Concerning *keeping old behaviour*: we should consider then to find a way to get a more meaningful reporting if the problem occurs. Note that we stumbled upon the error with a source we used for quite some time (cordis) due to changes they made on there end. Old behaviour doesn't even produce an error, we just silently didn't get any datasets though we adopted all XSLTs and stuff. Given that we don't expect an update from that source every day it took quite some time to see the problem. It took a while to even fiddle out that there was indeed a problem on the invenio side let alone where the problem lay (given we got valid XML).

Given that I'm not sure that adding lxml to the requirements isn't the cheaper price to pay for "the admin out there"(tm).

aw-bib: WRT throwing a warning: I'll have a look at this. I understand that I should just fail here (gracefully ;) if lxml is not installed.

WRT 1.2 I had a separate pull request for master as the code changed there a bit.

tiborsimko: > WRT 1.2 I had a separate pull request for master as the code changed there a bit.

Yes, I noticed.  If we merge this functionality into maint-1.1, then the other PR will come later.  If we don't merge this functionality into maint-1.1, then I can simply merge the other PR, and we are done.  (Because there we don't have to worry about lxml optional-or-mandatory, but we can make it mandatory.)  Hence my wondering whether you would like to include this into maint-1.1?  If yes, please try to make try/expect gentler.  If not, then I'll merge into 1.2 only, and we are done.

tiborsimko:
Thanks for the changes, tested, squashed, amended commit log message, and merged as 3d60603e08d0a4e95e79d70d37bf9107531be30a.

Issue #2611, opened 2014-12-04
alerts: remove dependency on `invenio.legacy.search_engine`


jirikuncar:
- [ ] use Jinja2 templates
- [ ] use SQLAlchemy models
- [ ] use `invenio.modules.search.api` and `invenio.modules.formatter`

jirikuncar:
WebAlerts module is going to be removed and re-implemented when new search engine is finished.

--
see also #3233

PR #2619, opened 2014-12-09
WIP alerts: port to Flask/SQLAlchemy

jirikuncar:
@kasioumis can you have a look and help us with porting the alerts engine?

--
addresses #2611
cc @hachreak for forms (depends on autocompletion field from #2615)


egabancho:
@jirikuncar should we close this one? I think WebAlert will "reborn" as an external package

PR #2628, opened 2014-12-15
WIP workflows: synchronise with new processing factory

dset0x:
* Work with the new processing factory in `workflow` module.

Naturally tests will always fail until the changes are pulled into `pu`. To test, you have to use `dset0x/workflow:delegate-storage`


invenio/modules/workflows/api.py
```diff
@@ -51,18 +57,15 @@ def worker(self):
...
             ## Let's report about broken plugins
             register_exception(alert_admin=True)
 
-    def __call__(self, *args, **kwargs):
+    def __call__(self):
         """Action on call."""
-        if not self.worker:
-            raise WorkflowWorkerError('No worker configured')
-        return self.worker(*args, **kwargs)
+        return self.worker()
```
dset0x: Hmmm, what?

invenio/modules/workflows/client.py
```diff
@@ -38,79 +31,24 @@ def run_workflow(wfe, data, stop_on_halt=False,
...
-            if stop_on_error:
-                if isinstance(exception_triggered, WorkflowError):
-                    raise exception_triggered
-                else:
-                    raise WorkflowError(
-                        message=msg,
-                        id_workflow=wfe.uuid,
-                        id_object=wfe.getCurrObjId(),
-                    )
+    # FIXME: module_name does nothing
```
jalavik: `module_name` could be part of `**kwargs` as it was useful to identify objects created with tests for example. BTW, should'nt we keep `**kwargs` still? E.g. it would pass things here to extra_data from the api call.

invenio/modules/workflows/models.py
```diff
@@ -682,16 +727,15 @@ def change_status(self, message):
...
     def get_current_task(self):
         """Return the current task from the workflow engine for this object."""
-        extra_data = self.get_extra_data()
         try:
-            return extra_data["_task_counter"]
+            return self.extra_data["_task_counter"]
         except KeyError:
             # Assume old version "task_counter"
-            return extra_data["task_counter"]
+            return self.extra_data["task_counter"]
```
jalavik: Shall we define `task_counter` and `action` as direct column types in the `DbWorkflowObject` model? They seem quite important. `action` is important for Holding Pen, `task_counter` is important for the workflow itself.

invenio/modules/workflows/client.py
dset0x: If tests are the only use-case for "updating `extra_data` while calling eng.process", then I will work around this purely in the tests' code and drop this `FIXME`.

PS I have grown very wary of `**kwargs` and `*args` (bitten too many times to count!). All arguments should have a name that declares its purpose.

invenio/modules/workflows/models.py
dset0x: >`task_counter`

Yes, it would be a good idea for the model to natively understand the `MachineState` so that we can load and dump it easily.

>`action`

If this is only used by the HoldingPen, I think it should either be in a dedicated model or we should structure `extra_data` better, eg:

```python
extra_data['holdingpen']['action']
```

invenio/modules/workflows/client.py
jalavik: Aight. Makes sense.

invenio/modules/workflows/models.py
jalavik: Sounds good in regards to task_counter.

Regarding Holding Pen, I think I would prefer that the Invenio DB model we'll use has this distinction at top-level as this is quite important for the Holding Pen/Invenio use case.

dset0x:
Regarding the apparent hiatus here: We have decided to continue development on inspire-labs/qa for a while to reveal and resolve any remaining bugs, so that this PR can be finalized.

jirikuncar:
> We have decided to continue development on inspire-labs/qa for a while to reveal and resolve any remaining bugs, so that this PR can be finalized.

@dset0x is there a reason why you can't update this PR?

dset0x:
@jirikuncar There were some bugs that arose in qa, and due to differences between that branch and invenio master, we decided it would cost too much effort to bounce the fixes between the two branches as they were happening. So we are targeting for a final push now that the problems have gone away.

Given the recent developments in #3367, I could look into getting this done even sooner.

jirikuncar:
@dset0x if you are not developing on master you are kind of blocking other people to test and review your changes easily. There was not much status report since Dec 15, 2014 except developer forum on March 23, 2015 (https://indico.cern.ch/event/376433/).

jalavik:
@jirikuncar We did some testing with this branch in real-life use-cases for INSPIRE Labs in June since it was such a big refactoring. It was based then on work done in #3136. So it has been resting with me for the last few weeks. Now that #3136 is prepared we can rebase this PR (which we actually did today while looking at #3367) and go towards integration. Sorry for the delays.

jirikuncar:
As we would like to stabilize Invenio for the release, please port it to separate *Invenio-Workflows* package. There it can get more care and reviews after we release *Invenio-Workflows v0.1.0* with the current code. Thank you!

--
cc @lnielsen @tiborsimko @jalavik

PR #2636, opened 2014-12-17
pages: global url_map modification fix

jmartinm:
* Restores the url_map after the request to avoid affecting
  subsequent requests.


jmartinm:
Updated PR, I am now using https://github.com/mitsuhiko/flask/blob/6fcc2ae3f4536e4d2651c1a00abe2306212a5bfd/flask/ctx.py#L43

Also added 2 extra commits, one with some tests and the other one to remove the info messages that were being printed too often in the console.

jirikuncar:
`Reviewed-by: me` :shipit:

Issue #2470, opened 2014-10-29
admin: index page does not work without trailing slash in Apache


jmartinm:
In an installation with Apache, accessing ```/admin``` without trailing slash gives ```403 Forbidden``` while accessing with the trailing slash works fine.

In a local installation with dev server the redirect works correctly.

Is this related to https://github.com/inveniosoftware/invenio/blob/pu/invenio/base/factory.py#L193 ?

Why are we changing the default behaviour of redirecting URLs without slash to their version with trailing slash? http://werkzeug.pocoo.org/docs/0.9/routing/#rule-format

jirikuncar:
@jmartinm is it still broken? Maybe @lovasko and @dset0x can check that it works with #2645.

jmartinm:
Yes, on our labs site https://labs.inspirehep.net/admin I can still reproduce the problem. Accessing admin without trailing slash gives me:

```
Forbidden

You don't have permission to access /admin on this server.

Apache Server at labs.inspirehep.net Port 443
```

kaplun:
@jirikuncar the issue needs to be fixed into the `pages` module which creates several files:
`filters-1.0.0.js  form-1.0.0.js`. 

There is also a strange `rediscli-1.0.0.js` that we can't figure out the real usage. Possibly Flask Admin? @hachreak, maybe you know a bit more?

PR #2645, opened 2015-01-06
WIP base: rewrite apache configuration generator

### LINKED ISSUE #2626: installation: Apache configuration
## Apache configuration
We found multiple problems with the current Apache configuration and the code that generates it:

### Apache version
We found two places, where the Apache version is used: as information to the users about their system as well as to choose the correct configuration above and below the 2.4 version. The latter is done directly by Apache `<IfVersion>` so there is no need to parse it. The other usage that was mentioned first is used only to print the version for humans (therefore no special format required). This can be either omitted, so the users will have to run the `apache2 -v` themselves, or we can keep it, but the combination of `locate`, (missing `updatedb`), `strings` and `grep` should be replaced with correctly set `$PATH` and standard Python library `subprocess` and module `re` instead of running external `grep`. Using `$PATH` has also the benefit of being correct while inside a virtualenv. In case of a custom Apache installation, user can easily add his custom path to the system-wide `$PATH`.

Example:
```python
import subprocess
import re
import sys

known_apache_binary_names = ('apache2', 'apache', 'httpd')

for binary_name in known_apache_binary_names:
    try:
        output = subprocess.check_output([binary_name, "-v"])
        break
    except OSError:
        continue
else:
    print 'No apache binary found'
    sys.exit(1)

matches = re.findall('(Apache/.*)', output)
```

### Changing the configuration benevolently
Jinja template for the Apache config that configures the `WSGIProcessDaemon` user is specifically using `config.CFG_RUNNING_AS_USER`. Before rendering the template, the code updates the config to set the `CFG_RUNNING_AS_USER` to the current user with `getuid`. Silently changing the configuration values is very misleading. Is there some special reason for this hack? If not, it might be wise to fix this properly. Otherwise, we would suggest to respect the user defined value with `inveniomanage config set`. In case of no value set, we can use the current user, but we should inform the user anyway.

### `or` fallacy and commenting
The code that sets the variables for further Jinja usage is mistakenly using this concept: `boolean and "hello %s" % value or "world %s" % value2`. The problem with the `%` operator is that it does not return a boolean value indicating the success/failure of the variable expansion, it just returns the string. The first string (the one between "and" and "or") is always non-empty and therefore _always_ evaluated to true, even if the substituted string is empty. This leads to incorrect un/commenting Apache configuration lines that deal with SSL settings.

### Template duality
Currently, for apache configuration generation, two templates exist. Both of them inherit `invenio-apache-vhost.tpl`, even though that is where the bulk of the configuration exists.

It would make the code more clear and manageable, to always render the main file, and inside it decide which blocks we need to import from the more specialised files - like SSL.

For example, the setting `AddDefaultCharset UTF-8` appears only in the HTTP configuration, while it may be used in the secure version too. Is this a bug?

### `os.path.join()`
Multiple uses of `os.sep` instead of standard `os.path.join()`. Similar thing goes for checking if file exists with `exists` instead of `isfile`.

### Configuration reading inconsistency
Inside Jinja templates used for generating the Apache configuration, the Invenio configuration object is queried at different places: either in the `apache.py` or in the `base/templates/<template_name>.tpl`. There should be a standard way to: create paths, read configuration, create lists, do
if/else logic. What is the best way to do so?

### Port conflict & wrong protocol
The default Invenio configuration contains conflicting port numbers for secure and non-secure sites (variables `CFG_SITE_SECURE_URL` and `CFG_SITE_URL`). Also, the secure URL defaults to the HTTP protocol. Is this the way to denote that there is no secure version of the site? If so, it might be wise to consider a separate configuration variable.

@lovasko & @dset0x

dset0x:
* The templates were merged into one, and so was the resulting
  configuration file. Should the user need to enable or disable SSL for
  example, the recommended way is to edit the invenio configuration and
  re-run `create-config`.

* All variables that are used in the template must now pass through
  `apache.py`, even if no operation is performed on them. All logic
  that does not logically belong in Jinja2 has been removed from the
  template.

* Apache version 2.2 OpenSSL 1.0.1e are now the minimum supported. These
  are the official CentOS 6.6 packages.

* mod_headers is now necessary. So is socache_shmcb for apache >= 2.3.3
  in order to use SSLStaplingCache.

* SSL allowed protocols and ciphers have been set according to mozilla's
  recommendations.

* The switches for `inveniomanage apache create-config` have changed:
  `--force` was dropped as it was a NO-OP.

* It is now possible to manually set the paths to the desired
  certificates, as the following configuration directives have been
  ported from zenodo: 'APACHE_CERTIFICATE_PEM_FILE',
  'APACHE_CERTIFICATE_FILE', 'APACHE_CERTIFICATE_KEY_FILE'.

* `invenio/legacy/inveniocfg.py:_detect_ip_address` exception handling
  is improved and it now uses 'google.com' instead of 'invenio.org' to
  test connectivity, because of their uptime.

Fixes #2626


invenio/legacy/inveniocfg.py
```diff
@@ -832,9 +832,9 @@ def _detect_ip_address():
     """
     try:
         s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.connect(('invenio-software.org', 0))
```
jirikuncar: I would keep the `invenio-software.org`

jirikuncar:
@dset0x it looks good. If you are done, please remove `WIP` from the PR title, improve PEP8/257, rebase and re-push. Thanks

invenio/legacy/inveniocfg.py
dset0x: I understand this is not really part of this patch, but I expect `google.com` to have a greater uptime. Perhaps the issue is mentioning google in the code?

jirikuncar: > Perhaps the issue is mentioning google in the code?

`invenio-software.org` should be up and running if you want to install demosite hence it should be fine for this case too.

dset0x: I may be missing something here. What is the relationship between `invenio-software.org` being up and someone creating the apache configuration for their website (it doesn't have to be the demosite).

jirikuncar: Do not mention google in the code.

invenio/base/scripts/apache.py
```diff
@@ -17,194 +17,414 @@
...
+
 from invenio.ext.script import Manager, change_command_name
+from invenio.ext.template import render_template_to_string
+from invenio.utils.text import wrap_text_in_a_box
+
+
+try:
+    from shutil import which
+except ImportError:
+    from distutils.spawn import find_executable as which
```
jirikuncar: Please check and amend `invenio.utils.shell:which`. Thank you

invenio/base/scripts/apache.py
```diff
@@ -17,194 +17,414 @@
...
-    """ % '\n\n'.join(tuple(map(
-        lambda x: "Include " + apache_conf_dir.encode('utf-8') + os.sep + x,
-        list(conf_files[:1 if no_ssl else 2]))))))
-    print(">>> Apache conf files created.")
+                             """.format(apache_vhost_file)))
 
 
 def main():
     from invenio.base.factory import create_app
-    app = create_app()
```
jirikuncar: Do you have any reason for changing these 2 lines? If there is serious reason we should consider amending other manage scripts too.

jirikuncar:
Please add `NOTE` to commit message with explanation how to migrate current templates. (cc @lnielsen )

jirikuncar:
@dset0x when you are done can you try to enable Apache tests in [`.travis.yml`](https://github.com/inveniosoftware/invenio/blob/pu/.travis.yml#L92)? Thank you

dset0x:
@jirikuncar I have yet to see this run once:

@lovasko Any progress with testing this? Are there any tests other than `wget -O /dev/null http://localhost` you would like to see ran?

jirikuncar:
> Are there any tests other than `wget -O /dev/null http://localhost` you would like to see ran?

That one should be enough for now.

invenio/base/scripts/apache.py
```diff
@@ -17,194 +17,414 @@
...
+                'enabled': whether to enable SSO,
+                'login_url': SSO login URL
                 }
+            """
+            sso = {
+                'enabled': app.config['CFG_EXTERNAL_AUTH_USING_SSO'],
+                'login_url': app.config.get('SSO_LOGIN_URL', None)
+            }
+            if sso['enabled']:
+                assert sso.login_url is not None, 'SSO enabled by URL not set'
```
dset0x: s/by /but /

invenio/base/scripts/apache.py
dset0x: I use `from flask import current_app as app` at the top of the file for the sake of typing less during the rest of the script, as I use it a lot. I wanted to make it clear that they are different. Therefore, no reason to propagate this change.

invenio/base/scripts/apache.py
```diff
@@ -17,194 +17,414 @@
...
+import pwd
+import re
+import sys
+from distutils import dir_util
+from os import path
+
+import pkg_resources
+import shutil
+import subprocess
+from flask import current_app as app
```
jirikuncar: please use `current_app` for consistency

- `current_app` - proxy to application object in current app context
- `app` - usually an instance of `Flask`

invenio/base/scripts/apache.py
```diff
@@ -17,194 +17,414 @@
...
-                'listen_directive_needed': listen_directive_needed,
-                'aliases': aliases,
+    class ApacheConfig(object):
+
+        """HTTP configuration resolver."""
+
+        def __init__(self):
+            """Initialize easily resolvable values."""
+            # Context
+            self._known_cert_directives = (
```
jirikuncar: IMHO `_known_cert_directives` and `_candidate_paths` would look better as class attributes  `KNOWN_CERT_DIRECTIVES` and `CANDIDATE_PATHS`.

dset0x:
Rebased ontop of #2809

dset0x:
#2809 merged. Rebased ontop of  #2818

jirikuncar:
@dset0x when you are done please remove WIP from the title, rebase and re-push.

dset0x:
@jirikuncar If all goes well, testing should be finished today.

jirikuncar:
> *8 days ago:* @jirikuncar If all goes well, testing should be finished today.

What did go wrong?

--
I'm changing milestone to 2.1.

dset0x:
> What did go wrong?
 #2818 happened. I am also waiting for @lovasko to write migration instructions.

jirikuncar:
@lovasko is there any progress (https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-77371714)?

lovasko:
@jirikuncar it is actively being worked on. The migration guide will be finished today :)

jirikuncar:
>> @jirikuncar If all goes well, testing should be finished today. (https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-75938186)

> The migration guide will be finished today :) (https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-84020920)

Cool :+1:

jirikuncar:
Btw how does it look like with the travis configuration mentioned in https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-75204219?

lovasko:
I ran the `inveniomanage apache create-config` along with the rest of the commands present in the Travis configuration in my `master` Docker image (based on Ubuntu 14.10) with some additional `CFG_APACHE...` settings and I was able to browse the Atlantis page served by Apache. After merging this, the `.travis.yml` file should be changed accordingly, possibly using the settings from my Docker image. 

There are other issues with the Travis configuration, e.g. creating the snake oil certificate is incorrect - it assumes that the second argument is the output path, but in reality the output file is created in a static location that is not configurable (because `generate-default-snakeoil` is not a template, but rather a subcommand, see [man page](http://man.he.net/man8/make-ssl-cert)) and should be moved or copied to a correct location or simply linked. 

Do you want it to be addressed in this issue or in a separate one?

lovasko:
@jirikuncar regarding the migration use: who is the target audience for such document/text?
 * administrators of Invenio instances
 * overlay authors
 * Invenio developers

please select groups or add missing ones.

jirikuncar:
> Do you want it to be addressed in this issue or in a separate one?

It would be nice to present working solution including working example for Travis CI.

>  Who is the target audience for such document/text?

All of them. How difficult is the migration? Isn't it just `invenomanage apache create-config && service apache reload`?

lovasko:
> How difficult is the migration?

As mentioned in the issue description, there are some bumps in the minimal version supported, some required Apache modules or `CFG_APACHE_...` variables that need to be added. These are things that relate to administrators. 

Developers might be interested into changes in `apache.py`, while overlay authors probably care about merging the template files into single one.

dset0x:
Regarding the relationship of this PR with #2470: The problem is not strictly on the apache side. This is because there is no single entry point for static files: We try to serve whatever may be there before asking WSGI if it can serve some URL. Ideally we would have a `static/` folder and tell the webserver to serve that and only that as static.

In short there is nothing we can do to correctly fix the issue on the server front.

dset0x:
@dset0x don't forget to rebase to the correct branch that fixes `CFG_PREFIX`

invenio/base/scripts/apache.py
jirikuncar: Please these 2 lines as before:
```
app = create_app()
manager.app = app
```
Thank you!

jirikuncar:
> Ideally we would have a `static/` folder and tell the webserver to serve its contents and only its contents as static.

We can do it when we port *ALL* legacy module and we use only `url_for('static', filename='foo.png')` or `url_for('<blueprint>.static', filename='bar.png')`.

In order to fix #2470 we can create one extra redirect `/admin$ -> /admin/`.

@lovasko where can we find the migration guide (https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-84021272)?

invenio/base/scripts/apache.py
dset0x: Sorry, I missed that! Thanks for going over this code again.

dset0x:
I've added a dummy commit with migration instructions in the new format. @jirikuncar let me know if it looks okay style-wise, so that I can start squashing things together.

dset0x:
Migration notes finalized. This PR now assumes that #2829 gets merged (eg it does not access `CFG_PREFIX` anymore), so consider it blocked until that's done.

jirikuncar:
@lovasko @dset0x what is the status?

- [ ] test in `.travis.yml` (https://github.com/inveniosoftware/invenio/pull/2645#issuecomment-75208502)
- [ ] documentation + `INSTALL.rst`
- [ ] rebase and re-format the commit message (indentation, no past tense, etc.)

--
cc @kaplun @jmartinm @jalavik

dset0x:
> config/invenio.conf does not exists anymore and it's mentioned

@jirikuncar  All the comments in that file served as documentation for the variables. Where have they gone?

jirikuncar:
> All the comments in that file served as documentation for the variables. Where have they gone?

They should be moved to docstrings in `invenio.modules.<module>.config` or `invenio.base.config`.

invenio/base/scripts/apache.py
```diff
@@ -17,188 +17,427 @@
...
+                try:
+                    xsfpdirs[key] = app_cfg[val]
+                except KeyError:
+                    print(">>> Variable {} not set. Skipping.."
+                          .format(val), file=sys.stderr)
+                    del xsfpdirs[key]
+
+            xsfpdirs.update({
+                'attachfile': path.join(app_cfg['CFG_TMPDIR'], 'attachfile'),
+                'comments': path.join(app_cfg['CFG_DATADIR'], 'comments'),
```
jirikuncar: Where is `CFG_DATADIR` set?

jirikuncar:
@dset0x I have few "style" comments:
- `Please consult `config/invenio.conf` for their use.` - you should probably point to docs.
- Join `INCOMPATIBLE` bullet point about Apache and make the lists of variables more compact (e.g. `APACHE_{LISTEN_DIRECTIVE_HTTP, LISTEN_DIRECTIVE_HTTPS, ...}`).
- Replace `s/CFG_APACHE_/APACHE_/` to follow new naming conventions.
- Please check PEP{8,257} in new files and remove WIP **before** final re-push.

and one suggestion for future improvement:
- move all `apache` configuration and scripts to `invenio/ext/apache/{config, script, templates}`

invenio/base/scripts/apache.py
dset0x: Woops. Very bad cherry picking followed by leaving my prompt like [1] this and then going home.
The variables are now back.

```
Enter passphrase for key '/home/gs-sis/.ssh/id_rsa': 
Enter passphrase for key '/home/gs-sis/.ssh/id_rsa': 
```

invenio/base/config.py
```diff
@@ -601,7 +637,7 @@
 CFG_SITE_NAME = "Atlantis Institute of Fictive Science"
 CFG_SITE_RECORD = "record"
 SECRET_KEY = "change_me"
-CFG_SITE_SECURE_URL = "http://localhost:4000"
+CFG_SITE_SECURE_URL = "http://localhost:4040"
```
jirikuncar: You are changing default configuration here - please keep port 4000. This change should be discussed separately.

invenio/base/config.py
```diff
@@ -152,7 +168,27 @@
 CFG_ACCESS_CONTROL_NOTIFY_USER_ABOUT_NEW_ACCOUNT = 1
 CFG_ADS_SITE = 0
 CFG_APACHE_GROUP_FILE = "demo-site-apache-user-groups"
+CFG_APACHE_GUESS_CERT_PATHS = 0
```
jirikuncar: please change prefix of all new variables to `APACHE_`

invenio/base/scripts/apache.py
```diff
@@ -17,188 +17,427 @@
...
 
-@manager.option('-f', '--force', dest='force', action='store_true')
-@manager.option('--no-ssl', dest='no_ssl', action='store_true')
+@manager.command
 @change_command_name
-def create_config(force=False, no_ssl=False):
-    """
-    Create Apache configuration files for this site, keeping previous
-    files in a backup copy.
+def create_config():
```
jirikuncar: I would keep the options there with following code:
```python
if force or no_ssl:
    warnings.warn('Options --force and --no-ssl has been removed.', Invenio21Deprecated)
```

invenio/base/scripts/apache.py
```diff
@@ -17,195 +17,508 @@
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
+"""Apache configuration generator.
+
+This script relies on the following variables:
+
+    Note: The variables are followed by their defaults.
```
jirikuncar: see an example how to document config variables: http://invenio.readthedocs.org/en/latest/ext/arxiv.html?highlight=config#configuration-settings, https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/arxiv/__init__.py#L29-L44

jirikuncar:
@dset0x can you run test on travis using copy of the branch in your repo not to block the queue. Thanks

dset0x:
@jirikuncar I registered on travis and enabled the dset0x/invenio repository. Would that be all I should have to do?

jirikuncar:
yes

dset0x:
This PR has been stalled again due to not being able to pass its test on travis (regardless of it being temporarily marked as passing).

It seems that the travis installation has some issue with its python libraries that causes  [this error](https://gist.github.com/dset0x/a7927429ed1c4312ec19) that I have been unable to solve.

Regardless, this works locally. If someone would like to take a look at the problem, please do. You can find a lot more output in the build logs.

cc @jirikuncar @lovasko @jmartinm @kaplun

kaplun:
BTW seems by now the travis issue solved again on its own...

dset0x:
@kaplun Thanks for the heads up, but how did you know? Was this `openssl_md_meth_names` issue manifesting itself elsewhere too?

kaplun:
Ah nope sorry, I simply see that Travis succeeds right now. Also around on the web I have seen this was related to a temporary problem... Wasn't this PR blocked by the fact that Travis was not passing?

dset0x:
> I simply see that Travis succeeds right now.

I have edited `travis.yml` to continue after `wget` fails so that I can have it print some debugging information afterwards: 

> (regardless of it being temporarily marked as passing)

------

> Wasn't this PR blocked by the fact that Travis was not passing?

Yes, but there is also a new problem now: https://github.com/inveniosoftware/invenio/pull/3077#issuecomment-99063228 . Unfortunately I have not gotten around to working towards some RFC about it, due to working on other modules.

jirikuncar:
There has been more than one month of inactivity. Shall we close this PR as someday?

dset0x:
@jirikuncar I understand that this functionality will be moved to the demosite anyway (as I see we no longer want it to be tested in invenio itself (https://github.com/inveniosoftware/invenio/pull/3371)). If you prefer this to be closed for organizational reasons until/if it works, feel free.

jirikuncar:
@dset0x the functionality of generating Apache configuration stays in Invenio. It's just hard to review your changes as the codebase is evolving a lot and you are not updating this PR.

dset0x:
@jirikuncar This PR is not awaiting review - it is not working: There is commentary about the issues above.

PR #2651, opened 2015-01-08
JSONAlchemy: any indicator support

### LINKED ISSUE #1722: jsonalchemy: Invenio should accept any indicator as valid for any tag
Originally on 2014-03-28

A majority of Marc21 tags accept different values for indicators.  Currenty Invenio just uses a restricted subset, mostly empty or undefined values.

When valid Marc21 records from other sources are imported to Invenio, most of the fields are not recognized, because don't fall in this restricted subset.  For example, a title with valid tags and subfields, like:

 245 14 $a The Merchand of Venice

is not recognized as such: it is not displayed, not indexed, not exported.

Given the amount of combinations in Marc21, accepting any indicator as valid is a much better solution than the current behaviour.

The sooner this behaviour is corrected, the better.

kaplun:
* Introduces support for accepting MARC fields having any indicator
  (closes #1722)

* Fixes the definition of time_and_place_of_event_note,
  series_statement and source_of_description fields.

Reported-by: Ferran Jorba <Ferran.Jorba@uab.cat>

### DIFF
--- a/invenio/modules/authors/recordext/fields/author_base.cfg
+++ b/invenio/modules/authors/recordext/fields/author_base.cfg
@@ -38,7 +38,7 @@ ids:
                   'required': True},
                   'value': {'type': 'string', 'required': True}}}}
     creator:
-        marc, "035__", {'type': value['9'], 'value': value['a']}
+        marc, "035..", {'type': value['9'], 'value': value['a']}
     producer:
         json_for_marc(), {"035__9": "type", "035__a": "value"}
 
@@ -65,7 +65,7 @@ name:
          'schema': {'last': {'type': 'string', 'empty': False},
                     'first': {'type': 'string', 'empty': False}}}}
     creator:
-        marc, "100__", {'last': util_split(value['a'], ', ', 0),
+        marc, "100..", {'last': util_split(value['a'], ', ', 0),
                         'first': util_split(value['a'], ', ', 1)}
     producer:
         json_for_marc(), {"100__a": "last + ', ' + first"}
@@ -80,6 +80,6 @@ publications_list:
         {'publications_list': {'type': 'list', 'force': True,
          'required': True}}
     creator:
-        marc, "900__", int(value["a"])
+        marc, "900..", int(value["a"])
     producer:
         json_for_marc(), {"900__a": ""}
--- a/invenio/modules/jsonalchemy/testsuite/fields/authors.cfg
+++ b/invenio/modules/jsonalchemy/testsuite/fields/authors.cfg
@@ -24,7 +24,7 @@ _first_author, creator:
                 ("100__h", "CCID"),
                 ("100__i", "INSPIRE_number"),
                 ("100__u", "first author affiliation", "affiliation"))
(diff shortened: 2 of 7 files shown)


invenio/modules/authors/recordext/fields/author_base.cfg
```diff
@@ -65,7 +65,7 @@ name:
          'schema': {'last': {'type': 'string', 'empty': False},
                     'first': {'type': 'string', 'empty': False}}}}
     creator:
-        marc, "100__", {'last': util_split(value['a'], ', ', 0),
+        marc, "100._", {'last': util_split(value['a'], ', ', 0),
```
jalavik: Whats the difference between `._`, `_.`, `..` and `__`?

kaplun: To my understanding: `.` is wildcard for *any character*. `_` means instead *no character*.

The fact that you have `._` is because in MARC you have two indicators: in this case the first one can be anything while the second one need to be empty.

jirikuncar: > Whats the difference between `._`, `_.`, `..` and `__`?

:black_joker: @jalavik  are we talking morse code - `A` `N` `I` and `M` :wink: ?

kaplun:
@egabancho do you think this is going to implement what @fjorba asks for? Do I just need to amend the indicator definitions in the `marc`-`creator` part? Or do I have to amend also the `@legacy` part? If yes, how?

fjorba:
Please, please, please...  Let both indicators be anything.  Please.  It is much easier for everybidy.  A 9 is a (local) valid indicator.  So anything is better that strict values.

invenio/modules/authors/recordext/fields/author_base.cfg
egabancho: @kaplun the values that you can put in there are python regular expressions, so you can put something like `100[013]_` if you want to restrict the first indicator to some values.

kaplun:
@fjorba sure. In my pull request I am not requiring a particular value for an indicator. When MARC defines some values for a given indicator I have amended the configuration to accept **any** value. However if MARC does say: "undefined" then I have left the indicator as empty. Are you proposing we should not even care about that and simply amend the whole configuration to accept everywhere both indicator1 and 2 regardless of MARC standard?

invenio/modules/authors/recordext/fields/author_base.cfg
jalavik: @jirikuncar Hehe. 10 points! :fireworks:

fjorba:
> @kaplun: Are you proposing we should not even care about that and simply amend the whole configuration to accept everywhere both indicator1 and 2 regardless of MARC standard?

Yes, yes!  That's exacty it!  Any value in any indicator in any tag.  Follow the Postel Principle:

> Be conservative in what you do, be liberal in what you accept from others.

Let me be your friendly librarian (yes, I do have a librarianship academic title!).

PXke:
I definitely agree with @fjorba, could it be reflected in this pull request ?

kaplun:
@jirikuncar  I applied the kwalitee remarks.

jirikuncar:
@kaplun similar fixes would be nice also for `invenio-demosite` package.

Issue #2578, opened 2014-11-25
BibCirculation: problem with 'Edit physical copies'


shesselbach:
When accessing the 'Item details' page of an item by e.g. pressing the 'Edit physical copies' button in bibedit a '500  Internal Server Error' is shown and an error mail is sent detailing an exception 
'HTTPError: HTTP Error 400: Bad Request' from urllib2 originating from the `BIBCIRCULATION_OPENER.open` call in the `get_book_cover(isbn)` function in bibcirculation_utils.py, lines 104ff. This call tries to retrieve a book cover for the item and is not essential for the functionality. (In Invenio 1.0 this problem did not arise because in the underlying code urllib instead if urllib2 had been used which throws no HTTPError exception.)

In the master branch this problem seems to be fixed by putting the BIBCIRCULATION_OPENER call into """ """, lines 179ff. A possible solution might be to back port this into the maint-1.1 branch.

I was able to get the BibCirculation working in our 1.1.4 test installation by porting the get_book_cover() function from master, i.e.
`"""cover_xml = BIBCIRCULATION_OPENER.open(...`
`...`
`...                 'Keywords=' + isbn)"""`
`cover_xml=""`
`...`
`except:`  (instead of `except AttributeError:` important!)

tiborsimko:
@shesselbach Indeed this works well is master.  Can you please submit your maint-1.1 patch as a pull request?

PR #2653, opened 2015-01-08
Backport of the fix in the master branch into the maint-1.1 branch related to issue #2578

shesselbach:
Prevent an 'HTTPError: HTTP Error 400: Bad Request' from urllib2
when accessing the 'Item details' page which originated from the
BIBCIRCULATION_OPENER.open call in the get_book_cover(isbn) function.

### DIFF
--- a/modules/bibcirculation/lib/bibcirculation_utils.py
+++ b/modules/bibcirculation/lib/bibcirculation_utils.py
@@ -101,20 +101,20 @@ def get_book_cover(isbn):
     from xml.dom import minidom
 
     # connect to AWS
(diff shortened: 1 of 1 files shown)


tiborsimko:
Thanks for the backport, I'll merge the code, so we'll have maint-1.1 and master code equivalence, but we should improve it later...  E.g. I saw related issue that the record editor does not report "correct" number of copies in the UI.  I'll open a new issue about it.

tiborsimko:
Thanks, amended commit log message and merged as 347a100eb2b8cab96c965a8ee87f9851df70acfb.

Issue #11, opened 2014-04-30
use specific jQuery versions


tiborsimko:
Originally on 2010-04-20

One more thing to do before 1.0 in the jQuery department: there are
still some svn/tags/latest parts in install-jquery-plugins.  This
is too bleeding edge, and we have been bitten by this in the past, so
to speak, e.g. file renames in
0c841f12e403c0bdb9cc38dd9a7cfbd2d3e4ff45, e.g.  calendar changes in
e2a978e7746fb2a929e58c23034f4e216bf6afe2.

So it would be good to go through the whole install-jquery-plugins
target and change remaining parts in order to wget always some very
specific version of a jQuery library, as is needed, like in my changes
mentioned above.  (I have been doing that as was necessary for
merging, I have not gone systematically through every jQuery
dependency we have.)

Can you please look at that?

If there is a trouble with some dependency lib URL stability, or
versions are lacking on the remote site, then we can host the
dependency on cdsware.cern.ch site, like the other installation files,
e.g. demo records, e.g. keyword ontologies.

aw-bib:
Looking into the `Makefile`'s section related to this issue as we recently had a problem with some plugin. Note alos that ther're still some wgets without a forced -O as mentioned by @kaplun.

Unfortunatley, it is a bit beyond me which versions are really required. Probably, someone in the know could have a look at it?

Here is what I found in maint-1.1 as plugins without explicit version:

----------------------------------------------------------------------
* wget https://raw.github.com/malsup/form/master/jquery.form.js --no-check-certificate

Last tagged version on github is v3.19:
https://github.com/malsup/form/blob/v3.19/jquery.form.js

However, one may get 3.50 from jsDelivr:
http://origin.jsdelivr.net/jquery.form/3.50/jquery.form.js

----------------------------------------------------------------------
* wget http://jquery-multifile-plugin.googlecode.com/svn/trunk/jquery.MultiFile.pack.js

Latest version is r54:
https://jquery-multifile-plugin.googlecode.com/svn-history/r54/trunk/jquery.MultiFile.js

----------------------------------------------------------------------

* wget http://www.datatables.net/download/build/jquery.dataTables.min.js

current stable is 1.10.4
wget http://cdn.datatables.net/1.10.4/js/jquery.dataTables.min.js

Note: there is a further wget for
https://raw.githubusercontent.com/DataTables/DataTables/1.10.0/media/css/demo_table_jui.css
that looks related.

----------------------------------------------------------------------

[...]

PR #2655, opened 2015-01-08
installation: explicitly specify versions for jquery plugins

aw-bib:
Explicitly specify versions for jquery plugins:
* jquery.form.js
* jquery.MultiFile.js
* jquery.dataTables.min.js
Add `-O` parameters for all wget commands within jquery installations to avoid piling up files.
Dump `jquery.min.js` to the proper name right away.
Add `-u` to `unzip` calls to avoid queries in case file exists

*TODO*: `json2.js` does not provide any version tags on github. This is the only one that still comes in the latest version.

This addresses issue #11 for maint-1.1

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -143,19 +143,19 @@ install-jquery-plugins:
 	mkdir -p $(prefix)/var/www/css
 	(cd ${prefix}/var/www/js && \
 	wget -O jquery.min.js http://invenio-software.org/download/jquery/jquery-1.7.1.min.js && \
-	wget -O jquery-ui.min.js http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.17/jquery-ui.min.js && \
-	wget -O jquery.jeditable.mini.js http://invenio-software.org/download/jquery/v1.5/js/jquery.jeditable.mini.js && \
-	wget -O jquery.form.js https://raw.githubusercontent.com/malsup/form/3.51/jquery.form.js --no-check-certificate && \
-	wget -O jquery.MultiFile.js http://jquery-multifile-plugin.googlecode.com/svn-history/r54/trunk/jquery.MultiFile.js && \
+	wget -N http://ajax.googleapis.com/ajax/libs/jqueryui/1.8.17/jquery-ui.min.js && \
+	wget -N http://invenio-software.org/download/jquery/v1.5/js/jquery.jeditable.mini.js && \
+	wget -N https://raw.githubusercontent.com/malsup/form/3.51/jquery.form.js --no-check-certificate && \
+	wget -N http://jquery-multifile-plugin.googlecode.com/svn-history/r54/trunk/jquery.MultiFile.js && \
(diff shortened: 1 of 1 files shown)


aw-bib:
@tiborsimko I added the `-O` to address the point mentioned by @kaplun as well as your comment in the original issue (piling up `foo.1`, `foo.2`, `foo.3`) which happens in case of subsequent install calls without intermediary cleanup. We actually observed a problem here during install where people installed a new version using the install-call, but cause it was named `zip.1` while the old `.zip` it didn't get unpacked. So they believed to be on the correct version but weren't. (Not easy to find without access to the shell.)
However, of course I see you point with the version number in two places. I could go into the code a bit more to do something like `wget -O jquery.bookmark.package.zip http://invenio-software.org/download/jquery/jquery.bookmark.package-1.4.0.zip` and then unpack the *un*versioned zip. Otherwise have to fix version numbering in at least two places anyway as the unzip calls also run against versioned packages. Would that be "best of two worlds"?
One could even go so far as to refactor the code into something along the lines of
     export jquery_bookmark_package='jquery.bookmark.package-1.4.0.zip'
     wget -O $jquery_bookmark_package http://invenio-software.org/download/jquery/$jquery_bookmark_package
     unzip -u $jquery_bookmark_package
which would allow to move all version informations to one block in front of the fetch/unpack routines. This is something I usually find handy in other scripts I do for similar issues. I just didn't want to change to much in the code. That's why I left this out. 
Anyway, if I go into this, I admit, that I'd also move the unpacking routines either at the end of the queue after all downloads where done or at directly after the download as such. Right now they are a bit scattered around and in some cases it's not yet clear when a zip get's unpacked as ther're several `wget` in between. I could do this as well if you like the idea.The diff would, however, then get a bit larger, basically with all proposed the whole block would be new.
BTW: please note that I've the fixes ready for master as well, just didn't make the PR yet. I'll keep that here till you drop me a note how you like it best.

tiborsimko:
> [...] which happens in case of subsequent install calls without intermediary cleanup

Ah, I see, yes, if people do not call "uninstall" explicitly, then bad things would happen.

>  I could go into the code a bit more to do something like [...]

A simpler option would be to use `wget -N http://example.org/foo.zip`. This switches on the timestamp checking, and if the remote file is newer, it will download and override any local file there may be.

It is probably not worth the trouble to make any more advanced changes; we can keep your branch as is, and eventually change some of the statements to use `-N` where appropriate (e.g. when we keep versions.)  

(In Invenio 2 this is solved by other means anyway.)

> BTW: please note that I've the fixes ready for master as well, just didn't make the PR yet. I'll keep that here till you drop me a note how you like it best.

Perfect; please make a PR in the same form as you have.

tiborsimko:

Thanks, rebased against `maint-1.1`, squashed, amended commit log message, and merged as bdf145891b5f44044bcbe44f7c7013aeb4af805f.

Issue #1613, opened 2014-04-30
installation: check fixtures WRT important master->next differences


tiborsimko:
Originally on 2013-11-04

The demo site fixtures should be proof-read WRT master->next differences.

In September 2013, I had to amend Invenio/next collection fixtures in 4bfd345ea341282cae02c473196f9e30af788da3 regarding:

```
[...]

     class preprints:
         id = 2
         name = 'Preprints'
-        dbquery = 'collection:PREPRINT'
+        dbquery = '980:"PREPRINT"'
         names = {
             ('en', 'ln'): u'Preprints',
             ('fr', 'ln'): u'Prétirages',

[...]
```

Looking at Invenio/master branch, the collection>980 change happened already in March 2012 in 216f08227405af282a62e3ed88e7198f7d0a7e79

This means that it is not sure that changes done to Invenio/master demo site as far back as February 2012 were all well propagated to Invenio/next.  We'd better check this for all modules.

PR #2664, opened 2015-01-13
WebSearch: removes hardcoded collection requirement

### LINKED ISSUE #2592: BibCirculation: holdings do not display due to hardcoded collection requirement
In websearchadminlib, around line 3527 (https://github.com/inveniosoftware/invenio/blob/master/modules/websearch/lib/websearchadminlib.py#L3527) one finds the following code:

```
#Disable holdings tab if collection != Books
collection = run_sql("""select name from collection where id=%s""", (colID, ))
if collection[0][0] != 'Books':
    tabs['holdings']['enabled'] = False
```

This effectively hides the _contents_ of the _holdings_ tab in case the collection is not called `Books` regardless of the status of the tab itself. That is, the tab can well be visible, but still will not contain anything.

I would suggest, as already stated in the same module some lines higher up, to remove this block or (as @tiborsimko suggested) at least hide it in an if CFG_SOMESITE block if needs be. OTOH, if one doesn't want to have holdings to be displayed one can (now) easily disable the tab itself through websearchadmin interface, so there seems to be no reason to explicitly hide the contents but still display the tab.

Additionally, many library catalogues handle contents beyond books, so the naming is also a bit counterintuitive. (Something like OPAC would be nearer the mark.)

Please strongly consider also backporting this fix to 1.1 maint.

blixhavn:
* Removes hardcoded requirement that only allows the 'holdings'
  tab to be displayed for the 'Books' collection (closes #2592)

* Adds an upgrade file that can be run on any installation to
  reflect this in the database (i.e. adds a new rule to show all
  but 'holdings' for every collection without a preexisting rule)

* Adds insert statements in the demo SQL file to reflect this.

Signed-off-by: Oystein Blixhavn <oystein@tind.io>

### DIFF
--- a/modules/miscutil/lib/upgrades/invenio_2015_01_13_hide_holdings.py
+++ b/modules/miscutil/lib/upgrades/invenio_2015_01_13_hide_holdings.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2014 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation,
+# Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Upgrade for explicitly hiding the 'holdings' tab for all collections\
+ except 'Books' without an existing rule"""
+
+from invenio.dbquery import run_sql
+
+depends_on = ['invenio_release_1_1_0']
+
+
+def info():
+    """Upgrade recipe information."""
+    return "Updates the collectiondetailedrecordpagetabs to hide 'holdings' for every collection \
+            except 'Books' without an existing rule."
+
+
+def do_upgrade():
+    """Upgrade recipe procedure."""
+    # Show holdings for the 'Books' collection unless there's an existing rule
+    run_sql("""
+    INSERT IGNORE INTO collectiondetailedrecordpagetabs (id_collection, tabs)
+    SELECT id, "files;references;keywords;plots;hepdata;holdings;comments;linkbacks;citations;usage;metadata"
+    """)
+
+    # Insert the rest of the rules
+    run_sql("""
+    INSERT IGNORE INTO collectiondetailedrecordpagetabs (id_collection, tabs)
+    SELECT id, "files;references;keywords;plots;hepdata;comments;linkbacks;citations;usage;metadata"
+    FROM collection
+    """)
+
+
+def estimate():
+    """Upgrade recipe time estimate."""
+    return 1
--- a/modules/miscutil/lib/upgrades/invenio_2015_01_13_hide_holdings.py
+++ b/modules/miscutil/lib/upgrades/invenio_2015_01_13_hide_holdings.py
@@ -37,6 +37,7 @@ def do_upgrade():
     run_sql("""
     INSERT IGNORE INTO collectiondetailedrecordpagetabs (id_collection, tabs)
     SELECT id, "files;references;keywords;plots;hepdata;holdings;comments;linkbacks;citations;usage;metadata"
+    FROM collection WHERE name = 'Books'
     """)
 
     # Insert the rest of the rules
--- a/modules/miscutil/demo/democfgdata.sql
+++ b/modules/miscutil/demo/democfgdata.sql
@@ -96,7 +96,6 @@ INSERT INTO collectiondetailedrecordpagetabs VALUES (29, 'files;references;keywo
 INSERT INTO collectiondetailedrecordpagetabs VALUES (30, 'files;references;keywords;plots;hepdata;comments;linkbacks;citations;usage;metadata');
 INSERT INTO collectiondetailedrecordpagetabs VALUES (31, 'files;references;keywords;plots;hepdata;comments;linkbacks;citations;usage;metadata');
 INSERT INTO collectiondetailedrecordpagetabs VALUES (32, 'files;references;keywords;plots;hepdata;comments;linkbacks;citations;usage;metadata');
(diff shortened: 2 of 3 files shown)


modules/miscutil/lib/upgrades/invenio_2015_01_13_hide_holdings.py
```diff
@@ -0,0 +1,53 @@
...
+depends_on = ['invenio_release_1_1_0']
+
+
+def info():
+    """Upgrade recipe information."""
+    return "Updates the collectiondetailedrecordpagetabs to hide 'holdings' for every collection \
+            except 'Books' without an existing rule."
+
+
+def do_upgrade():
```
jirikuncar: I'm not sure about the upgrade recipe as one is upgrading only demo data.

tiborsimko: @jirikuncar Writing upgrade recipe is good in this case, because the removed code would have visible effect even on non-demo installations, which is not wanted.  

@blixhavn However we need to be more careful here.  If people set something for the Home collection, we should reuse the same settings for daugher collections, or leave them untouched.  The current code seems to force Atlantis settings on daugther collections even on those sites where people customised the Home collection settings, without customising daugter collections, which is perhaps not probable but which can occur. When dealing with collection Ci, we should either (1) traverse the collection tree and pick closent parent collection Cj's settings for Ci, or (2) leave the collection Ci settings out, so that the home collection settings would be picked.  (Have not tested, just read the diff.)  And you can end up the upgrade recipe with print warning for people to check out their collection tab settings in the WebSearch Admin interface.

blixhavn: In Invenio, when applying settings to subcollections, an entry is made into the database for each subcollection as well as the main collection. I.E. every collection (demo or custom) that has any settings other than default (show all) will not be touched by this upgrade script, due to the `GNORE` statement in the SQL query which will avoid inserting rules for collection IDs already found in the `collectiondetailedrecordpagetabs` table.

tiborsimko: @blixhavn Yes, I mentioned that my concern "is perhaps not probable but which can occur".  Imagine a site having 37 collections defined, where `collectiondetailedrecordpagetabs` only specifies values for collections 1, 8, 17, 18, 19.  (The current demo site offers something similar.) Please note that there are no data for collection 2 defined; it is missing from the table.  With your script, you would force Atlantis hard-coded defaults onto collection 2, while with my amendment, you would take existing collection 1 settings, and apply them to all daugher collections that do not have anything defined.  Which I think would be better.  

That's is the difference I was having in mind. An improvement would be to traverse the collection tree in the order of closest parents, and apply that, instead of taking collection 1 settings.  But provided we print some warning for people to check their collection tab settings, we do not really need to do that, I think.

blixhavn: > while with my amendment, you would take existing collection 1 settings, and apply them to all daugher collections that do not have anything defined. 

This is equal to checking "apply for subcollection", which is not necessarily what you want. Imagine the following scenario:
Collection 1 is the parent of collections 2 and 3, and it has a `collectiondetailedrecordpagetabs` entry to show only usage, comments and metadata. Collection 2 is set to show usage, comments, metadata and references, while collection 3 has no entry in `collectiondetailedrecordpagetabs`. The result is that collection 3 displays all tabs, even though its parent is restricted to only three.

However, if one was to apply the upgrade you propose, collection 3 would suddenly also be restricted by its parents settings, changing the behavior (and if it's one thing this upgrade script should do, it's to preserve behavior).

tiborsimko: @blixhavn I see your point.  However, collection 1 usually does not have any records attached to it, so even if people do have settings for it, they do not see its results.  If they had set something up without clicking on "apply for subcollections", then why not to take advantage of this change and propose those instead?  But I agree with you that this is an ultra corner case (kind of equal to forgotten click on "apply for subcollections") and that it may be better to preserve Atlantis way anyway.  Depends on whether we look at this situation more as a bug or as a feature.  I guess I tended to prefer bug-point-of-view, while you opted for feature-point-of-view.    Since "apply for subcollections" is unchecked by default in the UI, you are probably right that it may be better to look at this as at a feature...

tiborsimko: BTW the build fails with:

```
>>> Going to create demo site...
ERROR 1062 (23000) at line 1017: Duplicate entry '33' for key 'PRIMARY'
ERROR: failed execution of /opt/invenio/bin/dbexec < /opt/invenio/lib/sql/invenio/democfgdata.sql
```

Can you have a look?

blixhavn: That's what I did with #7e2e16c, but it seems Travis still fails?

jirikuncar:
We should update fixtures in *pu* after the merge (see #1613).

blixhavn:
Anything happening here?

tiborsimko:
> Anything happening here?

Have you fixed the Travis problem?  Basically you just need to remove offending entry for collection 33 that in invenio/master file is located down below the other entries:

```console
$ git grep -Hni collectiondetailedrecordpagetabs modules/miscutil/demo/
modules/miscutil/demo/democfgdata.sql:71:INSERT INTO collectiondetailedrecordpagetabs VALUES (8, 'usage;comments;metadata');
modules/miscutil/demo/democfgdata.sql:72:INSERT INTO collectiondetailedrecordpagetabs VALUES (19, 'usage;comments;metadata');
modules/miscutil/demo/democfgdata.sql:73:INSERT INTO collectiondetailedrecordpagetabs VALUES (18, 'usage;comments;metadata');
modules/miscutil/demo/democfgdata.sql:74:INSERT INTO collectiondetailedrecordpagetabs VALUES (17, 'usage;comments;metadata');
modules/miscutil/demo/democfgdata.sql:986:INSERT INTO collectiondetailedrecordpagetabs(id_collection,tabs) VALUES(33,'');
```

and then `git rebase master -i` on your feature branch to squash the commits together,  and then repush your branch with the  `-f` option that will update the PR.

tiborsimko:
OK, I'll do the necessary changes during merge now, because we are close to the release time.  
(BTW note also that the upgrade recipe is missing in `tabcreate.sql`; I'll fix this during merge as well.)

tiborsimko:
Thanks, amended as mentioned, amended one regression test case associated to the Holdings tab,  amended commit log message, and merged as 8b0feb06b744a19fb605a59e650c7bada731b15a.

PR #2666, opened 2015-01-13
jsonalchemy: @hidden decorator addition

jirikuncar:
* Adds @hidden decorator for field definition and extends `dumps()`
  method with `filter_hidden=False` argument.  (addresses #2197)


jirikuncar:
@egabancho please have a look.

Issue #2647, opened 2015-01-07
accounts: store first and last name in user model


jmartinm:
Current ``user`` model contains:

```
id = db.Column(db.Integer(15, unsigned=True), primary_key=True,
                   autoincrement=True)
    email = db.Column(db.String(255), nullable=False, server_default='',
                      index=True)
    _password = db.Column(db.LargeBinary, name="password",
                          nullable=False)
    note = db.Column(db.String(255), nullable=True)
    settings = db.Column(db.MutableDict.as_mutable(db.MarshalBinary(
        default_value=get_default_user_preferences, force_type=dict)),
        nullable=True)
    nickname = db.Column(db.String(255), nullable=False, server_default='',
                         index=True)
    last_login = db.Column(db.DateTime, nullable=False,
                           server_default='1900-01-01 00:00:00')
```

We would like to store the user real name in order to address that person in emails and around the interface.

One proposed solution is to extend the user model, like Django https://docs.djangoproject.com/en/1.7/ref/contrib/auth/ to include optional first and last name.

What do you think?

lnielsen:
+1 but call it ``given_names`` and ``family_name`` (see http://www.w3.org/International/questions/qa-personal-names#fielddesign).

And what about ``nickname`` -> ``username``.

Lastly, it would be useful to have a profile (adding extra information like  website, location, address) which would be custom per instance (similar to https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#extending-the-existing-user-model)

lnielsen:
Actually do we need to split in given/family name? Could we just use ``full_name``?

jirikuncar:
Is milestone 2.1 ok?

lnielsen:
I guess we can keep ``nickname`` until 2.0 is released, and then change it afterwards (as discussed in the triage team)

jirikuncar:
#### User model

* `given_names` and `family_name`
* `nickname` -> `username`

tiborsimko:
> given_names and family_name 

Can go to 2.0 already, if INSPIRE needs it urgently for LABS?

>  nickname -> username 

Can go to 2.1 or even later.  (Not sure if it is worth it? Just do `git grep nickname` and `git grep username` to see the current occurrences; ideally one would have to amend both these terms to keep nice distinction.)

jmartinm:
> Can go to 2.0 already, if INSPIRE needs it urgently for LABS?

I will create a PR for this very soon as it is indeed needed.

[...]

PR #2670, opened 2015-01-14
accounts: user full name addition to model

jmartinm:
* Adds corresponding upgrade recipe. (addresses #2647)

### DIFF
--- a/invenio/modules/accounts/upgrades/accounts_2015_01_14_add_name_columns.py
+++ b/invenio/modules/accounts/upgrades/accounts_2015_01_14_add_name_columns.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+##
+## This file is part of Invenio.
+## Copyright (C) 2015 CERN.
+##
+## Invenio is free software; you can redistribute it and/or
+## modify it under the terms of the GNU General Public License as
+## published by the Free Software Foundation; either version 2 of the
+## License, or (at your option) any later version.
+##
+## Invenio is distributed in the hope that it will be useful, but
+## WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Add full name columns to User table."""
+
+from invenio.ext.sqlalchemy import db
+from invenio.modules.upgrader.api import op
+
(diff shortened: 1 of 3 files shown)


invenio/modules/accounts/upgrades/accounts_2015_01_14_add_name_columns.py
```diff
@@ -0,0 +1,51 @@
...
+
+from sqlalchemy import *
+from invenio.ext.sqlalchemy import db
+from invenio.modules.upgrader.api import op
+
+depends_on = [u'accounts_2014_11_07_usergroup_name_column_unique']
+
+
+def info():
+    return "Short description of upgrade displayed to end-user"
```
jirikuncar: can you amend the description so it makes sense for admin?

jmartinm:
Updated.

jirikuncar:
Can you please check http://kwalitee.invenio-software.org/inveniosoftware/invenio/branches/165b52daabd0dd0ef861f62a22cb8d3783bd19f0/jmartinm:pu-add-name-column

jmartinm:
kwalitee updated.

invenio/modules/accounts/upgrades/accounts_2015_01_14_add_name_columns.py
```diff
@@ -0,0 +1,44 @@
...
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Add full name columns to User table."""
+
+from sqlalchemy import *
```
jirikuncar: unused import

tiborsimko:
BTW see my comment https://github.com/inveniosoftware/invenio/issues/2647#issuecomment-69912649 about an option of not introducing new fixed columns, but rather one blob column behaving like a key-value property store.  Could be more future-proof.

jmartinm:
Kwalitee addressed.

Regarding https://github.com/inveniosoftware/invenio/issues/2647#issuecomment-69909510 we decided IRL that keeping them separate instead of in one field should be ok as otherwise we have the complication, for example, after login in with ORCID (which is the use case for the introduction of these columns) on how to form the full name (as we receive it as given-names and family-name)

> BTW see my comment #2647 (comment) about an option of not introducing new fixed columns, but rather one blob column behaving like a key-value property store. Could be more future-proof.

Good point. I personally don't see the need to go that far and prefer the explicit ``given_names`` and ``family_name`` in the model so I would be happy with this commit.

jmartinm:
I have added the fields to the ``legacy_user.py`` so that we can use them in the templates.

jmartinm:
Kwalitee updated in ``legacy_user.py``

PR #2671, opened 2015-01-14
BibClassify: ontology cache check improvement

jalavik:
* Fixes an issue in the ontology cache checking where the wrong
  file path was being checked.


jalavik:
Same problem exists on master: https://github.com/inveniosoftware/invenio/blob/master/modules/bibclassify/lib/bibclassify_ontology_reader.py#L99 

I can provide a PR for master as well.

jalavik:
This got merged as 2ad9e80b1832eb9cc6cf7f4ea0393c373a7416e1

PR #2674, opened 2015-01-14
groups: REST API and API addition

hachreak:
* Adds API and REST API for the groups.

* Implements a policy for "pending" and "member" state of user
  in a group.

* Removes current_user dependency in accounts model.

* Refactors the views to use api instead of access directly to
  the models.

Signed-off-by: Konstantinos Kostis <konstantinos.kostis@cern.ch>
Reviewed-by: Leonardo Rossi <leonardo.r@cern.ch>

### DIFF
--- a/invenio/modules/accounts/models.py
+++ b/invenio/modules/accounts/models.py
@@ -285,6 +285,16 @@ class Usergroup(db.Model):
             UserUsergroup.id_user == id_user
         )
 
+    @classmethod
+    def query_list_usergroups(cls, id_user):
+        """Return query to have a list of groups of the user.
+
+        :param id_user: user's id
+        :return: query to read list
+        """
+        return cls.query.join(UserUsergroup).filter(
+            UserUsergroup.id_user.like(id_user))
+
     def is_part_of(self, id_user):
         """Return True if the user is an admin or member of the group."""
         return db.session.query(
@@ -439,6 +449,16 @@ class UserUsergroup(db.Model):
     def is_part_of(self):
         """Return True if user is member or admin."""
         return self.is_admin() or self.is_member()
+
+    @classmethod
+    def query_list(cls, id_user):
+        """Return query to have a list of UserUsergroups of the user.
+
+        :param id_user: user's id
+        :return: query to read list
+        """
+        return cls.query.filter(
+            cls.id_user.like(id_user))
 
 
 # define query to get admins
--- a/invenio/modules/accounts/models.py
+++ b/invenio/modules/accounts/models.py
@@ -19,16 +19,18 @@
 
 """Account database models."""
 
-# General imports.
 from datetime import datetime
-from flask.ext.login import current_user
+
+from invenio.base.i18n import _
+from invenio.ext.sqlalchemy import db
+from invenio.ext.sqlalchemy.utils import session_manager
+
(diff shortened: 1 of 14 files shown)


invenio/modules/accounts/models.py
```diff
@@ -184,65 +211,95 @@ def login_method_is_external(self):
...
+    @session_manager
+    def approve(self, user):
+        """Approve user from group.
+
+        :param user: User to approve into the group.
+        """
+        # approve user
+        uug = UserUsergroup.query.filter(
+            UserUsergroup.id_user.like(user.id),
+            UserUsergroup.id_usergroup.like(self.id),
```
jirikuncar: maybe you could do it at once with update:
```python
UserUsergroup.query.filter(
    UserUsergroup.id_user.like(user.id),
    UserUsergroup.id_usergroup.like(self.id),
    UserUsergroup.user_status == UserUsergroup.USER_STATUS['PENDING']
).update(dict(user_status=UserUsergroup.USER_STATUS['MEMBER'])
```

invenio/modules/accounts/models.py
```diff
@@ -269,10 +326,66 @@ def get_users_not_in_this_group(self, nickname=None, email=None,
...
+        except NoResultFound:
+            raise IntegrityUsergroupError(
+                'User with id "{0}" not exists').format(id_user_admin)
+        group.join_policy = group.join_policy or \
+            Usergroup.JOIN_POLICIES['VISIBLEOPEN']
+        group.login_method = group.login_method or \
+            Usergroup.LOGIN_METHODS['INTERNAL']
+        group.description = group.description or ''
+        # create the new group
+        db.session.add(group)
```
jirikuncar: IMHO you should join the user to the group before commit

hachreak: done.

invenio/modules/groups/api.py
```diff
@@ -0,0 +1,362 @@
+# -*- coding: utf-8 -*-
```
lnielsen: Couldn't a lot of these methods be class methods on the individual model classes? It makes it much more clear what you are dealing with.

lnielsen:
We can integrate this PR, but I'd like to have more general discussion regarding login method/join policy plus class vs function based apis, hence current PR should not set the API in stone :-) 

cc @bouzlibop

jirikuncar:
> class vs function based apis

I would say we should stick with something similar to `invenio.modules.records.api:Record` - hence I'm in favor of class based APIs.

invenio/modules/groups/api.py
```diff
@@ -0,0 +1,362 @@
...
+    :param id_user: user's id
+    :param group_name: string to filter group list
+    :return filtering query
+    """
+    return models_get_groups_user_not_joined(
+        id_user=id_user,
+        group_name="%%%s%%" % group_name)
+
+
+def query_members_of_group(id_usergroup, curr_uid=None):
```
jirikuncar: `curr_uid` should be `id_user`

hachreak:
@jirikuncar  I refactored the `api.py` code (and adapted `restul.py` and `views.py`)

lnielsen:
This PR is superseded by inveniosoftware/invenio-groups#1

PR #2679, opened 2015-01-15
pidstore: backport

egabancho:


egabancho:
I don't know how much me need this now since we are going for two different databases ... @tiborsimko  WDYT?

jirikuncar:
There is no need of backporting PIDStore since the recommended way is to use Invenio-Migrator instead of running different versions of Invenio on same DB.

PR #2684, opened 2015-01-19
BibUpload: creation_date based on incoming 005

### LINKED ISSUE #1604: BibUpload: adjust creation_date based on incoming 005

kaplun:
* Adjusts creation_date based on incoming 005 tag.
  (closes #1604)

### DIFF
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013 CERN.
+## Copyright (C) 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
@@ -226,7 +226,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
     affected_tags = {}
     original_record = {}
     rec_old = {}
-    now = datetime.now() # will hold record creation/modification date
+    record_modification_date = datetime.now() # will hold record creation/modification date
     record_had_altered_bit = False
     is_opt_mode_delete = False
 
@@ -261,6 +261,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
             write_message("   -Check if the xml marc file is already in the database: DONE" , verbose=2)
 
     record_deleted_p = False
+    record_creation_date = None
     if opt_mode == 'insert' or \
     (opt_mode == 'replace_or_insert') and rec_id is None:
         insert_mode_p = True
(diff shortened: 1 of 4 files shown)


kaplun:
Have to amend regression tests

kaplun:
OK. Done.

modules/bibupload/lib/bibupload.py
```diff
@@ -2274,11 +2276,11 @@ def _process_document_moreinfos(more_infos, docname, version, docformat, mode):
 
 ### Update functions
 
-def update_bibrec_date(now, bibrec_id, insert_mode_p, pretend=False):
+def update_bibrec_date(now, bibrec_id, insert_mode_p, original_005=None, pretend=False):
     """Update the date of the record in bibrec table """
     if insert_mode_p:
         query = """UPDATE bibrec SET creation_date=%s, modification_date=%s WHERE id=%s"""
-        params = (now, now, bibrec_id)
+        params = (original_005, now, bibrec_id)
```
tiborsimko: What if `original_005` is `None` here?  I guess we should force `now`, as was happening before.  (e.g. `original_005 and original_005 or now`)

modules/bibupload/lib/bibupload.py
```diff
@@ -2274,11 +2276,11 @@ def _process_document_moreinfos(more_infos, docname, version, docformat, mode):
 
 ### Update functions
 
-def update_bibrec_date(now, bibrec_id, insert_mode_p, pretend=False):
+def update_bibrec_date(now, bibrec_id, insert_mode_p, original_005=None, pretend=False):
     """Update the date of the record in bibrec table """
```
tiborsimko: If you add `original_005` not as the last optional argument, then you'll have to update all calls to it.  E.g. here:

```console
$ git grep -Hni update_bibrec_date
modules/bibupload/lib/bibupload_regression_tests.py:1545:        bibupload.update_bibrec_date(convert_datestruct_to_datetext(now), rec_id, False)
```

modules/bibupload/lib/bibupload.py
```diff
@@ -288,6 +289,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
                 error = None
         else:
             write_message("   Note: 005 already existing upon inserting of new record. Keeping it.", verbose=2)
+        original_005 = time.strftime("%Y-%m-%d %H:%M:%S", time.strptime(record['005'][0][3].split('.')[0], "%Y%m%d%H%M%S"))
```
tiborsimko: Wouldn't it read better if this statement was indented inside the `else` branch?

modules/bibupload/lib/bibupload.py
kaplun: Somehow this situation should not happen, because in `inser_mode_p` there is always an `original_005` (in the sense that if it is not there the outer call would have initialized it anyway...

modules/bibupload/lib/bibupload.py
kaplun: Then it would be wrong, The meaning of `original_005` is in terms of the flow of code, not WRT where 005 comes from. An `original_005` value has always to be initilaized in `insert_mode_p == True`

tiborsimko: In this case we should rename the variable, because `original_005` does not represent "original 005", as it were.  (See also the above comment that it could be that `original_005=None`, if somebody only reads the `update_bibrec_date()` function signature.)  
Maybe call it `initial_creation_date` or something?

modules/bibupload/lib/bibupload.py
tiborsimko: OK, though explicit conditional statement would leave no doubt on this, hence would be more readable in my eyes.

Perhaps you could enrich docstring saying that when `insert_mode_p` is True,  there should always be `original_005`... even when record does not have any original 005 to boot.  It seems that this variable name is a kind of misnomer, it might be nicer to call variables "original_005 -> record_creation_date" and "now -> record_modification_date", or somesuch, if clients are expected to send both.

kaplun:
Applied your suggestions related to naming things. Let's see if this is good :-)

tiborsimko:

Thanks, amended commit log message and merged as 38bc4c38a3d827024964108aff18c3b5911c89c6.

PR #2688, opened 2015-01-20
deposit: workflow delete fix

lnielsen:
* Fixes issue where deleting a single unpublished deposit object
  could result in other objects being deleted as well.


jirikuncar:
:shipit:

PR #2691, opened 2015-01-21
deposit: eonasdan-bootstrap-datetimepicker fix

### LINKED ISSUE #2689: deposit: bootstrap-datetimepicker v4 not compatible
I just confirmed that v3.1.3 work on Zenodo while v4 does not. I assume the fix is as easy as just locking the bower bundle version to something <v4. 

Alternatively, upgrade the JS to work with v4, but then INSPIRE should be notified that they need to upgrade their bower components (i.e. just add NOTE in commit message).

switowski:
* Locks the bundle version of the library to 3.1.3, because the latest
  one (4.0.0) is not working properly (closes #2689).

### DIFF
--- a/invenio/modules/deposit/bundles.py
+++ b/invenio/modules/deposit/bundles.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2014 CERN.
+## Copyright (C) 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


jirikuncar:
see https://github.com/inveniosoftware/invenio/issues/2689#issuecomment-70984968

jirikuncar:
@switowski can you please rebase? The pu HEAD should be `d1c3791`.

jirikuncar:
Btw we usually write dot before brackets. Example: `... properly.  (addresses #2689)`

switowski:
@jirikuncar Done, although Kwalitee is still complaining about that dot (after moving it as you said).

jirikuncar:
No. It's just the `Acked-by` I have to whitelist it.

jirikuncar:
The real issues are about imports.
```
23:1: I101 Imported names are in the wrong order. Should be invenio, jquery
23:1: I100 Imports statements are in the wrong order. from invenio.base.bundles should be before from invenio.ext.assets
```

PR #2695, opened 2015-01-21
login: fix last_login column update

### LINKED ISSUE #2669: accounts: last_login is never updated
In the user table, the ``last_login`` column always contains ``1900-01-01 00:00:00`` regardless of if the user logged in or not, that value never changes.

``last_login`` seems to be correctly updated in the master branch.

jirikuncar:
* Adds handler for `user_logged_in` signal that updates
  `User.last_login` column.  (closes #2669)

### DIFF
--- a/invenio/ext/login/__init__.py
+++ b/invenio/ext/login/__init__.py
@@ -21,14 +21,18 @@
 
 import urllib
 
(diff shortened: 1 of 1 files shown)


jirikuncar:
@jmartinm can you have a quick look?

PR #2698, opened 2015-01-22
knowledge: slugify and flag to access rest api

### LINKED ISSUE #2686: knowledge: rest api issues
@hachreak I had look at the new knowledge REST API looks great - just have some few minor comments that I'd like to fix:

- [x] ``/api/knowledges`` -> ``/api/knowledge`` (knowledge is uncountable, thus no plural)
- [x] ``/api/knowledges/<int:id>`` -> ``/api/knowledge/<str:id>`` i think it would make more sense for an end-user to access ``/api/knowledges/licenses/`` than ``/api/knowledges/2`` and ``name`` is already unique in the database model so it should be fine.
- [x]  Possibly allow disabling REST API access to a mapping (to ensure that not everything is exposed).
- [ ] 404 response includes HTML (see below)

```http
HTTP/1.0 404 NOT FOUND
Content-Type: text/html; charset=utf-8
Content-Length: 9319
Server: Werkzeug/0.9.6 Python/2.7.9
Date: Tue, 20 Jan 2015 16:15:01 GMT





<!-- base/bundles_base.html: jquery.js, invenio.js, invenio.css -->
<!-- base/bundles_base.html: require.js -->
<!DOCTYPE html>
<html lang="en">
<head>
```

hachreak:
* Adds a new field slug in knowledge and a mechanism
  to automatically generate it.

* Uses the slug as key in REST API instead of id.

* Adds a new field is_api_accessible to allow disabling
  REST API access to a mapping.
  (closes #2686)

### DIFF
--- a/invenio/modules/knowledge/api.py
+++ b/invenio/modules/knowledge/api.py
@@ -25,6 +25,7 @@ import re
 import warnings
 
 from invenio.base.globals import cfg
+from invenio.base.i18n import _
 from invenio.ext.sqlalchemy import db
 from invenio.ext.sqlalchemy.utils import session_manager
 from invenio.modules.search.models import Collection
@@ -48,9 +49,21 @@ except ImportError:
         pass
 
 
+def get_kb_by_slug(slug):
+    """Return the knwKB object with given slug.
+
+    :param slug: slug of knowledge
+    :return: knowledge's object
+    :raises: :exc:`~sqlalchemy.orm.exc.NoResultFound` in case not exist.
+    """
+    return models.KnwKB.query.filter_by(slug=slug).one()
+
+
 def get_kb_by_id(kb_id):
     """Return the knwKB object with given id.
 
+    :param kb_id: id of knowledge
+    :return: knowledge's object
     :raises: :exc:`~sqlalchemy.orm.exc.NoResultFound` in case not exist.
     """
     return models.KnwKB.query.filter_by(id=kb_id).one()
@@ -274,10 +287,9 @@ def kb_exists(kb_name):
     """Return True if a kb with the given name exists.
 
     :param kb_name: the name of the knowledge base
+    :return: True if kb exists
     """
-    return db.session.query(
-        models.KnwKB.query.filter(
-            models.KnwKB.name.like(kb_name)).exists()).scalar()
+    return models.KnwKB.exists(kb_name)
 
 
 def get_kb_name(kb_id):
@@ -300,7 +312,7 @@ def update_kb_attributes(kb_name, new_name, new_description=''):
         .update({"name": new_name, "description": new_description})
 
 
-def add_kb(kb_name="Untitled", kb_type=None, tries=10):
+def add_kb(kb_name=u"Untitled", kb_type=None, tries=10):
     """Add a new kb in database, return the id.
 
     Add a new kb in database, and returns its id
@@ -339,9 +351,9 @@ def add_kb(kb_name="Untitled", kb_type=None, tries=10):
 
     if created is False:
         # TODO raise the right exception
-        raise Exception("Can't create kb \"{0}\".\n" +
-                        "Probabily the server is busy! " +
-                        "Try again later.".format(kb_name))
+        raise Exception(_("Can't create knowledge base \"%(name)s\".\n"
+                          "Probabily the server is busy! "
+                          "Try again later.", name=kb_name))
 
     return kb.id
 
--- a/invenio/modules/knowledge/models.py
+++ b/invenio/modules/knowledge/models.py
@@ -25,8 +25,10 @@ from invenio.base.globals import cfg
 from invenio.ext.sqlalchemy import db
 from invenio.ext.sqlalchemy.utils import session_manager
 from invenio.modules.search.models import Collection
+from invenio.utils.text import slugify
 
 from sqlalchemy.event import listens_for
+
 from sqlalchemy.orm.collections import attribute_mapped_collection
 
 
@@ -43,9 +45,27 @@ class KnwKB(db.Model):
     __tablename__ = 'knwKB'
     id = db.Column(db.MediumInteger(8, unsigned=True), nullable=False,
                    primary_key=True, autoincrement=True)
-    name = db.Column(db.String(255), server_default='', unique=True)
-    _description = db.Column(db.Text, nullable=False, name="description")
+    _name = db.Column(db.String(255), server_default='',
+                      unique=True, name="name")
+    _description = db.Column(db.Text, nullable=False,
+                             name="description", default="")
     _kbtype = db.Column(db.Char(1), nullable=True, default='w', name="kbtype")
+    slug = db.Column(db.String(255), unique=True, nullable=False, default="")
+    # Enable or disable the access from REST API
+    is_api_accessible = db.Column(db.Boolean, default=True, nullable=False)
+
+    @db.hybrid_property
+    def name(self):
+        """Get name."""
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        """Set name and generate the slug."""
+        self._name = value
+        # generate slug
+        if not self.slug:
+            self.slug = KnwKB.generate_slug(value)
 
     @db.hybrid_property
     def description(self):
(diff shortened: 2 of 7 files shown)


invenio/modules/knowledge/models.py
```diff
@@ -166,6 +185,62 @@ def set_dyn_config(self, field, expression, collection=None):
                                     search_expression=expression,
                                     collection=collection)
 
+    @staticmethod
+    def generate_slug(name, tries=10):
```
hachreak: @jirikuncar I know that it's not the best way to generate slug... but it works if e.g. we already have in db two knowledge like: "FAQ" and "FAQ!".
And if someone try to add e.g. "faq" or "$FAQ", in this case we are sure that function returns a valid slug (not considering any racing conditions.. but this is another problem.. XD)

jirikuncar: We can say, that only valid names are `[:alnum:]`.

hachreak: @lnielsen What do you think is better?
1. Use a function like this to generate the slug.
2. Put a restriction on the name (valid names are only `[:alnum:]`)
3. Any better idea? :)

jirikuncar: I would try some more naive/efficient implementation:
```python
i = KnwKB.query.filter(db.or_(
   KnwKB.slug.like(slug),
   KnwKB.slug.like(slug + '-%'),
)).count()
slug = slug + ( '-{0}'.format(i)' if i > 0 else '')
```

hachreak: ok, I like this implementation! :D

hachreak: done.

invenio/modules/knowledge/models.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2011, 2012, 2014 CERN.
+## Copyright (C) 2011, 2012, 2013, 2014, 2015 CERN.
```
lnielsen: Tiny detail…the lack of 2013 just means that the file was not edited in 2013, so it's ok that it is not there.

invenio/modules/knowledge/models.py
```diff
@@ -39,9 +40,26 @@ class KnwKB(db.Model):
...
+    # Enable or disable the access from REST API
+    is_api_accessible = db.Column(db.Boolean, default=True, nullable=False)
+
+    @db.hybrid_property
+    def name(self):
+        """Get name."""
+        return self._name
+
+    @name.setter
+    def name(self, value):
```
lnielsen: It looks to me that the slug is generated every time the name is changed. I think this not desirable, since the slug is pretty close to a primary key that you want to keep stable. I think it's only when the slug is empty that you should try to generate it, and likely I would just try one time and if it fails simply throw an exception.

lnielsen:
Looks good :+1: beside the small comments

invenio/modules/knowledge/models.py
hachreak: code changed to generate only if slug is empty :)

tiborsimko:
@jirikuncar Please add milestone.

invenio/modules/knowledge/api.py
```diff
@@ -321,15 +334,15 @@ def add_kb(kb_name="Untitled", kb_type=None, tries=10):
...
+        except Exception:
             db.session.rollback()
             raise
 
     if created is False:
         # TODO raise the right exception
-        raise Exception("Can't create kb \"{0}\".\n" +
-                        "Probabily the server is busy! " +
-                        "Try again later.".format(kb_name))
+        raise Exception(_("Can't create kb \"{0}\".\n" +
```
jirikuncar: IMHO this should/does not have to be translated. Btw inside bracket you don't need `+` for string concatenation.

invenio/modules/knowledge/models.py
jirikuncar: Cool. Can you add me as co-author? (`Co-authored-by: me`)

invenio/modules/knowledge/api.py
hachreak: And where should it be translated? :)

invenio/modules/knowledge/models.py
hachreak: yep! XD

invenio/modules/knowledge/api.py
jirikuncar: Only if the user can see the error - e.g. flash message.

invenio/modules/knowledge/api.py
```diff
@@ -321,15 +334,15 @@ def add_kb(kb_name="Untitled", kb_type=None, tries=10):
...
+        except Exception:
             db.session.rollback()
             raise
 
     if created is False:
         # TODO raise the right exception
-        raise Exception("Can't create kb \"{0}\".\n" +
-                        "Probabily the server is busy! " +
-                        "Try again later.".format(kb_name))
+        raise Exception(_("Can't create kb \"{0}\".\n"
```
jirikuncar: anyway if you would like to make it translatable you need to use following format:
```python
_("Message %(x_variable_name)s bar.", x_variable_name="John Doe")
```

invenio/utils/text.py
```diff
@@ -737,7 +787,8 @@ def strip_accents(x):
 
 
 def slugify(text, delim=u'-'):
-    """Generates an ASCII-only slug."""
+    """Generate an ASCII-only slug."""
+    text = unicode(text)
```
hachreak: @jirikuncar It's ok this fix? It resolve my warning in upgrade. :)

kaplun: Won't it crash as soon as there is a non ascii character?
```python
In [1]: "bè"
Out[1]: 'b\xc3\xa8'

In [2]: unicode("bè")
---------------------------------------------------------------------------
UnicodeDecodeError                        Traceback (most recent call last)
<ipython-input-2-f6b3400323a3> in <module>()
----> 1 unicode("bè")

UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1: ordinal not in range(128)
```

jirikuncar: @hachreak it would be better to encode the input in upgrade recipe where you are sure that you have UTF8 string.

invenio/modules/knowledge/models.py
```diff
@@ -166,6 +185,46 @@ def set_dyn_config(self, field, expression, collection=None):
...
                                     collection=collection)
 
+    @staticmethod
+    def generate_slug(name):
+        """Generate a slug for the knowledge.
+
+        :param name: text to slugify
+        :return: slugified text
+        """
+        slug = slugify(unicode(name))
```
hachreak: @jirikuncar I move the conversion in unicode here (the same in the upgrade script). :)

jirikuncar: I don't think you need it here.

invenio/modules/knowledge/models.py
```diff
@@ -166,6 +185,46 @@ def set_dyn_config(self, field, expression, collection=None):
...
+        return slug + ('-{0}'.format(i) if i > 0 else '')
+
+    @staticmethod
+    def exists(kb_name):
+        """Return True if a kb with the given name exists.
+
+        :param kb_name: the name of the knowledge base
+        :return: True if kb exists
+        """
+        return db.session.query(
```
jirikuncar: why aren't you re-using the `query_exists`?

invenio/modules/knowledge/restful.py
```diff
@@ -386,38 +438,86 @@ def get(self, id):
...
+        '/api/knowledge/<string:slug>/mappings/to'
     )
     api.add_resource(
         KnwKBMappingsFromResource,
-        '/api/knowledge/<int:id>/mappings/from'
+        '/api/knowledge/<string:slug>/mappings/from'
+    )
+
+    # for other urls, return "Method Not Allowed"
+    api.add_resource(
```
hachreak: @jirikuncar @lnielsen 
I used `NotImplementedFooResource` and `NotImplementedFooBarResource` to cover any possible url.
Do you know if exists the possibility to set automatically: if user request any other paths, then returns "Method Not Allowed" (instead of html page)?

invenio/modules/knowledge/restful.py
```diff
@@ -386,38 +438,86 @@ def get(self, id):
...
+        NotImplementedFooResource,
+        '/api/knowledge/<string:slug>/<string:foo>'
```
jirikuncar: @hachreak could you try something like `/api/knowledge/<string:slug>/<path:foo>`?

invenio/modules/knowledge/restful.py
hachreak: done! 
Thanks @jirikuncar :D

PR #2700, opened 2015-01-23
WebStyle: ping handler to return 200

kaplun:
* When queried by a load-balancer (such as haproxy) have the /ping
  web handler to return HTTP status code 200, even e.g. if
  ACCESS_CONTROL_LEVEL_SITE is set to 1 (which implies returning
  HTTP error status code 503 for any other web handler so that
  crawlers don't register these requests).


tiborsimko:

Thanks, amended commit log message and merged as 28cf7d105cda1a7f081e165971ed379d0ae729be.

PR #2703, opened 2015-01-26
deposit: DynamicItemWidget remove button fix

jmartinm:
* Do not show the remove button on the first item of a
  DynamicItemWidget.

### DIFF
--- a/invenio/modules/deposit/field_widgets.py
+++ b/invenio/modules/deposit/field_widgets.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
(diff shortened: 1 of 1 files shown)


jmartinm:
Can you have a look @lnielsen ?

lnielsen:
In the airport, so didn't have much time to look, but I would think it makes sense to have a remove button even for the first line?

jmartinm:
Well it looks like a bug if there is only one element:



But I see your point, if there are several elements in the list, then it can make sense to remove the first element...

jmartinm:
Closing this as for the moment I am not sure what the best approach is.

PR #2712, opened 2015-01-28
oauthclient: account info session fix

jmartinm:
* Sets the account info session upon retrieval to make it
  available for remote setup handlers.

### DIFF
--- a/invenio/modules/oauthclient/contrib/orcid.py
+++ b/invenio/modules/oauthclient/contrib/orcid.py
@@ -127,17 +127,20 @@ def account_info(remote, resp):
 @session_manager
 def account_setup(remote, token):
     """Perform additional setup after user have been logged in."""
+    import json
+    import requests
+
+    from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+
     from invenio.modules.accounts.models import User, UserEXT
     from invenio.ext.sqlalchemy import db
     from ..handlers import token_session_key
 
-    from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
-
-    import json
-    import requests
-
     orcid = session.get(token_session_key(remote.name) +
-                        "_account_info").get("external_id")
+                        "_account_info", {}).get("external_id")
+
+    if not orcid:
+        return
 
     extra_data = {
         "orcid": orcid


jmartinm:
This avoids an error I got in the 'rare' case when ``remoteACCOUNT`` and ``remoteTOKEN`` are truncated and then the [orcid setup handler](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/contrib/orcid.py#L128) did not have access to the ``account_info`` session throwing the error:

```python
AttributeError: 'NoneType' object has no attribute 'get' (orcid.py:140:account_setup)
```

BTW, @lnielsen I was wondering why the account setup is not called when there is extra_data?
https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L231

jirikuncar:
@jmartinm I've fixed some PEP8/257 issues in this file in #2714.

lnielsen:
This problem I've also seen, and it happens when the user is already logged in (e.g. go to Profile > Linked accounts and link your account).

Actually account_info should only be used to identify the user if you don't already know who it is. Account setup is a one time setup when the link between the the remote and local account is created. Currently this is determined by lack of extra_data which is properly not the right way.

lnielsen:
I think it's actually ORCID setup handler that has to be adapted in this case?

jmartinm:
Yes, for my error it would be enough with adding a sanity check in the ORCID setup https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/contrib/orcid.py#L139

to not assume that account_info is in the session.

I will create a new PR with such change.

jmartinm:
I have updated the PR moving the sanity check to the ORCID setup handler. Please, have a look.

invenio/modules/oauthclient/contrib/orcid.py
```diff
@@ -127,17 +127,20 @@ def account_info(remote, resp):
...
-    import requests
+    from invenio.modules.accounts.models import User, UserEXT
+    from invenio.ext.sqlalchemy import db
+    from ..handlers import token_session_key
 
     orcid = session.get(token_session_key(remote.name) +
-                        "_account_info").get("external_id")
+                        "_account_info", {}).get("external_id")
+
+    if not orcid:
```
jirikuncar: When this happens is it considered an error? Shall we add a logging?

Alternative:
```python
try:
     orcid = session.get(token_session_key(remote.name) +
                        "_account_info").get("external_id")
except KeyError:
    current_app.logger.exception("No ORCID available in user session.")
    return
```
WDYT @jmartinm @lnielsen?

jmartinm: With the current code, if I understood correctly, this situation will happen when the user has already linked his local account to an external one and then [this condition](https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/oauthclient/handlers.py#L246) is True

```python
if not token.remote_account.extra_data and \
       remote.name in signup_handlers:
        handlers['setup'](token)
```

which is the case when no ``extra_data`` is available in that table.

I am not sure if it should be considered as an error to be logged or something to ignore, @lnielsen maybe has some input on this?

jirikuncar: ping @lnielsen

jirikuncar: @jmartinm @lnielsen do you have any updates?

lnielsen:
This commit can be integrated as it fixes the issue, but I've discovered quite some problems with the external account linking. E.g. if the user is already authenticated the link is not created. I will submit a PR that refactors this, and it will touch the above link as well.

lnielsen:
@jmartinm If ok with you, I'll close this PR as it's superseded by #2826.

jmartinm:
> @jmartinm If ok with you, I'll close this PR as it's superseded by #2826.

@lnielsen sure go ahead.

jirikuncar:
Superseded by #2826.

PR #2713, opened 2015-01-28
mailutils: email header type detection

dset0x:
* Fixes an issue where unicode strings were split by commas in the
  email.

Reported-by: Thorsten Schwander <thorsten.schwander@gmail.com>

### DIFF
--- a/modules/miscutil/lib/mailutils.py
+++ b/modules/miscutil/lib/mailutils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014 CERN.
+## Copyright (C) 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


modules/miscutil/lib/mailutils.py
```diff
@@ -24,24 +24,24 @@
 
 __revision__ = "$Id$"
 
+import os
+import re
 import sys
+from cStringIO import StringIO
 from time import sleep
+
+import six
```
tiborsimko: If `six` is required, you may need to amend the `INSTALL` file and the `requirements.txt` file about this.

modules/miscutil/lib/mailutils.py
```diff
@@ -24,24 +24,24 @@
 
 __revision__ = "$Id$"
 
+import os
+import re
 import sys
+from cStringIO import StringIO
```
jirikuncar: if you are using `six` for Python 3 compatibility then use check also [`six.StringIO`](https://pythonhosted.org/six/#six.StringIO).

jirikuncar: Ok, I see that you have removed the `six` import now.

modules/miscutil/lib/mailutils.py
dset0x: Woops. `six` on master is probably unnecessary to begin with. Replaced with `basestring` (as `six` does: https://bitbucket.org/gutworth/six/src/78315705481c3f5e491a3816e41aa66c54c8f965/six.py?at=default#cl-49 )

tiborsimko:

Thanks, amended commit log message and merged as 34f63332d6cba3dc4b8828fbcf44ddfcee7409d1.

PR #2719, opened 2015-01-28
relationships: initial implementation

MSusik:
* Introduces Relationship SQLAlchemy model.

* Introduces Node class for the representation of a node in
  relationship graph.

* Includes tests.

Few notes about the module:
+ I'm not sure if I should implement more features (if yes, which ones?).
+ How should I add you (@jirikuncar) as another author of the code? Is second `Signed-off by` enough?

To do:
- [x] - Make everybody discuss this PR.
- [x] - Switch to use `Record`/`JSONAlchemy`/`whatever` instead of using `RecordMetadata` cache.
- [x] - Make tests working.
- [x] - Polish the code.
- [x] - Document `Node` class.
- [x] -Switch from using `JSON` to `JSONB`. -> `sqlalchemy_utils` will probably switch from using `JSON` to `JSONB`.

### DIFF
--- a/invenio/modules/relationships/api.py
+++ b/invenio/modules/relationships/api.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+##
+## This file is part of Invenio.
+## Copyright (C) 2015 CERN.
+##
+## Invenio is free software; you can redistribute it and/or
+## modify it under the terms of the GNU General Public License as
+## published by the Free Software Foundation; either version 2 of the
+## License, or (at your option) any later version.
+##
+## Invenio is distributed in the hope that it will be useful, but
+## WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Relationship API."""
+
+import importlib
+
+from .config import RELATIONSHIP_ENGINE, RELATIONSHIP_NODE
+
+try:
+    __engine = __import__('flask').current_app.config['RELATIONSHIP_ENGINE']
+except RuntimeError:
+    __engine = RELATIONSHIP_ENGINE
+
+try:
+    __node = __import__('flask').current_app.config['RELATIONSHIP_NODE']
+except RuntimeError:
+    __node = RELATIONSHIP_NODE
+
+Node = getattr(importlib.import_module(__node[0]), __node[1])
+Relationship = getattr(importlib.import_module(__engine[0]), __engine[1])
--- a/invenio/modules/relationships/engines/sqlalchemy.py
+++ b/invenio/modules/relationships/engines/sqlalchemy.py
@@ -1,0 +1,145 @@
+# -*- coding: utf-8 -*-
+##
+## This file is part of Invenio.
+## Copyright (C) 2015 CERN.
+##
+## Invenio is free software; you can redistribute it and/or
+## modify it under the terms of the GNU General Public License as
+## published by the Free Software Foundation; either version 2 of the
+## License, or (at your option) any later version.
+##
+## Invenio is distributed in the hope that it will be useful, but
+## WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+## General Public License for more details.
+##
+## You should have received a copy of the GNU General Public License
+## along with Invenio; if not, write to the Free Software Foundation, Inc.,
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Implementation of ``Node`` using ``SQLAlchemy``."""
+
+from invenio.ext.sqlalchemy import db
+
+from ..models import Relationship as SQLAlchemyRelationship
+from ..node import Node
+
+
+class SQLAlchemyNode(Node):
+
+    """Node implementation for ``SQLAlchemy``."""
+
+    def __init__(self, id=None, storagename=None):
+        """Initialize node.
+
+        :param json: SmartJSON
+            The entity. If not provided, both ``id`` and ``namespace`` need
+            to be provided.
+        :param id: integer
+            The id of the entity.
+        :param namespace: string
+            The namespace the entity belongs to.
+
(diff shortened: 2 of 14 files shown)


invenio/modules/relationships/models.py
```diff
@@ -0,0 +1,163 @@
...
+    link_type = db.Column(db.String(255), index=True)
+    model_to = db.Column(db.String(255), index=True)
+    id_to = db.Column(db.Integer(unsigned=True), index=True)
+
+    confidence = db.Column(db.Float, nullable=True)
+
+    def __init__(self, model_from, id_from, link_type, model_to, id_to):
+        """Initialize the relationship between two records.
+
+        Parameters
```
jirikuncar: please remove these sections

jirikuncar:
@MSusik what about keeping the original commit and adding your changes further? Otherwise `Co-authored-by: me` would be nice.

invenio/modules/relationships/errors.py
```diff
@@ -0,0 +1,34 @@
...
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Errors used by relationship module."""
+
+
+class NoSuchRecordException(Exception):
+
+    """Raised when the record was expected, but the given id is invalid."""
+
+    pass
```
jirikuncar: `pass` is not needed

invenio/modules/relationships/models.py
```diff
@@ -0,0 +1,163 @@
...
+            raise WrongModelNameException("Wrong model for record %s" %
+                                          self.id_to)
+        return model
+
+    @utils.session_manager
+    def save(self):
+        db.session.add(self)
+
+
+class Node(object):
```
jirikuncar: Please check the original implementation. The idea behind was to allow linking between different models.

invenio/modules/relationships/models.py
```diff
@@ -0,0 +1,163 @@
...
+
+from .errors import NoSuchRecordException, WrongModelNameException
+
+
+class Relationship(db.Model):
+
+    """Represent a graph edge."""
+
+    uuid = db.Column(UUIDType(binary=False), primary_key=True)
+    model_from = db.Column(db.String(255), index=True)
```
kaplun: What is `model_from`/`model_to`? Aren't we going to link records all the time? If so, isn't the model already a property of the record?

invenio/modules/relationships/models.py
kaplun: So that I can learn: why?

kaplun:
> I'm not sure if I should implement more features (if yes, which ones?).

* We need syntactic sugar in JSONAlchemy to express such the concept of relationship, but it's not yet clear to me how to best express it. @jirikuncar?

* Basically it should be very easy to create in calculated fields in JSONAlchemy such as: record IDs of all the records to that refers to the current one. IDs of all the children record (in case of parent/children relationship), IDs of superseded records, etc.

invenio/modules/relationships/models.py
MSusik: This allows linking between different models as well. I didn't get the idea with your registry inside `Node` metaclass, though.

invenio/modules/relationships/models.py
MSusik: I'll explain it in your RFC.

Anyway, it should be easy to query for relationship for a given model. This is the simpliest solution.

MSusik:
How should I replace using `RecordMetadata`?

For example given a test case:
+ Create empty record.
+ Get JSON representation of it (with metadata).
+ Create relationship
+ etc.

It seems that only place where the `JSON` is stored is `RecordMetadata`. So, if I want to get JSON, I believe I should use legacy way to obtain record, create xml, then pass it to `Record.create(master_format="xml" ...)`, and only then use it. Am I correct?

MSusik:
@kaplun:
> We need syntactic sugar

>Basically it should be very easy to create in calculated fields in JSONAlchemy such as

With the methods provided by this PR, it should be very easy to implement both features inside JSONAlchemy.

MSusik:
@jirikuncar Ready to review.

MSusik:
Ready. Not sure why builds don't run correctly.

invenio/modules/relationships/api.py
```diff
@@ -0,0 +1,37 @@
...
+## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Relationship API."""
+
+import importlib
+
+from .config import RELATIONSHIP_ENGINE, RELATIONSHIP_NODE
+
+try:
+    __engine = __import__('flask').current_app.config['RELATIONSHIP_ENGINE']
```
jirikuncar: please use `LocalProxy` instead and register the engine and node on application context

invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,145 @@
...
+        :param outwards: boolean
+            Get the vertices that are destinations from this node.
+        :param link_type: string
+            Type of the edges. Optional.
+
+        :returns: set
+            Nodes that are neighbours of this node.
+        """
+        edges = self.edges(inwards, outwards, link_type, loops_doubled=False)
+        set1 = {SQLAlchemyNode(id=node.id_from,
```
jirikuncar: is it Python 2.6 friendly? we should check that nobody is planning to use *Invenio 2.0* on machines with Python < 2.7

invenio/modules/relationships/node.py
```diff
@@ -0,0 +1,170 @@
...
+
+    __storage_engine_registry__ = []
+    __entity_types_registry__ = {}
+
+    def __init__(cls, name, bases, dct):
+        """Register cls to type registry."""
+        if hasattr(cls, '__storagename__'):
+            cls.__storage_engine_registry__.append(cls)
+            if cls.__storagename__ not in cls.__entity_types_registry__:
+                print "UPDATER"
```
jirikuncar: forgotten print statements

invenio/modules/relationships/api.py
MSusik: There is a small problem with a `LocalProxy`. As `SmartJSON` inherits from `Node`, when the registries are run, before the application context is ready, the name is imported from this file. If I use `lazy_import` to do it, it doesn't change anything, as it still has to be imported when `SmartJSON` is defined.

To be honest, I can't see a nice solution to this problem. The one in this file is definitely wrong, although when sb will implement another engine, he will be forced to improve it.

As I mentioned before: I don't know what are the plans for integrating new `JSONAlchemy` package inside `Invenio`. As the new module is dependant on it, probably the implementation should be changed.

jirikuncar:
Superseded by #2823.

PR #2730, opened 2015-01-30
sqlalchemy: handling of composites in `todict`

dset0x:
This addition emerged from #2568.

In short, it is some convenience code for people who want to declare composites in SQLAlchemy models and have them easily converted into nested dictionaries.


invenio/ext/sqlalchemy/utils.py
```diff
@@ -90,20 +191,26 @@ def getRelationships(self):
...
         return retval
 
-    def todict(self):
-        """Convert model to dictionary."""
+    def todict(self, composites=False, without_none=True):
+        """Convert model to dictionary.
+
+        :param composites: also return composites
+        :type composites: bool
+        """
```
dset0x: Explain more here.

invenio/ext/sqlalchemy/utils.py
```diff
@@ -46,6 +46,107 @@ def save(self):
...
+
+    def __composite_orig_keys__(self):
+        """Iterate over the key names as defined in the database model."""
+        for orig_key in self.__composite_mapper__().values():
+            yield orig_key
+
+    def __composite_mapper__(self):
+        """Define the relationship between the model and local names.
+        For example if this composite is built from the following columns:
+            config_a = Column(..)
```
jirikuncar: `.. code-block:: python` could help with readability ...

invenio/ext/sqlalchemy/utils.py
```diff
@@ -46,6 +46,107 @@ def save(self):
 all_cap_re = re.compile('([a-z0-9])([A-Z])')
 
 
+class CompositeItems(MutableComposite):
```
jirikuncar: `NameError: name 'MutableComposite' is not defined`

jirikuncar:
@dset0x what is the status? Do you have any updates? Please include unit tests for this non-trivial functionality. Thank you!

dset0x:
@jirikuncar I'm afraid not. I have switched my focus towards SCOAP3 checker plugins during the past couple of weeks.

I agree, unit tests are a must here. I will try to get back to this as soon as possible.

dset0x:
Started some work on the tests.

I noticed that the `fromdict` function is a bit deceivingly named, as it only updates from a dict, not creates. Was this really its purpose?

dset0x:
One may notice the [support](http://docs.sqlalchemy.org/en/latest/orm/extensions/mutable.html#id1) for `__getstate__` and `__setstate__` in composites, but as you will see in the comments in the code, this is impractical for our case, as composites that have all `None` members, are not instantiated at all. We may still be able to partially leverage this down the line, but that should not affect the interface as far as modules are concerned.

invenio/ext/sqlalchemy/testsuite/test_composite_dicts.py
```diff
@@ -0,0 +1,206 @@
...
+from invenio.ext.sqlalchemy.utils import get_model_type
+from invenio.testsuite import make_test_suite
+
+
+Base = declarative_base()
+Base = get_model_type(Base)  # Patch with our methods
+
+
+class MyComposite(MutableComposite):
+
```
PXke: Could you please add a class comment ? :pray:

dset0x: @PXke Done, thanks for reminding me about this.

dset0x:
I will revise for missing tests and will be dropping the WIP in the following days.
Please excuse the SQLAlchemy introspection insanity and PEP8 violations for the sake of test readability.

invenio/ext/sqlalchemy/testsuite/test_composite_dicts.py
PXke: Thanks a lot ! It can seem picky but, I didn't realize when I was reading the code I was in the test area. (I admit I skipped to check the import order, (BTW I think that from unittest import TestCase should be on top)), so for my next time I will directly see that it is a test classes ;) ! Thanks again

dset0x: Ah yes, outside of a test file, naming something `MyComposite` would definitely a prayer. 

Regarding the import order, it is now refactored with `rope`, so I don't think that's something I would manually change.

dset0x:
Commits squashed. `master` here we come.

jirikuncar:
> * NEW Handles composite columns in `todict` and `fromdict` model
  patched-in methods can now handle composite columns gracefully by
  representing them as dictionaries.

Can you split it into two sentences? Thanks

jirikuncar:
@dset0x nice trick to avoid kwalitee. Please rebase and re-push. image

invenio/ext/sqlalchemy/testsuite/test_composite_dicts.py
```diff
@@ -0,0 +1,224 @@
+# -*- coding: utf-8 -*-
```
jirikuncar: @dset0x `invenio/ext/sqlalchemy/testsuite/` is not part of test suite lookup registry. Please move the file to: `invenio/testsuite/test_ext_sqlalchemy_utils.py`

dset0x:
@jirikuncar That was not meant to be a trick. Why would I try to avoid kwalitee? I realized it was not passing the checks after removing WIP after bad squashing. Hence the WIP remove-add-remove. Hence the edit in the bump message.

If kwalitee only runs when there is no WIP, how can one safely remove WIP?

jirikuncar:
> how can one safely remove WIP?

One can run `kwalitee check message/files` locally.

jirikuncar:
As PostgreSQL support is coming we might have a look into http://sqlalchemy-utils.readthedocs.org/en/latest/data_types.html#module-sqlalchemy_utils.types.pg_composite instead. If urgent need for this general feature arises we should consider contributing to SQLAlchemy(-Utils) directly.

PR #2281, opened 2014-09-15
WIP search: new facets gui engine

kneczaj:
This pull request is to make a configurable gui for facets with intension to make it work later with Elastic Search.

TODO:
- [x] write tests with Jasmine
- [x] there are some problems with parentheses at query
- [x] with 'links' theme, there are state propagation problems
- [x] rebase with Lars's code
- [x] fix problem with jasmine-jquery
- [x] fix problem with jasmine-ajax
- [x] check with js debug mode off (uglification, bundling etc.)
- [x] "Any" should be translatable ... https://github.com/inveniosoftware/invenio/pull/2281#discussion_r17543212

ElasticSearch comment:
The adaptation should be quite easy, because every facet filter returns a list of checked options for every state, although I think there is no sense to do t without proper elastic search environment.

Alters: https://github.com/jirikuncar/invenio/pull/184, https://github.com/inspirehep/invenio/pull/79

Depends on:
https://github.com/inveniosoftware/invenio/pull/2537

PR #2752, opened 2015-02-06
search: final facets and jasmine tests

jacquerie:
Here's the rebased version of #2281 on top of the current pu branch. Jasmine-Ajax changed major version since, therefore I updated the tests (and the required version).

Tests are green and functionality seems to be there.

(CC: @lnielsen @jalavik @jmartinm @kaplun)


jacquerie:
About the [blind except](https://github.com/jacquerie/invenio/blob/e358f9f6a3f3026b4e482196459a06f593dc0f4d/invenio/modules/search/facet_builders.py#L133): it's hard to say what it is guarding against. Perhaps an `ImportError` from [line 50](https://github.com/jacquerie/invenio/blob/e358f9f6a3f3026b4e482196459a06f593dc0f4d/invenio/modules/search/facet_builders.py#L50)?

jirikuncar:
> About the blind except: ...

Can you change it at least to `except Exception:`? Thanks

jacquerie:
> Can you change it at least to `except Exception:?` Thanks

Yup! Done. I also spotted a typo in the commit message.

jirikuncar:
@jacquerie please keep `##` in the licence. I will update `kwalitee` it ignore this PEP8 issue.

jacquerie:
@jirikuncar Done. I guess I shouldn't always do what `kwalitee` says : )

BTW, there's still the missing copyright on deleted file issue.

jacquerie:
Ugh, sorry for the botched rebase. I misinterpreted #2852 as saying "rebase everything against master", but since this PR is against `invenio/pu` this results in this mess.

I'm wondering what should I do now, though. Since `invenio/pu` is an old branch, this PR doesn't make sense anymore, as far as I understand GitHub...

jirikuncar:
@jacquerie I've updated *pu* with latest *master* so it should be better after another rebase.

invenio/modules/search/facet_builders.py
```diff
@@ -19,18 +19,20 @@
...
-    get_search_results_cache_key_from_qid
-from .models import Collection
+from intbitset import intbitset
 
 from invenio.base.globals import cfg
-from intbitset import intbitset
+
+from six import iteritems
+
+from .cache import get_search_results_cache_key_from_qid, search_results_cache
```
jirikuncar: can you break the line?

invenio/modules/search/facet_builders.py
```diff
@@ -139,9 +141,15 @@ def get_facets_for_query(self, qid, limit=20, parent=None):
         """Return facet data."""
         from invenio.legacy.search_engine import get_most_popular_field_values,\
             get_field_tags
-        return get_most_popular_field_values(self.get_recids(qid),
-                                             get_field_tags(self.name)
-                                             )[0:limit]
+        most_popular_fields = get_most_popular_field_values(
+            self.get_recids(qid), get_field_tags(self.name))[0:limit]
+        most_popular_fields = map(lambda x: {
```
jirikuncar: - [ ] please use Python list comprehension (more effective)
- [ ] it would be nice to make the structure as similar to Elastic Search as possible where it makes sense (see http://www.elastic.co/guide/en/elasticsearch/reference/1.x/search-aggregations-bucket-nested-aggregation.html)

jirikuncar:
@nharraud can you have a look and take inspiration for the new ElasticSearch facet implementation?

PR #2768, opened 2015-02-12
Master fulltext by default

egabancho:


egabancho:
See https://github.com/CERNDocumentServer/cds-ops/pull/30#issuecomment-44002640

tiborsimko:
@egabancho What do we do with this PR in the light of @ludmilamarian's comment in https://github.com/CERNDocumentServer/invenio/pull/30#issuecomment-44002640 ?  

Note also that Jenkins does not test Solr; for this one has to launch `make install-solrutils` and set `CFG_SOLR_URL` manually...

egabancho:
Invenio 1.x is in feature freeze state.

Issue #2075, opened 2014-08-12
BibField: clean up `doi` field and other FIXME in `atlantis.cfg`


tiborsimko:
Similarly to #1557, there are more clean-ups to be done in `atlantis.cfg`.  For example, in `invenio.conf` we have:

```
CFG_OAI_ID_FIELD = 909COo
CFG_OAI_SET_FIELD = 909COp
CFG_OAI_PREVIOUS_SET_FIELD = 909COq
```

while in `atlantis.cfg` we have:

```
@persistent_identifier(4)
oai:
    creator:
        @legacy((("024", "0248_", "0248_%"), ""),
                ("0248_a", "oai"),
                ("0248_p", "indicator"))
        marc, "0248_", {'value': value['a'], 'indicator': value['p']}
    producer: 
        json_for_marc(), {"0248_a": "oai", "0248_p": "indicator"}
```

and:

```
FIXME_OAI:
    creator:
        @legacy((('909', '909CO', '909CO%'), ''),
                ('909COi', 'group'),
                ('909COs', 'number'),
                ('909COo', 'id'),
                ('909COp', 'set'),
                ('909COq', 'previous_set'))
        marc, "909CO", {'group':value['i'], 'number':value['s'], 'id':value['o'], 'set':value['p'], 'previous_set':value['q']}
```

Notice the naming differences: e.g. "oai.indicator" should rather be called "oai.set"; "oai.value" should rather be called "oai.id".

Historically, `909CO` MARC tag in Invenio Atlantis defaults was used to store OAI information.  Later, `0248` was chosen on CDS and elswhere as a better MARC tag location.  Invenio Atlantis defaults were not changed though, which permitted e.g. to test easily how Invenio is configurable for various site conditions.

However, as it is agreed that `0248` is better OAI location, we should probably move defaults there, so that new Invenio installations are using by default "good" MARC tags already.

(This is similar with 773 vs 909C4, etc, see #1557.)

P.S. See also other FIXMEs in `atlantis.cfg`.  It would be useful to go through the file, fix all FIXMEs, and fix `demobibdata.xml` and `democfgdata.sql` and `invenio.conf` accordingly.

tiborsimko:
@Kennethhole and @egabancho have been working on this in recent months, so I wonder whether we can co-sprint on this issue to fix it before forthcoming v2.0.0 is out. The motivation is to start with relatively "clean" JSON model, so that we don't force sites to change their data model once this issue will be fully completed. 

[...]

PR #2774, opened 2015-02-13
records: multiple `system_control_number`s

dset0x:
* Forces `system_control_number` to always return a list, since it is
  possible that multiple exist and the output be consistent regardless.

### DIFF
--- a/invenio/modules/records/testsuite/fields/atlantis.cfg
+++ b/invenio/modules/records/testsuite/fields/atlantis.cfg
@@ -881,6 +881,8 @@ subject_indicator:
 
 @persistent_identifier(2)
 system_control_number:
+    schema:
+        {'system_control_number': {'type': 'list', 'force': True}}
     creator:
         @legacy((("035", "035__", "035__%"), ""),
                 ("035__a", "system_control_number"),


tiborsimko:
Note that we are generalising MARC setup to use proper standard names everywhere, and to use schemas everywhere, see #2075, so that issues like this will be soon solved that way.

jirikuncar:
@dset0x please amend component name in the commit message: `jsonalchemy` -> `records`

jirikuncar:
You might also want to check `jsonalchemy` package (`jsonalchemy/contrib/marc`).

dset0x:
>You might also want to check jsonalchemy package (jsonalchemy/contrib/marc).

Will do. I believe there is more fields that should use the same fixup, so I will try to spot them and bundle that into one PR.

jirikuncar:
> I believe there is more fields that should use the same fixup

Please note that most of this work has been done by generating JSONAlchemy configuration from MARC standard in new `jsonalchemy` package.

dset0x:
I put some script together to get some insight on which keys are likely to return multiple values in the sample invenio database.

```python
from invenio.modules.records.api import get_record
from collections import defaultdict

def doit():
    found_keys = defaultdict(int)

    for i in range(1,500):
        r = get_record(i);
        if r:
            for key, val in r.items():
                if issubclass(type(val), list):
                    item_count = len(key)
                    if item_count > found_keys[key]:
                        found_keys[key] = item_count

    for count, key in sorted(((val, key) for key, val in found_keys.items()), reverse=True):
        print count, key
```

Everything that has a `no` in `explicit_list` is has potentially unstable output (ie: it may sometimes be a list and sometimes not).
`occurences` is "how many times this returned a list"

```
explicit_list   occurrences  key

yes             27           _additional_corporate_names
no              25           _additionla_meeting_names
no              22           accelerator_experiment
no              21           system_control_number
yes             20           classification_terms
no              19           other_report_number
yes             19           _additional_authors
no              16           publication_info
no              15           version_history
no              15           thesaurus_terms
??              15           files_to_upload
yes             15           corporate_names
no              15           cataloguer_info
no              14           internal_notes
no              14           author_archive
no              13           report_number
yes             13           meeting_names
no              12           journal_info
yes             11           collections
no              11           affiliation
no              9            reference
??              9            documents
yes             8            keywords
??              7            comment
yes             7            authors
yes             5            files
no              3            udc
no              3            fft
no              3            doi
```

#2075 is not very clear to me yet, so I don't know if this work is useful for it, or if I should invest time in applying the changes that the above output suggests. @jirikuncar? @jalavik?

jalavik:
@dset0x Thanks for this list! I believe it shows that we should have a stable interface for this. Most likely something like SQLAlchemy's query interface allowing to specify for `.one()` or many - out good ol' pal BibRecord also kinda had this idea via `record_get_field_value()` and `record_get_field_values()`. Note the trailing 's' in the last one to denote a list (we might wanna choose different naming to differentiate better).

jalavik:
@dset0x As for the immediate future, we can probably just solve this in overlay configuration until #2075. Future enhancements to `jsonalchemy` as a whole is coming so this issue could be addressed then.

PXke:
So what is the status ?

jirikuncar:
Can be added to *Invenio Demosite*.

PR #2785, opened 2015-02-17
global: `watchdog` package addition

### LINKED ISSUE #2778: global: flask app not reloading
As described here: http://stackoverflow.com/a/28242285/2230231, due to some changes in `Werkzeug 0.10`, now automatic changes detection is not working (unless it was my case in my `zenodo` local env and also @hachreak told me that he has the same issue) - it also affects noticeable increase of CPU usage.

The **proposed solution** (taking [this](https://github.com/mitsuhiko/werkzeug/blob/4e2c6746e565eaba7b3bd7489337580b62db610b/werkzeug/_reloader.py#L246-L251) into account) is to install [watchdog](https://pypi.python.org/pypi/watchdog).

Shall we add this somewhere in requirements?

bouzlibop:
* Enables automatic code changes detection. (closes #2778)

* Reduces CPU usage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -132,6 +132,7 @@ extras_require = {
     ],
     "development": [
         "Flask-DebugToolbar==0.9.0",
+        "watchdog==0.8.3",
     ],
     "elasticsearch": [
         "pyelasticsearch>=0.6.1"


setup.py
```diff
@@ -132,6 +132,7 @@ def run(self):
     ],
     "development": [
         "Flask-DebugToolbar==0.9.0",
+        "watchdog==0.8.3"
```
jirikuncar: Please add comma (`,`) at the end of the line so next time somebody adds a package we get again nice one line diff.

jirikuncar:
@bouzlibop please fix a typo in commit message `useage` -> `usage`

bouzlibop:
Thanks @jirikuncar. I just fixed these things you mentioned.

kaplun:
Out of curiosity: how the watchdog package is used in Invenio? It is not referred anywhere.

bouzlibop:
@kaplun, `flask` is based on `Werkzeug`, which is dealing (besides huge amount of other stuff) with automatic reloading on code changes and it's using `watchdog` if [available](https://github.com/mitsuhiko/werkzeug/blob/4e2c6746e565eaba7b3bd7489337580b62db610b/werkzeug/_reloader.py#L246-L251).

PR #2787, opened 2015-02-17
WIP BibIndex: improved unicode support

kaplun:
- Introduces transparent support for uft8_unicode_520_ci collation,
  by discovering on the fly if a given index has been created with
  this particular collation and by using it upon queries.
- Lowers and Strips accents also for phrase indexes (and not only
  word and pair indexes).
- Introduces explicit stripping of the Polish ł character.


kaplun:
@tiborsimko can you have an opinion over this branch?

tiborsimko:
> can you have an opinion over this branch?

We'd probably have to configure appropriate collation in tabcreate and friends in order to profit from this functionality?

> Lowers and Strips accents also for phrase indexes (and not only word and pair indexes).

This may be dangerous as it may break the accurate display of "nearest phrases" in URLs such as:

- http://demo.invenio-software.org/search?p=Theee&f=title&action_browse=Browse

It would look weird if titles are suddently printed all lowercase...

> WIP

Is this PR still WIP?

Issue #1918, opened 2014-07-09
BibDocFile: Escaping characters in Document Files Manager


switowski:
Hi,

yesterday I've noticed that there is a bug in the Document File Manager that allows user to upload a file with /,(,),. characters in the file name encoded using URL-encoding, for example "art%3A10.1007%2FJHEP01%282013%29057.pdf".
Aside from the fact that if those characters weren't encoded, Document File Manager wouldn't allow to upload files like that (at least not with front-slash in the file name), after a user uploads a file like that then we can have multiple problems:
- If a user deletes the file then the bibupload:FFT task with FIX-MARC is failing with error: `Stage 2B failed: ERROR: while synchronizing 8564 tags: invalid literal for int() with base 10: '12345678/art:10.1007`. It's caused by this function `decompose_file(urllib.unquote(recid_file))`
- Probably the same error might occur when a user is uploading more files to that record.
- Some format elements are not creating a correct URL to the file, for example Default_HTML_files.bft is creating link with decoded URL characters (so something like that: `/record/12345678/files/art:10.1007/JHEP01(2013)057.pdf` and that returns 404 instead of a file).

Tibor mentioned that someone was already working on patching that ?

tiborsimko:
Closed by #3729.

PR #2788, opened 2015-02-17
BibDocFile: disallow percent in filenames

switowski:
- Disallows % character in filenames since it may cause some problems
  with URL encoding/decoding leading to incorrect URLs.
  (addresses #1918)


jirikuncar:
@switowski is this ready for `2.0.0`? (cc @tiborsimko)

tiborsimko:
@jirikuncar This PR is for master and it is prioritised for v1.x, so it should not go to v2.0.x...

jirikuncar:
@tiborsimko ok.

kaplun:
This should have been fixed by @drjova's #2908, isn't it?

drjova:
@kaplun are quite different so I guess we can keep both :smile:, my fix is only when you request the file via url to escape any special character and @switowski's as far as I can see is for sanitizing the file name on submission time.

tiborsimko:

PR #2789, opened 2015-02-17
base: page template block addition

jmartinm:
* Creates an extra block trackingcode to allow addition of
  tracking JavaScript like Piwik without conflicting with the
  usage of JavaScript block in other modules.

### DIFF
--- a/invenio/base/templates/page_base.html
+++ b/invenio/base/templates/page_base.html
@@ -1,6 +1,6 @@
 {#
 ## This file is part of Invenio.
-## Copyright (C) 2012, 2013, 2014 CERN.
+## Copyright (C) 2012, 2013, 2014, 2015 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


invenio/base/templates/page_base.html
```diff
@@ -153,6 +153,7 @@
   {% endblock %}
 {%- endblock %}
 {%- block javascript %}{% endblock %}
+{%- block trackingcode %}{% endblock %}
```
jirikuncar: You can use `javascript`, can't you?

jmartinm: The issue I have seen is that many templates in modules use the ``javascript`` block like https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/deposit/templates/deposit/run_base.html#L67

So if in the overlay template you don't want to use the ``javascript`` part and remove that block/rewrite it you will lose as well the tracking code (which is not desired)

Like this, we can separate the tracking code (see https://github.com/jmartinm/inspire-next/blob/6337772406d9aa40134d4697ccc88a26899bc465/inspire/base/templates/page.html#L50) and don't conflict with ``javascript`` block.

PR #2722, opened 2015-01-29
editor: partial legacy port

jmartinm:
* Moves templates from legacy to Jinja2.

* Moves web interface to Flask views.

* Creates editor bundles.

* Several small changes and bugfixes.

Signed-off-by: Javier Martin Montull <javier.martin.montull@cern.ch>

PR #2790, opened 2015-02-17
formatter: exception fixes

jmartinm:
* Fix formatter bug introduced in https://github.com/inveniosoftware/invenio/pull/2722#discussion-diff-23777064R2021 when no 001 tag is provided.

* Fix on exception formatting.

### DIFF
--- a/invenio/modules/formatter/engine.py
+++ b/invenio/modules/formatter/engine.py
@@ -2018,7 +2018,7 @@ class BibFormatObject(object):
             # If record is given as parameter
             self.xml_record = xml_record
             self.record = create_record(xml_record)[0]
-            recID = int(record_get_field_value(self.record, "001")) or None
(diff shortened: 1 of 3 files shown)


invenio/modules/formatter/engine.py
```diff
@@ -2018,7 +2018,7 @@ def __init__(self, recID, ln=CFG_SITE_LANG, search_pattern=None,
             # If record is given as parameter
             self.xml_record = xml_record
             self.record = create_record(xml_record)[0]
-            recID = int(record_get_field_value(self.record, "001")) or None
+            recID = record_get_field_value(self.record, "001") or None
```
jirikuncar: Since you introduced this in 947910cd2, can you prepare test case for it? Thanks

jmartinm:
Test case added.

PR #2794, opened 2015-02-17
WebStyle: add CERN specific URL handlers

ludmilamarian:
* Adds six CERN specific URL handlers.


ludmilamarian:
This is CDS specific, so it does not make sense anymore to go to the Invenio repo.

PR #2795, opened 2015-02-17
Bibformat: add --missing-formats option

switowski:
* Replaces the --no-missing option with --missing-formats, so by default
  missing formats are not generated.

### DIFF
--- a/.kwalitee.yml
+++ b/.kwalitee.yml
@@ -47,6 +47,7 @@ components:
 - BibSword
 - DocExtract
 - ElmSubmit
+- I18N
 - OAIHarvest
 - OAIRepositoy
 - PdfChecker
--- a/INSTALL
+++ b/INSTALL
@@ -850,7 +850,7 @@ Contents
            you may want to create the above symlink in order to
            configure completion for the inveniocfg command.
 
-Good luck and thanks for choosing Invenio.
+Happy hacking and thanks for flying Invenio.
 
       - Invenio Development Team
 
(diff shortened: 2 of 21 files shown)


kaplun:
Since this branch is proposing a change in the default behavior I think a dummy upgrade recipe should be added spitting out a warning for the admin to amend his/her bibreformat scheduled tasks to add `--missing-formats`. Additionally the default documentation in Invenio advises to just schedule `bibreformat -s5m` and this should do the right thing.

switowski:
@kaplun hmm, that's true. I have changed the code, so now we no longer need any additional changes.
Now, we will skip formatting missing cache if there is a pattern, collection, id, etc. specified, but we will still generate missing cache if the bibformat is run without parameters (or with format parameter).

kaplun:
:shipit:

kaplun:
@tiborsimko this micro-patch can also go into 1.2.0 IMHO

tiborsimko:
- [ ] Please update the commit log message to put the FIX label in front, so that this will be well reported in the automatically generated release notes.

- [ ] Please add `Reviewed-by: Samuele Kaplun <samuele.kaplun@cern.ch>`

switowski:
@tiborsimko Done

egabancho:
Running on CDS

kaplun:
Have to be rebased though...

tiborsimko:
Cherry-picked into `maint-1.2` as 2eb74cc7aa7bd35a76bf991943264aee08fc1c05.

PR #2796, opened 2015-02-18
BibEdit: fix recID type

ludmilamarian:
- Various functions working with the recid expect it to be
  an integer, rather than a string.

### DIFF
--- a/modules/bibformat/lib/elements/bfe_edit_record.py
+++ b/modules/bibformat/lib/elements/bfe_edit_record.py
@@ -36,7 +36,7 @@ def format_element(bfo, style):
     out = ""
 
     user_info = bfo.user_info
-    if user_can_edit_record_collection(user_info, bfo.recID):
+    if user_can_edit_record_collection(user_info, int(bfo.recID)):
         linkattrd = {}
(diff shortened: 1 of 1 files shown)


modules/bibformat/lib/elements/bfe_edit_record.py
```diff
@@ -36,7 +36,7 @@ def format_element(bfo, style):
     out = ""
 
     user_info = bfo.user_info
-    if user_can_edit_record_collection(user_info, bfo.recID):
+    if user_can_edit_record_collection(user_info, int(bfo.recID)):
```
jirikuncar: Shouldn't it be fixed somewhere in [`BibFormatObject`](https://github.com/inveniosoftware/invenio/blob/master/modules/bibformat/lib/bibformat_engine.py#L1870)?

jirikuncar:
@ludmilamarian is the patch still needed or can we close it?

egabancho:
It is deploy on CDS and is a bug fix, thus I is probably relevant for either `legacy` or `maint-1.2`

tiborsimko:

PR #2808, opened 2015-02-20
WIP general: fix LESS_RUN_IN_DEBUG behaviour

### LINKED ISSUE #2777: general: `LESS_RUN_IN_DEBUG` has effect only on first load
Steps to reproduce:

1. Setup an Invenio pu instance as described by `INSTALL.rst`, including the demosite.
2. Have the following additional settings in `invenio.cfg`:
```
    (invenio2)~/.virtualenvs/invenio2/var/invenio.base-instance$ cat invenio.cfg
    [...]
    COLLECT_STORAGE = u'flask.ext.collect.storage.link'
    DEBUG = True
    LESS_RUN_IN_DEBUG = False
    ASSETS_DEBUG = True
    DEBUG_TB_INTERCEPT_REDIRECTS = False
    TESTING = True
```
3. Run `inveniomanage runserver` and see that `less.js` is running:
01_less_working
4. Reload the page:
02_hard_reload
5. Now `less.js` isn't running anymore:
03_less_not_working

It should be noted that restarting the server "fixes" the problem for one more load.

It'd be helpful to have this running so that styles can be more easily debugged using source maps, as described for instance here: http://robdodson.me/debug-less-with-chrome-developer-tools/

jacquerie:
* Allows LESS_RUN_IN_DEBUG to have effect for more than one
   reload.  (fixes #2777)

Signed-off-by: Jacopo Notarstefano <jacopo.notarstefano@cern.ch>

(CC: @jmartinm)

### DIFF
--- a/invenio/ext/assets/extensions.py
+++ b/invenio/ext/assets/extensions.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
+#
 # This file is part of Invenio.
-# Copyright (C) 2012, 2013, 2014 CERN.
+# Copyright (C) 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -134,7 +135,6 @@ class BundleExtension(Extension):
                     bundle.extra.update(static_url_path=static_url_path)
                     if bundle.has_filter("less"):
                         if less_debug:
-                            bundle.filters = None
                             bundle.extra.update(rel="stylesheet/less")
                         else:
                             bundle.extra.update(static_url_path="")


jirikuncar:
@jacquerie what is the status of this pull request? Did @jmartinm or @bouzlibop test it?

jacquerie:
I spoke with @jmartinm and I know he didn't. I don't know about @bouzlibop. In any case, this has low priority for me, since it won't fix the real issue I have: generating the source maps for easier debugging.

bouzlibop:
I gave it a try, but it seems that it's not working for me. Instead of previous behaviour I'm getting this error:

screenshot from 2015-03-06 15 17 21

jacquerie:
Ugh, I was able to reproduce that error for one reload, but now I see it running on every reload.

jmartinm:
I can confirm I get the same error as @bouzlibop 

The correct logic here is to leave:

```python
bundle.filters = None
```

As we don't have to apply any filters on the server side (run lessc) when running in debug mode (all the work is left for the client side ``less.js``).

Somehow this is what is breaking the code on the client side (see the wrong URL in the screenshot: http://gen/invenio.css)

Now the problem I can see here is the following:

1. First time the code runs, [this condition](https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/assets/extensions.py#L135) is ``True`` so the page has the correct ``rel=stylesheet/less`` and all is good.

2. Second time the code runs (after page refresh), the bundles have been cached [here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/assets/extensions.py#L102) so for the same ``invenio.css`` Bundle now [this condition](https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/assets/extensions.py#L135) is not ``True`` because we have set previously [here](https://github.com/inveniosoftware/invenio/blob/pu/invenio/ext/assets/extensions.py#L137) filters to ``None`` so the page has ``rel=stylesheet`` and ``less.js`` does not compile those assets.

@jirikuncar maybe it would be best when running in DEBUG mode to not do any caching? Suggestion on where to do the changes?

invenio/ext/assets/extensions.py
```diff
@@ -134,7 +135,6 @@ def get_bundle(suffix):
                     bundle.extra.update(static_url_path=static_url_path)
```
jirikuncar: I would try to make a `(deep)copy` of `bundle` before modifying the filters.

```python
if env.debug:
    bundle_copy = copy.deepcopy(bundle)
    bundle_copy.extra.update(static_url_path=static_url_path)
    if bundle.has_filter("less"):
        if less_debug:
            bundle_copy.filters = None
            bundle_copy.extra.update(rel="stylesheet/less")
    ...
```

jmartinm:
Thanks @jirikuncar . I have created a new PR https://github.com/inveniosoftware/invenio/pull/2864 which works for me.

jacquerie:
Tested, and working for me.

bouzlibop:
Working, for me too :+1:

jacquerie:
Closed by #2864, I guess.

Issue #2812, opened 2015-02-23
Also use CFG_OAI_ID_FIELD for deduping


aw-bib:
It seems advisable that `find_records_from_extoaiid()` also considers `CFG_OAI_ID_FIELD` for record matching.

Usecase: 
1. join2 repositories harvest each other to get copies from records. We use this process e.g. for authority records.
2. join2 repositories act as institutional bibliographies, this implies to collect all publications from a given institution (even if it's also catalogued at another partner)

In case of _authorities_ the authority ID in 035 ensures deduping. Recent upcoming use cases, however, also involve harvesting of _bibliographic_ records, e.g. in case two institutions from the join2 collaboration work together, we want to ensure, that our scientists have to submit their publications only once. E.g. a scientist from Jülich should only need to submit to JuSER even if he would have to submit to iMPULSE as well. Harvesting ensures that publications from JuSER are exposed to iMPULSE. It doesn't make to much sense to move the OAI-ID to another field in this case. as we're really producing a second instance of the record in another service. At the same time from a bibliographic point of view usage of 035__a for oai-ids will not work in our environment as we need a field for _the id_(tm) (ie. the one that get's stored with the linked records) in case of authority records. This is currently done via 035__.

An alternative, more complex approach might be to allow a list of marc tags for `CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG` probably consolidating with `CFG_BIBUPLOAD_EXTERNAL_SYSNO_TAG`.

join2 might come up with an implementation of the simple case. (`CFG_OAI_ID_FIELD`).

tiborsimko:
Closed by #3721.

PR #2816, opened 2015-02-24
bibupload: Also use CFG_OAI_ID_FIELD for deduping

aw-bib:
In case of replicating datasets between instances by means of oai harvesting it
is plausible to check for external OAI-IDs not only in the field specified by
CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG but also in CFG_OAI_ID_FIELD.

Enh: find_records_from_extoaiid() to call find_record_from_oaiid() in case no
records were found in CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG.

Enh/Fix: find_record_from_oaiid() always returned the first record with
a given OAI-ID assuming that it is unique. However, due to manual intervention
one may have accidentially produced dupes here. Thus check, if more than one
record with an OAI-ID exists. If so, check if all records except one are
deleted. In this case return this surviving record, as cataloguers have resolved
dupes manually. In case we still have dupes resort to the old behaviour for sake of
compatibility, but at least throw a warning. This is somewhat of a
TODO. Probably one should refuse merge here and wait for manual
curation (to be discussed).

addresses #2812

### DIFF
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -655,7 +655,7 @@ def find_records_from_extoaiid(extoaiid, extoaisrc=None):
             # in the locally used CFG_OAI_ID_FIELD that matches. This can happen
             # if oai harvesting is used to duplicate records.
             recid = find_record_from_oaiid(extoaiid)
-            if recid != None:
+            if recid is not None:
                 ret.add(recid)
         return ret
     except Error, error:
@@ -2303,7 +2303,7 @@ def task_run_core():
     return not stat['nb_errors'] >= 1
 
 def log_record_uploading(oai_rec_id, task_id, bibrec_id, insertion_db, pretend=False):
-    if oai_rec_id != "" and oai_rec_id != None:
+    if oai_rec_id != "" and oai_rec_id is not None:
         query = """UPDATE oaiHARVESTLOG SET date_inserted=NOW(), inserted_to_db=%s, id_bibrec=%s WHERE oai_id = %s AND bibupload_task_id = %s ORDER BY date_harvested LIMIT 1"""
         try:
             if not pretend:
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -654,7 +654,9 @@ def find_records_from_extoaiid(extoaiid, extoaisrc=None):
             # no oaiid in CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG. Check if we have one
             # in the locally used CFG_OAI_ID_FIELD that matches. This can happen
             # if oai harvesting is used to duplicate records.
-            ret.add(find_record_from_oaiid(extoaiid))
+            recid = find_record_from_oaiid(extoaiid)
+            if recid != None:
+                ret.add(recid)
         return ret
     except Error, error:
         write_message("   Error during find_records_from_extoaiid(): %s "
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -689,7 +689,7 @@ def find_record_from_oaiid(oaiid):
             for r in res:
                 lst.append('recid:"%s"' % str(r[0]))
             candidates = search_pattern(p=' or '.join(lst))
-            deleted    = search_pattern(p='980__c:"DELETED"')
+            deleted    = search_pattern(p='980:"DELETED"')
             recs = candidates.difference(deleted)
             if len(recs) == 1:
                 # all dupes are resolved, use the not deleted record downstream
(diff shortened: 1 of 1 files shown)


modules/bibupload/lib/bibupload.py
```diff
@@ -674,7 +679,35 @@ def find_record_from_oaiid(oaiid):
...
+            return res[0][0]
+        else:
+            # Principle dupes detected.
+            # If the dupes are resolved only one record should have survived,
+            # and all others are deleted. => get non-deleted record(s)
+            lst = []
+            for r in res:
+                lst.append('recid:"%s"' % str(r[0]))
+            candidates = search_pattern(p=' or '.join(lst))
+            deleted    = search_pattern(p='980__c:"DELETED"')
```
kaplun: Note the defintion of deleted in Invenio is a bit larger: any subfield of 980 could in principle contain the word `DELETED` (at least that is what is assumed as convention in `bibupload`)

aw-bib: So it shold be `980%%:"DELETED"`?

kaplun: It's perfect 980:"DELETED"

aw-bib: >> +            deleted    = search_pattern(p='980__c:"DELETED"')
> 
> It's perfect 980:"DELETED"

This and a minor fix for handling the return of None is in the updated PR. (BTW: shouln't the intbitset.add(None) just work?)
 
-- 
Kind regards,

Alexander Wagner

Deutsches Elektronen-Synchrotron DESY
Library and Documentation

Building 01d Room OG1.444
Notkestr. 85
22607 Hamburg

phone:  +49-40-8998-1758
fax:    +49-40-8994-1758
e-mail: alexander.wagner@desy.de

kaplun: >  (BTW: shouln't the `intbitset.add(None)`  just work?)

Nope, `intbitset.add` expect always a positive integers.

aw-bib: @kaplun yes it is, and I missed it. It was more a suggestion if it wouldn't be good to allow `intibset.add(None)` as a "do nothing". (Looks logical at first glance.)

tiborsimko:
@aw-bib @kaplun I'm not sure whether checking all 980 subfields is the right way...  We have "advertised" 980 $c to be the canonical place to store information about deteled records, e.g. this is what the record merges does, e.g. we have many places in the code that do:

```
if "DELETED" in self.get_field(recID, "980__c"):
```

We'd probably have to amend this more globally...

modules/bibupload/lib/bibupload.py
```diff
@@ -650,6 +650,13 @@ def find_records_from_extoaiid(extoaiid, extoaisrc=None):
                     if extoaisrc is None:
                         write_message('WARNING: Found recid %s for extoaiid="%s" that specify a provenance (%s), while input record does not have a provenance.' % (id_bibrec, extoaiid, this_extoaisrc), stream=sys.stderr)
 
+        if len(ret) == 0:
+            # no oaiid in CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG. Check if we have one
+            # in the locally used CFG_OAI_ID_FIELD that matches. This can happen
+            # if oai harvesting is used to duplicate records.
+            recid = find_record_from_oaiid(extoaiid)
+            if recid != None:
```
tiborsimko: `if recid is not None` is better style, see:

- https://www.python.org/dev/peps/pep-0008/#programming-recommendations

tiborsimko:
@aw-bib Any progress with this PR regarding my comments on 980 DELETED and on recid None? I'd like to release 1.1.7 in the next few days... Can you please have a look or shall we aim at 1.1.8?

aw-bib:
WRT `980__c`/980`: the first version of the PR used only `980__c` the later all `980`. Probably, it's up to a decision what to use.

WRT `is not None` hope I found all of those :)

tiborsimko:
> Probably, it's up to a decision what to use.

980__c would be safer, 980 would be more liberal, at a certain risk.  Well, chances are people aren't misusing the word DELETED in fields like 98012, 980zz, so the change should be basically OK...

aw-bib:
@tiborsimko how to proceed? Should I do anything? Do you want to adopt 980__c?

kaplun:
@tiborsimko, IMHO, although bibedit will write in 980__c, I believe we can safely check upon reading in all 980 subfields. No reasonable admin would  call a collection tag `DELETED`.

This would also follow the https://en.wikipedia.org/wiki/Robustness_principle

tiborsimko:
> how to proceed? Should I do anything?

Nope, I squashed your commits and made a new PR #3721 so that the release notes would be nicely generated. There is nothing more to do on your side 😄

tiborsimko:
Closed by #3721.

PR #2818, opened 2015-02-25
WIP / Superseded script: erroneous use of `CFG_PREFIX`

dset0x:


jirikuncar:
> Fixes references to `CFG_PREFIX`, which in reality referred to the location of installation of invenio executables. These should be in the `PATH` environment variable.

IMHO the `CFG_PREFIX` is fine as it's defined in `invenio/base/config.py`.

dset0x:
@jirikuncar The problem is that, in the code, it is used interchangeably with root (`/` if installing as root outside a virtualenv).

So the binaries would end up in `/usr/local/bin`, because [that's what pip defaults to](https://docs.python.org/2/install/#how-installation-works), while `CFG_PREFIX` would be `/usr`. So code tries to look into `/usr/bin/foo`.

dset0x:
Additionally, will can there ever be a sane reason to edit `CFG_PREFIX`?

jirikuncar:
@dset0x I don't think so. `CFG_PREFIX` should die anyway.

invenio/legacy/inveniocfg.py
```diff
@@ -713,11 +713,11 @@ def cli_cmd_create_demo_site(conf):
             print("ERROR: failed execution of", cmd)
             sys.exit(1)
     cli_cmd_reset_fieldnames(conf) # needed for I18N demo ranking method names
-    for cmd in ["%s/bin/webaccessadmin -u admin -c -r -D" % CFG_PREFIX,
-                "%s/bin/webcoll -u admin" % CFG_PREFIX,
-                "%s/bin/webcoll 1" % CFG_PREFIX,
-                "%s/bin/bibsort -u admin --load-config" % CFG_PREFIX,
-                "%s/bin/bibsort 2" % CFG_PREFIX, ]:
+    for cmd in ("webaccessadmin -u admin -c -a -D",
```
PXke: If I am not wrong this part is the same than https://github.com/inveniosoftware/invenio/pull/2818/files#diff-2d38438784b9f730ac13a86cf37fa97aR59

Do you think it is possible to factorize ?

PXke:
There is a lot of parts like : 

```
 for cmd in ("webaccessadmin -u admin -c -a -D",
    "webcoll -u admin",
    "webcoll 1",
    "bibsort -u admin --load-config",
    "bibsort 2", ):
    if os.system(cmd):
        print("ERROR: failed execution of", cmd)
        sys.exit(1)
```

It will be cool if the code could be factorized by creating a function taking a list of command to execute.

jirikuncar:
> creating a function taking a list of command to execute

The goal is to remove all `os.system` calls in the future.

dset0x:
Please note that killing `CFG_PREFIX` goes hand-in-hand with #2829, which is why I have reflected the changes from this commit in that PR as well.

> The goal is to remove all os.system calls in the future.
In the sense of getting rid of calls to executables or replacing these calls with `invenio.utils.shell:run_shell_command`?

dset0x:
> It will be cool if the code could be factorized by creating a function taking a list of command to execute.
@PXke That would be nice, except I am not sure what a logical name for that list is, or where it would make sense to define it.

@jirikuncar [Comment above](https://github.com/inveniosoftware/invenio/pull/2818#issuecomment-82818932)

jirikuncar:
@dset0x it would be nice not use OS call just to run our own python scripts. What about calling directly the appropriate `main` functions with correct arguments (e.g. something like https://github.com/inveniosoftware/invenio/blob/master/invenio/testsuite/test_manage.py#L63-L92)?

dset0x:
@jirikuncar I'm not sure about the advantages of this approach. Given that these modules do not have a pythonic api (like we are do with new modules), I think it's better if we leave it to the OS to handle them (for example we are not risking some random exception bubbling up to the calling code or crashing).

jirikuncar:
> the advantages of this approach.

You don't have to rely on OS `PATH` and you don't have to instantiate new application contexts over and over again when calling scripts.

>  we are not risking some random exception bubbling up to the calling code or crashing

I see it as positive. If something crashes then we should get clear pythonic exception.

dset0x:
@jirikuncar That was enough to convince me! I'll look into doing it when possible.

dset0x:
Migration notes finalized. This PR now assumes that #2829 gets merged (eg it does not access `CFG_PREFIX` anymore), so consider it blocked until that's done.

PR #2820, opened 2015-02-25
BibDocfile: download all files for a BibRecDoc added

switowski:
* Modifies the getfile() function from WebInterfaceFilesPages class,
  so if the filename starts with "allfiles-", tar archive is created
  and send to the user.

* Allows user to download only those files that are not restricted

* Adds the stream_archive_of_latest_files(), that creates the archive
  with the latest files of a specific format (original or subformat) and
  streams this archive to the user.

* Adds the subformat parameter to the get_total_size_latest_version().
  If that parameter is given, function limits result to this subformat.

* Adds CFG_BIBDOCFILE_SUBFORMATS_TRANSLATIONS to the bibdocfile_config.

### DIFF
--- a/modules/bibdocfile/lib/bibdocfile.py
+++ b/modules/bibdocfile/lib/bibdocfile.py
@@ -59,6 +59,7 @@ import binascii
 import cgi
 import sys
 import copy
+import tarfile
 
 if sys.hexversion < 0x2060000:
     from md5 import md5
@@ -102,6 +103,7 @@ from invenio.bibrecord import record_get_field_instances, \
     encode_for_xml
 from invenio.urlutils import create_url, make_user_agent_string
 from invenio.textutils import nice_size
+from invenio.webuser import collect_user_info
 from invenio.access_control_engine import acc_authorize_action
 from invenio.access_control_admin import acc_is_user_in_role, acc_get_role_id
 from invenio.access_control_firerole import compile_role_definition, acc_firerole_check_user
@@ -123,7 +125,7 @@ from invenio.config import CFG_SITE_URL, \
     CFG_BIBCATALOG_SYSTEM
 from invenio.bibcatalog import BIBCATALOG_SYSTEM
 from invenio.bibdocfile_config import CFG_BIBDOCFILE_ICON_SUBFORMAT_RE, \
-    CFG_BIBDOCFILE_DEFAULT_ICON_SUBFORMAT
+    CFG_BIBDOCFILE_DEFAULT_ICON_SUBFORMAT, CFG_BIBDOCFILE_STREAM_ARCHIVE_FORMATS
 from invenio.pluginutils import PluginContainer
 
 import invenio.template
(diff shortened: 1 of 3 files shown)


egabancho:
Running on CDS

modules/bibdocfile/lib/bibdocfile_config.py
```diff
@@ -69,3 +69,12 @@
 # CFG_BIBDOCFILE_DEFAULT_ICON_SUBFORMAT -- this is the default subformat used
 # when creating new icons.
 CFG_BIBDOCFILE_DEFAULT_ICON_SUBFORMAT = "icon"
+
+# CFG_BIBDOCFILE_SUBFORMATS_TRANSLATIONS -- a list (not dictionary, because
+# we want to preserve the order) that connects the different format sizes
+# (like 'small', 'medium', etc.) with internal format sizes (like 'icon-180', 'icon-640', etc.)
+CFG_BIBDOCFILE_SUBFORMATS_TRANSLATIONS = [
```
tiborsimko: `CFG_BIBDOCFILE_SUBFORMATS_TRANSLATIONS` may sound too generic.  If this variable is not going to be used for any other functionality but the "allfile" exporting, then you may want to rename it in order to indicate its purpose better.  For example `CFG_BIBDOCFILE_STREAM_ARCHIVE_FORMATS` or something of the kind, which would be nicely close to the new function `stream_archive_of_latest_files()` that does the job.

modules/bibdocfile/lib/bibdocfile.py
```diff
@@ -1564,6 +1571,34 @@ def get_text(self, extract_text_if_necessary=True):
 
         return " ".join(texts)
 
+    def stream_archive_of_latest_files(self, req, format=''):
```
tiborsimko: The symbol `format` has a precise meaning in the `bibdocfile` context.  Did you mean "small/medium/large" here?  If so, then the function parameter may rather be called differently than `format`... e.g. in `getfile()` you call these `filesizes`...

modules/bibdocfile/lib/bibdocfile_config.py
switowski: Sure, I'm actually applying some fixes to that commit now (to not export the hidden files), so I will change the name as suggested.

tiborsimko:
```
| BibDocFile: add download all files in BibRecDoc
|     
|     * Modifies the getfile() function from WebInterfaceFilesPages class,
|       so if the filename starts with "allfiles-", tar archive is created
|       and send to the user.
|     
|     * Allows user to download only those files that are not restricted
|     
|     * Adds the stream_archive_of_latest_files(), that creates the archive
|       with the latest files of a specific format (original or subformat) and
|       streams this archive to the user.
|     
|     * Adds the subformat parameter to the get_total_size_latest_version().
|       If that parameter is given, function limits result to this subformat.
|     
|     * Adds CFG_BIBDOCFILE_SUBFORMATS_TRANSLATIONS to the bibdocfile_config.
```

This commit message may sound too "low-level" for casual users. Can you please add a release notes item explaining the functionality of the feature, such as:

```
* NEW Adds a new functionality that permits to obtain tarball of all the files
  associated with a record. For example, `/record/123/files/allfiles` includes
  all the original files in the tarball, `/record/123/files/allfiles-small`
  includes only icons of 180 pixels, etc. The behaviour is governed by internal
  configuration variable `CFG_BIBDOCFILE_STREAM_ARCHIVE_FORMATS`.

```

switowski:
@tiborsimko I have applied all the changes as you have suggested. Please let me know if there is anything else to do here.

switowski:
Note to myself - when this PR is merged, change the CFG_BIBDOCFILE_STREAM_ARCHIVE_FORMATS variable name (in bibdocfile and in bfe_CERN_download_all_photos.py) to a new name to avoid problems in the future.

tiborsimko:
One more cosmetics: `(closes ...)` should not be on an item of its own, but in the preceding item, after the phrase ends.  Can you please amend and repush?

switowski:
@tiborsimko done

tiborsimko:
Thanks, cherry-picked into `legacy` as fa009b4.

PR #2822, opened 2015-02-26
BibField: bug fixes

egabancho:


egabancho:
This PR is only relevant for CDS, it breaks some tests cases and it might not be useful to others

PR #2830, opened 2015-03-03
installation: fix missing pkg-config dependency

brucellino:
The pkg-config package needs to be installed, else `pip install -r requirements.txt` will fail when installing matplotlib. See https://github.com/matplotlib/matplotlib/issues/3029/ and https://github.com/brucellino/zenodo-docker-role/issues/3

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -36,6 +36,7 @@ Contents
               python-4suite-xml python-simplejson python-xml \
               gnuplot poppler-utils \
               gs-common clisp gettext libapache2-mod-wsgi unzip \
+              pkg-config \
               python-dateutil python-rdflib python-pyparsing \
               python-gnuplot python-magic pdftk html2text giflib-tools \
               pstotext netpbm python-pypdf python-chardet python-lxml \


jirikuncar:
@brucellino thank you for your fix.

- [ ] Please add yourself to `AUTHORS` file;
- [ ] check our commit message style (`git log`, `git commit --amend -s`);
- [ ] rebase your branch on top of `next` branch (`git fetch upstream`, `git rebase -i upstream/next`).

INSTALL
```diff
@@ -36,6 +36,7 @@ Contents
               python-4suite-xml python-simplejson python-xml \
               gnuplot poppler-utils \
               gs-common clisp gettext libapache2-mod-wsgi unzip \
+              pkg-config \
```
tiborsimko: BTW this `aptitude` string is obsolete in the `next` branch; it is a remainder from the `master` branch.
(Even there, in the `master` branch, it is recommended to use `pip` rather than system packages.)

brucellino: Hi @tiborsimko 

Thanks for the response. I found that without this line, pip didn't work - is there a corresponding pip package that should be called ?
Also, forgot to add : 

Distributor ID:	Ubuntu
Description:	Ubuntu 14.10
Release:	14.10
Codename:	utopic
(perhaps the newer ubuntu is causing problems?)

jirikuncar: Have you tried it with latest `next` branch? The installation process has been updated.

brucellino:
Invalid since I was using the `master` branch - can't verify on `next` which is good :)

PR #2840, opened 2015-03-04
installation: package.json addition

jmartinm:
* Adds package.json to simplify installation of npm packages.

* This allows as well to control what versions we install.

* Updates installation guide accordingly.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -458,12 +458,9 @@ they are not in the environment ``$PATH`` already.
 
 .. code-block:: console
 
-    # Global installation
-    $ sudo su -c "npm install -g less clean-css requirejs uglify-js"
-
-    # or
-    # Local installation
-    (invenio)$ npm install less clean-css requirejs uglify-js
+    # Local installation (using package.json)
+    (invenio)$ cdvirtualenv src/invenio
+    (invenio)$ npm install
     (invenio)$ inveniomanage config set LESS_BIN `find $PWD/node_modules -iname lessc | head -1`
     (invenio)$ inveniomanage config set CLEANCSS_BIN `find $PWD/node_modules -iname cleancss | head -1`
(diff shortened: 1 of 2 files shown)


jmartinm:
This was started because of problems on ``clean-css``.

Installing latest version causes problems as version 3 is quite different to version 2 (https://github.com/jakubpawlowicz/clean-css#how-to-upgrade-clean-css-from-2x-to-3x)

So the overlay css was not taking precedence over the Invenio one any more.

jacquerie:
Another argument for having this is the (possible) future addition of the Karma test runner, which ships plugins for PhantomJS or Chrome integrations as Node modules.

package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "less": "*",
```
lnielsen: I think we should be using less 1.7.5

INSTALL.rst
```diff
@@ -458,12 +458,9 @@ they are not in the environment ``$PATH`` already.
...
 
-    # Global installation
-    $ sudo su -c "npm install -g less clean-css requirejs uglify-js"
-
-    # or
-    # Local installation
-    (invenio)$ npm install less clean-css requirejs uglify-js
+    # Local installation (using package.json)
+    (invenio)$ cdvirtualenv src/invenio
+    (invenio)$ npm install
```
lnielsen: perhaps ``-g``?

lnielsen:
:+1:

INSTALL.rst
jmartinm: I don't think ``-g`` is supported from a ``package.json`` ? At least it did not work for me and could not find online how to do it.

http://stackoverflow.com/questions/6480549/install-dependencies-globally-and-locally-using-package-json

package.json
jmartinm: Have you found problems with newer version?

PR #2848, opened 2015-03-05
MiscUtils: urlutils: escape unsafe characters

drjova:
* Adds escaping url from unsafe characters.

* Updates bibdocfile templates.

Signed-off-by: Harris Tzovanakis <me@drjova.com>

### DIFF
--- a/modules/bibdocfile/lib/bibdocfile_templates.py
+++ b/modules/bibdocfile/lib/bibdocfile_templates.py
@@ -22,7 +22,7 @@ from invenio.config import CFG_SITE_URL, \
      CFG_SITE_LANG, CFG_SITE_RECORD, CFG_INSPIRE_SITE
 from invenio.messages import gettext_set_language
 from invenio.dateutils import convert_datestruct_to_dategui
-from invenio.urlutils import create_html_link
+from invenio.urlutils import create_html_link, url_safe_escape
 from invenio.textutils import nice_size
 
 def list_types_from_array(bibdocs):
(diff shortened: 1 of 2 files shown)


kaplun:
@drjova if this is needed to put a URL inside an `src` HTML attribute it should be enough to use standard `cgi.escape(url, quote=True), isn't it? That's how it is done everywhere else in Invenio...

drjova:
@kaplun  Unfortunately `cgi.escape` is not enough for our case which some files have spaces in their file name.

``` python

>> a = "hello is it me that looking for"
>> In [1]: cgi.escape(a, quote=True)
>> Out[1]: 'hello is it me that looking for'
>> In [2]: quote(b, safe="%/:=&?~#+!$,;'@()*[]")
>> Out[2]: 'hello%20is%20it%20me%20that%20looking%20for'
```

I couldn't find anything similar in the invenio utils :/

kaplun:
Good point. But then it would be better to correct this centrally. I.e. in `bibdocfile.BibDocFile.__init__()` where these URLs are generated. In this way all the places were BibDocFile URLs are used are going to see the correct version (e.g. when they are written in MARCXML, etc.)

This happens in https://github.com/inveniosoftware/invenio/blob/maint-1.2/modules/bibdocfile/lib/bibdocfile.py#L2977-L2982

drjova:
:+1: I'll prepare a PR

drjova:
PR#2908

kaplun:
@drjova can we close this PR, as superseeded by #2908?

drjova:
@kaplun yes!  

superseeded-by https://github.com/inveniosoftware/invenio/pull/2908

PR #2849, opened 2015-03-05
Template: Adds 'characters' filter (similar to sentences)

blixhavn:
* Adds a 'characters' filter, where the input string is truncated
  by `limit` characters, rounded up to nearest space (' ') and
  extended with `extension` (default ' [...]').

### DIFF
--- a/invenio/ext/template/__init__.py
+++ b/invenio/ext/template/__init__.py
@@ -170,6 +170,12 @@ def setup_app(app):
         """Return first `limit` number of sentences ending by `separator`."""
         return separator.join(value.split(separator)[:limit])
 
+    @app.template_filter('characters')
+    def _characters(value, limit, extension=' [...]'):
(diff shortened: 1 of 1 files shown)


blixhavn:
Well, which component names are recognized?

invenio/ext/template/__init__.py
```diff
@@ -170,6 +170,12 @@ def _sentences(value, limit, separator='. '):
         """Return first `limit` number of sentences ending by `separator`."""
         return separator.join(value.split(separator)[:limit])
 
+    @app.template_filter('characters')
+    def _characters(value, limit, extension=' [...]'):
```
jirikuncar: IMHO `truncate` http://jinja.pocoo.org/docs/dev/templates/#truncate should do the job

blixhavn: Heh, would you look at that. Never mind then :)

PR #2850, opened 2015-03-05
upgrader: invenio_2_0_0 release recipe

jirikuncar:


jirikuncar:
It might not be necessary not to create dependencies.

PR #2855, opened 2015-03-06
base: foreign key check workaround removal

hachreak:
* Remove the workaround to drop the database in mysql.


hachreak:
merged!

PR #2864, opened 2015-03-09
assets: less debug fix

### LINKED ISSUE #2777: general: `LESS_RUN_IN_DEBUG` has effect only on first load

jmartinm:
* Avoids bundle changes from being cached when running in debug mode.
  (fixes #2777)

### DIFF
--- a/invenio/ext/assets/extensions.py
+++ b/invenio/ext/assets/extensions.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
+#
 # This file is part of Invenio.
-# Copyright (C) 2012, 2013, 2014 CERN.
+# Copyright (C) 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -18,6 +19,7 @@
 
 """Custom `Jinja2` extensions."""
 
+import copy
 import os
 import six
 from flask import current_app, _request_ctx_stack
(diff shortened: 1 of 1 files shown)


invenio/ext/assets/extensions.py
```diff
@@ -127,23 +129,28 @@ def get_bundle(suffix):
...
                         if less_debug:
-                            bundle.filters = None
-                            bundle.extra.update(rel="stylesheet/less")
+                            bundle_copy.filters = None
+                            bundle_copy.extra.update(rel="stylesheet/less")
                         else:
-                            bundle.extra.update(static_url_path="")
-                    if bundle.has_filter("requirejs"):
+                            bundle_copy.extra.update(static_url_path="")
+                    if bundle_copy.has_filter("requirejs"):
```
jirikuncar: This one can be just `bundle`.

invenio/ext/assets/extensions.py
```diff
@@ -127,23 +129,28 @@ def get_bundle(suffix):
...
                 # beginning of the assets in ran in debug mode as well as
                 # killing the filters if they are not wanted in debug mode.
                 if env.debug:
-                    bundle.extra.update(static_url_path=static_url_path)
-                    if bundle.has_filter("less"):
+                    # Create a deep copy to avoid filter removal from
+                    # being cached
+                    bundle_copy = copy.deepcopy(bundle)
+                    bundle_copy.extra.update(static_url_path=static_url_path)
+                    if bundle_copy.has_filter("less"):
```
jirikuncar: I would keep this just `bundle`.

PR #2871, opened 2015-03-11
installation: better description and pip tests target

tiborsimko:
* Expands project description in place of "Invenio is Fun".

* Adds `tests` target to `extras_require` so that developers could
  install test-related dependencies via `pip install -e .[tests]`.

* Improves PEP-8 compliance and code kwalitee.


jirikuncar:
`setup` -> `installation` then :shipit:

Issue #2853, opened 2015-03-05
RFC Dropping INSPIRE-centric legacy code


kaplun:
Part of Invenio code is completely INSPIRE specific and went into the `v1.2.0` (future former `master` branch) partially because of the limitations in plugging optional functionalities in Invenio. Nowadays thanks to the modular approach of `v2.0.0` it is possible to separate such code and move it where it is due to, i.e. the INSPIRE overlay.

This parts amount to:
- [x] **pdfchecker** (a completely INSPIRE-specific module, needed to refresh a given HEP record related to an arXiv record, upon arXiv updates)
- [ ] **HEPData** (this functionality is scattered around in miscutils and search engine, in particular because it introduces a new *Data* tab in the detailed record view.)
- [ ] **plotextractor** (a completely INSPIRE-specific module, upon arXiv updates)

lnielsen:
+1

I guess this could even be expanded to include remove instance specific code from Invenio (e.g. CFG_CERN_SITE and friends):
```
$ git grep 'if CFG_CERN_SITE' | wc -l
     102
```

jalavik:
:+1:  Indeed `CFG_XXX_SITE` should be purged where feasible for `1.x` and should exist nowhere in `>2.x`.

PXke:
+1 Really good ! If possible it would be still cool if the specific INSPIRE modules could be avaiable separately for if needed plug them in an instance through an overlay.

tiborsimko:
Is anyone already working on implementing this, considering the target v2.1?

hachreak:
see also `git grep "CERN.*hack"`

jalavik:
I can volunteer for the pdfchecker and plotextractor. Anything HepData we can just kill outright as it is re-implemented by @eamonnmag and on INSPIRE Labs we'll reuse any widgets from there.

tiborsimko:
Superseded by Invenio 3.0.

PR #2873, opened 2015-03-11
accounts: profile update and password hashing

lnielsen:
* Adds support for granting author/viewer rights to records via
  tags by specifying CFG_ACC_GRANT_AUTHOR_RIGHTS_TO_USERIDS_IN_TAGS
  and/or CFG_ACC_GRANT_VIEWER_RIGHTS_TO_USERIDS_IN_TAGS.

* Adds support for granting access to records in home collection prior
  to indexing by webcoll if home collection is defined by a dbquery
  that uses pure MARC tags.

* Fixes double message flashing issues.

* Fixes issue with empty records not returning 404.

* Adds support for allowing users to update their profile (nickname,
  email, family name and given name).

* Adds support for users to re-request and verification email to be
  sent.

* Adds new Passlib Flask extension to support configurable password
  contexts in Invenio.

* NOTE Changes user model fields family name/given names to store
  empty string as default instead of null.

* NOTE Adds support for users to change email address/nickname. If you
  store email addresses in e.g. records or fireroles you are
  responsible for propagating the users change of email address by
  adding listeners to the 'profile-updated' signal. Alternatively you
  can migrate records (using
  CFG_ACC_GRANT_AUTHOR_RIGHTS_TO_USERIDS_IN_TAGS and
  CFG_ACC_GRANT_VIEWER_RIGHTS_TO_USERIDS_IN_TAGS) and fireroles (using
  "allow/deny uid <uid>") to restrict access based on user id instead
  of user email address.

* NOTE Refactors password hashing to (a) explicitly specify password
  salt instead of relying on the email address, since a change of
  email would cause the password to be invalidated (b) support
  multiple password hashing algorithms concurrently (c) automatic
  migration of deprecated hashes when users logs in (d) allows
  overlays to specify their preferred hashing algorithms

* NOTE Deprecates legacy Invenio's hashing algorithm based on AES
  encryption of email address using the password as secret key in
  favor of SHA512 using random salt and 100000 rounds.


invenio/ext/login/__init__.py
```diff
@@ -26,6 +26,7 @@
 from flask import flash, g, redirect, request, session, url_for
 from flask.ext.login import (
     LoginManager,
+    current_app,
```
jirikuncar: Shouldn't this be imported from `flask` itself?

lnielsen: Ups, yes it should.

invenio/base/templates/_formhelpers_base.html
```diff
@@ -23,7 +23,7 @@
   {% set label_size = fieldkwargs.get('label_size') %}
   {% set field_size = fieldkwargs.get('field_size') %}
   {% set with_label = fieldkwargs.pop('with_label', True) %}
-  {%- if field.name == "csrf_token" or field.type == 'HiddenField' -%}
+  {%- if field.type == "CSRFTokenField" or field.type == 'HiddenField' -%}
```
lnielsen: We should find a different way to do this, since ``field.type == 'HiddenField'`` doesn't work for subclasses of ``HiddenField``.

invenio/ext/login/__init__.py
```diff
@@ -126,10 +130,8 @@ def authenticate(nickname_or_email=None, password=None,
         return False
 
     if user.note == '2':  # account is not confirmed
-        logout_user()
```
lnielsen: Note: This statement I don't think have had any effect since control flow continues down to the return statement below (``return login_user(..)``). Probably we should have a configuration variable to decide if a user can login without email confirmation, but I think that requires decomposing ``user.note`` into separate fields (e.g. active and confirmed booleans).

invenio/ext/login/__init__.py
```diff
@@ -151,11 +153,11 @@ def do_login_first(error=401):
         if not urllib.unquote(secure_url).startswith(request.base_url):
             return redirect(secure_url)
         if current_user.is_guest:
-            flash(g._("Please sign in to continue."), 'info')
+            if not session.get('_flashes'):
```
lnielsen: Checks if any message have been flash, if not it flashes one. Not sure if this is the best way to do it.

invenio/ext/passlib/hash.py
```diff
@@ -0,0 +1,74 @@
...
+    An upgrade recipe is provided to migrated existing binary password hashes
+    to hashes of this engine.
+    """
+
+    name = "invenio_aes_encrypted_email"
+
+    def _calc_checksum(self, secret):
+        """Calculate string."""
+        return str_to_uascii(
+            hashlib.sha256(mysql_aes_encrypt(self.user, secret)).hexdigest()
```
lnielsen: Note, in addition to Invenio's normal AES encryption of the email using the password as secret key, I additionally run it through a sha256 just to produce a string of a specific length instead of variable length encrypted values. There's an upgrade in this PR which migrates existing hashes so this is transparent to installations.

invenio/legacy/search_engine/__init__.py
```diff
@@ -312,10 +315,54 @@ def get_restricted_collections_for_recid(recid, recreate_cache_if_needed=True):
...
 
+
+def check_email_or_group(user_info, email_or_group):
+    """Check user email or group matches."""
+    if email_or_group in user_info['group']:
+        return True
+    email = email_or_group.strip().lower()
+    if user_info['email'].strip().lower() == email:
+        return True
+    if CFG_CERN_SITE:
```
lnielsen: Should we remove this straight away (re. RFC #2853).

invenio/legacy/search_engine/__init__.py
```diff
@@ -394,7 +433,7 @@ def check_user_can_view_record(user_info, recid):
         ## Perfect! It's authorized then!
         return (0, '')
 
-    restricted_collections = get_restricted_collections_for_recid(recid, recreate_cache_if_needed=False)
+    restricted_collections = get_restricted_collections_for_recid(recid, recreate_cache_if_needed=True)
```
lnielsen: This is a module level cache, which means that the cache is not cleaned until the process is cleaned, hence you not only had to wait for webcoll/bibindex to run, you also had to wait for the process to be restarted.

invenio/legacy/search_engine/__init__.py
```diff
@@ -3952,7 +3991,18 @@ def record_public_p(recID, recreate_cache_if_needed=True):
     """Return 1 if the record is public, i.e. if it can be found in the Home collection.
        Return 0 otherwise.
     """
-    return recID in get_collection_reclist(CFG_SITE_NAME, recreate_cache_if_needed=recreate_cache_if_needed)
+    if recID in get_collection_reclist(CFG_SITE_NAME, recreate_cache_if_needed=recreate_cache_if_needed):
+        return True
+
+    if cfg.get('CFG_WEBSEARCH_PUBLIC_RECORD_EXTRA_CHECK', False):
```
lnielsen: This check is disabled by default since it requires the root collection to be defined using MARC tags instead of indexes. When a record is uploaded and before it has been indexed by webcoll/bibindex, we allow to display the record in case it is in the root collection.

invenio/modules/records/views.py
```diff
@@ -73,13 +78,7 @@ def decorated(recid, *args, **kwargs):
         if not current_user.is_super_admin and 'verbose' in kwargs:
             kwargs['verbose'] = 0
 
-        if auth_code and current_user.is_guest:
```
lnielsen: Add extra note in commit message about incompatibility or better.

invenio/base/templates/_formhelpers_base.html
jmartinm: Maybe you can use https://flask-wtf.readthedocs.org/en/v0.8.3/#creating-forms

```python
<form method="POST" action=".">
    {{ form.hidden_tag() }}
```
to add all hidden fields (including CSRFTokenField)

Also to see if a field is hidden maybe better:

```python
field.widget.input_type == 'hidden'
```

lnielsen: Great. Thanks. Yes, I think that would be better.

invenio/legacy/search_engine/__init__.py
PXke: Yep !

invenio/modules/accounts/forms.py
```diff
@@ -31,13 +31,83 @@
...
 from .models import User
 from .validators import validate_email, validate_nickname, \
     validate_nickname_or_email, wash_login_method
 
 
+def nickname_validator(form, field):
+    """Validate nickname."""
+    validate_nickname(field.data)
+    # is nickname already taken?
+    try:
```
PXke: I know that Python is "Easier to ask for forgiveness than permission."

But I must admit that it is a bit tricky to read quickly.  

My 2cents but I think it will be better to do an 

except NoResultFound ( perhaps even MultipleResultsFound, even if in theory it should never happen)
Because there we will caught all the exception fro SQLAlchemy even the one that could be terribly annoying. 
Wouldn't it be a good idea to register the exception in this case ? (with a second except with SQLAlchemyError ?)

invenio/modules/accounts/models.py
```diff
@@ -85,12 +90,74 @@ def password(self):
...
 
-    @password.comparator
-    def password(self):
-        """Compare password."""
-        return self._password_comparator
+        return True
+
+    def verify_email(self, force=False):
+        """Verify email address."""
+        if force or self.note == "2":
```
PXke: I don't understand this part could you please enlight me ?

invenio/modules/accounts/upgrades/accounts_2015_03_06_passlib.py
```diff
@@ -0,0 +1,86 @@
...
+
+import warnings
+from binascii import hexlify
+from sqlalchemy import *
+from invenio.ext.sqlalchemy import db
+from invenio.modules.upgrader.api import op
+from invenio.utils.text import wait_for_user
+from sqlalchemy.dialects import mysql
+
+import hashlib
```
PXke: I am picky but shouldn't it place at the same level than warnings ? :cake:

invenio/modules/accounts/upgrades/accounts_2015_03_06_namedefaults.py
```diff
@@ -0,0 +1,57 @@
...
+               nullable=False,
+               server_default='')
+    op.alter_column('user', 'given_names',
+               existing_type=mysql.VARCHAR(length=255),
+               nullable=False,
+               server_default='')
+
+def estimate():
+    """Estimate running time of upgrade in seconds (optional)."""
+    return 1
```
PXke: Remember me http://xkcd.com/221/ :smile:

invenio/modules/accounts/models.py
jirikuncar: @PXke you can happily ignore this line as the `User.note` is going to be splitted to several columns like: email confirmed, user activated, etc.

PXke: Thanks ! :+1:

lnielsen: Yep, legacy Invenio knowledge :-) You just have to know the magic numbers :-)

FYI: We are looking into using Flask-Security for all views and just provide the underlying db models.

invenio/modules/accounts/upgrades/accounts_2015_03_06_namedefaults.py
lnielsen: Hehe ;-) Yes, actually it would be easier not having any estimates for upgrades :-) It's anyway pretty hard to make a reliable guesstimate.

invenio/modules/accounts/upgrades/accounts_2015_03_06_passlib.py
lnielsen: Yes :-)

invenio/modules/accounts/forms.py
lnielsen: True, always better to catch the most specific exceptions.

lnielsen:
Thanks for the reviews. I'll update the PR with the comments, plus I've found some more issues.

invenio/modules/records/views.py
```diff
@@ -65,6 +65,11 @@ def decorated(recid, *args, **kwargs):
         g.collection = collection = Collection.query.filter(
             Collection.name == guess_primary_collection_of_a_record(recid)).\
             one()
+        g.bibrec = Bibrec.query.get(recid)
+
+        record = get_record(recid)
```
lnielsen: @blixhavn I think this part here fixes the problem you have reported in #2891

blixhavn: No, because it is line 66 that throws the exception.

Traceback:
```
File "/home/tind/.virtualenvs/invenio/local/lib/python2.7/site-packages/invenio/modules/records/views.py", line 66, in decorated
Collection.name == guess_primary_collection_of_a_record(recid)).
File "/home/tind/.virtualenvs/invenio/local/lib/python2.7/site-packages/sqlalchemy/orm/query.py", line 2404, in one
raise orm_exc.NoResultFound("No row was found for one()")
```

lnielsen: Ok so perhaps I just move it higher?

blixhavn: Yea, that should work

invenio/modules/accounts/models.py
PXke: Flask-Security seems to be a cool thing :+1:

lnielsen:
Rebased to master.

invenio/ext/login/__init__.py
```diff
@@ -126,10 +129,12 @@ def authenticate(nickname_or_email=None, password=None,
         return False
 
     if user.note == '2':  # account is not confirmed
+        flash(_("You have not yet verified your email address.", ), 'warning')
```
jirikuncar: `... address.", ), ...` do we need the comma before bracket?

invenio/legacy/search_engine/__init__.py
```diff
@@ -454,7 +457,18 @@ def record_public_p(recID, recreate_cache_if_needed=True):
...
+        # Perhaps webcoll didn't run yet.
+        from invenio.modules.search.models import Collection
```
jirikuncar: `from invenio.modules.collections.models import Collection`

invenio/legacy/search_engine/__init__.py
```diff
@@ -454,7 +457,18 @@ def record_public_p(recID, recreate_cache_if_needed=True):
     """Return 1 if the record is public, i.e. if it can be found in the Home collection.
        Return 0 otherwise.
     """
-    return recID in get_collection_reclist(CFG_SITE_NAME, recreate_cache_if_needed=recreate_cache_if_needed)
+    if recID in get_collection_reclist(CFG_SITE_NAME, recreate_cache_if_needed=recreate_cache_if_needed):
+        return True
+
+    if cfg.get('CFG_WEBSEARCH_PUBLIC_RECORD_EXTRA_CHECK', False):
+        # Perhaps webcoll didn't run yet.
```
jirikuncar: There is no `webcoll` anymore.

invenio/ext/login/__init__.py
```diff
@@ -151,11 +156,11 @@ def do_login_first(error=401):
         if not urllib.unquote(secure_url).startswith(request.base_url):
             return redirect(secure_url)
         if current_user.is_guest:
-            flash(g._("Please sign in to continue."), 'info')
+            if not session.get('_flashes'):
+                flash(g._("Please sign in to continue."), 'info')
             from invenio.modules.accounts.views.accounts import login
-            return login(referer=request.url), 401
+            return login(referer=request.url)
```
jirikuncar: @lnielsen are you sure that we should not keep the 401 error code? Do you have any use where it causes troubles?

invenio/ext/sqlalchemy/__init__.py
```diff
@@ -43,6 +43,7 @@
                                           LegacyBigInteger)
 
 
+
```
jirikuncar: probably not needed

invenio/modules/accounts/models.py
```diff
@@ -85,12 +90,74 @@ def password(self):
...
+        changed_attrs = {}
+        for field in self.PROFILE_FIELDS:
+            if field in data and getattr(self, field) != data[field]:
+                changed_attrs[field] = getattr(self, field)
+                setattr(self, field, data[field])
+
+        if 'email' in changed_attrs:
+            self.verify_email(force=True)
+
+        db.session.commit()
```
jirikuncar: `try:`, `except ...`, `db.session.rollback()`?

invenio/ext/login/__init__.py
jirikuncar: found it here https://github.com/inveniosoftware/invenio/pull/2873/files#diff-d0cf81b2920b7bd09d191730c0433305R101 :+1:

invenio/modules/accounts/views/accounts.py
```diff
@@ -312,18 +313,27 @@ def lost():
 def access():
     try:
         mail = mail_cookie_check_mail_activation(request.values['mailcookie'])
-        User.query.filter(User.email == mail).one().note = 1
+
+        u = User.query.filter(User.email == mail).one()
+        u.note = 1
         try:
             db.session.commit()
         except:
```
jirikuncar: can you add `SQLAlchemyError`?

invenio/modules/records/access.py
```diff
@@ -133,7 +163,7 @@ def check_user_can_view_record(user_info, recid):
         return (0, '')
 
     restricted_collections = get_restricted_collections_for_recid(
-        recid, recreate_cache_if_needed=False
+        recid, recreate_cache_if_needed=True
```
jirikuncar: this might slow down things badly. do you recall the reason why did you change it?

jirikuncar:
@lnielsen I went quite carefully through the code and identified few :hankey: places. It might be enough mostly to clarify the reason behind some changes. Thanks!

invenio/ext/login/__init__.py
lnielsen: Yep, it's fixed in the Zenodo branch which I'll have to merge back in tomorrow.

invenio/modules/records/access.py
lnielsen: Yep, the never ending story about restricted public collections.

lnielsen: I just had an extra look without running the code, but as far as I remember, it's when you upload new records. First bibupload runs, and part of this fix ensures that a record which is in the root collection can be displayed even though webcoll haven't run. Now webcoll runs and assigns the record to a multiple restricted collections (the restricted root collection is authorised to guest).
Now you try to display the record, which means that ``check_user_can_view_record()`` is called. Without recreating the cache, the root collection cache does not contain the new record (even though webcoll have just assigned it). The other collection caches however do contain the record. The result is that the root collection is not included in the list of restricted collections that the record is part of, which means that a guest user (who is only authorised to the root restricted collection) will not be authorised to see the record because he/she doesn't have access to the other restricted collections.

All in all, this is a symptom fix for a much bigger issue of the underlying authorisation system and cache and search.

jirikuncar: This should not be a problem anymore on 2.1 if `indexer` clears collection caches after it runs.

lnielsen:
I've managed to take out the parts that needs further testing after the rebase to master. Please take a look.

invenio/modules/accounts/templates/accounts/emails/activation.tpl
```diff
@@ -0,0 +1,5 @@
+{% trans site_name=config.CFG_SITE_NAME_INTL.get(g.ln) %}Thanks for registering with {{site_name}}! To complete the registration, please validate your email address by clicking the link below:{% endtrans %}
```
jirikuncar: missing license

jirikuncar:
:shipit:

PR #2883, opened 2015-03-13
jsonalchemy: fix setting default value for dicts

MSusik:


invenio/modules/jsonalchemy/testsuite/fields/oldvalue.cfg
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+##
```
jirikuncar: please use only one `#` char. thank you

Issue #2856, opened 2015-03-06
RFC richer structure in commit log messages 


tiborsimko:
## Background

Invenio commit log messages have standardised format described in:

- https://github.com/inveniosoftware/invenio/blob/master/CONTRIBUTING.rst (recommendation 6)
- http://invenio-software.org/wiki/Tools/Git/Workflow#R2.Remarksoncommitlogmessages

Besides better QA, better readability, easier future bisecting, etc,
this permits also to prepare release notes in an automated manner.

This works OK for small incremental update releases with 100+
commits:

- https://raw.githubusercontent.com/inveniosoftware/invenio/v1.1.4/RELEASE-NOTES

but it does not work so well for large feature releases consisting of 1000+
commits, as the recent v1.2.0 and v2.0.0 releases:

- https://raw.githubusercontent.com/inveniosoftware/invenio/v1.2.0/RELEASE-NOTES
- https://raw.githubusercontent.com/inveniosoftware/invenio/v2.0.0/RELEASE-NOTES

The problems include:

- Shortness of messages.  The commit message headline is limited to
  some 50 characters, which is good for being concise, however this is
  often not sufficient to express given change with reasonable
  clarity to the end users.

- Nature of changes.  We have started to use "NOTE" in the commit logs
  to attract attention to important (compatibility) changes, however
  this is not enough to inform the user about the nature of the
  change.  "Is it a new feature?  A bug fix?  Or just some internal
  refactoring that I don't have to care about?"

- One feature branch may consist of many commits, while for the
  reporting we may only need one item.  Some of the commits may be
  better "silenced" in the high-level end-user report.

## Proposal

Let's introduce richer structural markup in our commit log messages
that would address these drawbacks so that we could generate nicer,
more informative *progress reports* in between releases, and *release
notes* later.

Sphinx can serve as an example of what I have in mind:

- http://sphinx.readthedocs.org/en/latest/changes.html#release-1-2-3-released-sep-1-2014

There are basically two approaches how to address this: (1) via commit
log messages; (2) via tickets (issues, pull requests).

### Option 1: use richer structure in commit messages

We could use more leading labels besides NOTE, for example:

```
foo: better support for xyzzy and zyxxy

* FEATURE Foo now supports xyzzy in all the user facing web pages.

* FIX Corrects interface to Zyxxy...

* INCOMPATIBLE Removes foo...

Reviewed-by: John Doe
```

(Side note: instead of textual labels, we could use a subset of
[...]

PR #2898, opened 2015-03-18
BibDocFile: safer new bibdocfile model upgrade

kasioumis:
* Uses the get() method instead of __getitem__() defaulting the return
  value to an empty dictionary in case of missing keys in more_info.


tiborsimko:
Thanks, look good.  Please update module name in the commit message, and add FIX label as per #2856; for example:

```
BibDocFile: safer new bibdocfile model upgrade

* FIX Safer upgrade recipe for migrations from the old document storage
  model (used in v1.1) to the new document storage model (used in v1.2).

* Uses the get() method instead of __getitem__() defaulting the return
  value to an empty dictionary in case of missing keys in more_info.

Reviewed-by: Tibor Simko <tibor.simko@cern.ch>
```

kasioumis:
@tiborsimko Thank you! In fact I'll use your example commit message as is, I don't see how it could be any better :-)

PR #2823, opened 2015-02-26
WIP relationships: picking up MSusik's work

jacquerie:
In this PR I will continue @MSusik's work on the `relationships` module than began in #2719. Here's a tentative roadmap:

- [x] Rebase on top of current `pu`
- [ ] Refactor the methods in `invenio/modules/relationships/engines/sqlalchemy.py`, and design a comprehensive API

(I will replace the `inwards` and `outwards` boolean parameters with the methods `in_degree`, `out_degree`, `in_edges`, and `out_edges`; I also think that in some cases the `neighbours` method should return a `list` and not a `set`, and I'm thinking how to accomplish that. I'm reading #2573 to understand all requirements.)

- [ ] Wait for `invenio/modules/jsonalchemy` to be fully evicted from Invenio (#2815) 
- [ ] Work around the fact that the code is currently using the internals of `jsonalchemy`.

I'm not sure about the rest. I've postponed fixing https://github.com/inveniosoftware/invenio/compare/inveniosoftware:pu...jacquerie:rebase-relationships?expand=1#diff-53f9b6baea75d29500a64134c2d2b8efR26, and I might try to avoid the cute but complex metaclass trick of https://github.com/inveniosoftware/invenio/compare/inveniosoftware:pu...jacquerie:rebase-relationships?expand=1#diff-c0075666ce0e4fba6cbd934a9f84ade7R39.

[skip ci]

PR #2903, opened 2015-03-18
WIP relations: initial implementation

jacquerie:
I can (finally) show some of my work. I decided to abandon the approach outlined in #2823 and start anew; I actually ended up reusing a lot of @MSusik's code anyway.

The main idea is that the code is not using metaclasses, but simple inheritance; which class holds the actual implementation is dynamically decided by a few `import_string`s. The implementation still resides in `invenio/modules/relationships/engines/sqlalchemy.py`, and it's a bit simpler than the previous one.

The file `invenio/modules/relationships/testsuite/test_nodes.py` only shows that this inheritance mechanism is working, while the main API is tested in `invenio/modules/relationships/testsuite/test_edges.py`.

I'm still missing a few things:

* Testing that `annotations` have the `Node` behaviour, and perhaps some more classes.
* Documenting, licensing and commenting the whole thing.

(CC: @kaplun @jalavik)

[skip ci]


invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,117 @@
...
+
+    model_from = db.Column(db.String(255), nullable=True, index=True)
+    id_from = db.Column(db.Integer(unsigned=True), nullable=True, index=True)
+
+    link_type = db.Column(db.String(255), index=True)
+    attributes = db.Column(db.JSON)
+
+    model_to = db.Column(db.String(255), nullable=True, index=True)
+    id_to = db.Column(db.Integer(unsigned=True), nullable=True, index=True)
+
```
jalavik: So I'm not on my computer now but usually there are issues with defining the constructor in db.Model objects. Due to SQLAlchemy "magic" (I can't recall what exactly). Did you see any issues with this?

jacquerie: No, no issues with that. Will investigate.

I've read [this](http://alextechrants.blogspot.ch/2013/11/10-common-stumbling-blocks-for.html) on this topic, point 2. I read that as "usually they aren't needed, but newbies defined them anyway". I thought that in this case is OK since there's a little logic in them.

invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,117 @@
...
+    """TODO"""
+
+    def _edges(self, inwards=True, outwards=True, link_type=None):
+        """TODO"""
+        query = SQLAlchemyEdge.query
+
+        if inwards and outwards:
+            query = query.filter((SQLAlchemyEdge.id_to == self['_id']) |
+                                 (SQLAlchemyEdge.id_from == self['_id']))
```
kaplun: In all these filters, I think you should also filter by accordingly for the `model_from` and `model_to`, coming accordingly from `self.__storagename__`.

invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,117 @@
...
+        self.id_to = str(json_to['_id'])
+
+    def get_uuid(self):
+        return self.uuid
+
+    def from_node(self):
+        return (self.model_from, self.id_from)
+
+    def to_node(self):
+        return (self.model_to, self.id_to)
```
MSusik: Running `from_node` or `to_node` methods I would expect to get a reocrd class instance, not a tuple. If there is no nice way to do it, you can add some map `model_name -> class`. Otherwise every user of this module will be forced to do translation on his own.

Other option would be storing the class path (like `modules.records.api:Record`) in the table in fields `model_to` and `id_to`, and building an object from it automatically

invenio/modules/relationships/testsuite/test_edges.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_predecessors(self):
+        predecessors = self._target.predecessors()
+
+        self.assertEqual(len(predecessors), 1)
+
+    def test_successors(self):
+        successors = self._source.successors()
+
+        self.assertEqual(len(successors), 1)
```
MSusik: Tests for a reflexive relation might be useful.

invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,117 @@
...
+        """TODO"""
+        in_edges = self.in_edges(link_type=link_type)
+
+        return [edge.from_node() for edge in in_edges]
+
+    def successors(self, link_type=None):
+        """TODO"""
+        out_edges = self.out_edges(link_type=link_type)
+
+        return [edge.to_node() for edge in out_edges]
```
MSusik: This way if there are two relations in the same direction between two nodes, we will receive the neighbour two times in the list. Why a `list` is prefered over a `set`?

invenio/modules/relationships/api.py
```diff
@@ -0,0 +1,6 @@
+from werkzeug.utils import import_string
+
+from .config import CFG_RELATIONSHIPS_NODE_ENGINE, CFG_RELATIONSHIPS_EDGE_ENGINE
+
+Node = import_string(CFG_RELATIONSHIPS_NODE_ENGINE)
```
MSusik: I think you should use `app.config`. In the usual case, the config file is not overwritten, there is additional config file outside of the source of `Invenio`.

invenio/modules/relationships/engines/sqlalchemy.py
```diff
@@ -0,0 +1,117 @@
+from uuid import uuid4
+
+from sqlalchemy_utils import UUIDType
+
+from invenio.ext.sqlalchemy import db, utils
+
+
+class SQLAlchemyNode(object):
```
MSusik: As every implementation of `Node` and every implementation of `Relationship` will have the same interface, it might be worth to create "abstract" superclasses for them.
Such solution is implemented for example for `StorageEngines` in `JSONAlchemy`:

https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/jsonalchemy/jsonext/engines/sqlalchemy.py
https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/jsonalchemy/storage.py

After doing so, a new developer who will work on another engines, will __surely__ implement API methods. If they are well documented, his methods will give expected output.

invenio/modules/relationships/engines/sqlalchemy.py
jacquerie: Yes, I should consider that. On the other hand, filtering by the appropriate `link_type` should yield the same result, since a specific `link_type` uniquely determines the `model_to` and `model_from` (or at least it does in the "relationships octopus" diagram we drew together).

invenio/modules/relationships/engines/sqlalchemy.py
jacquerie: Good point. The problem with directly returning a `db.Model` is that this requires a query, therefore the `successors` and `predecessors` method will run 1 query to retrieve the edges and then N, one for each result: the dreaded N+1 query problem (http://stackoverflow.com/questions/97197/what-is-the-n1-selects-issue).

Instead, returning a tuple forces the user of this class to unpack the ids and, hopefully, run a single `IN` query with them. I'm open for suggestions of a better design!

invenio/modules/relationships/testsuite/test_edges.py
jacquerie: I agree. I had those in a previous version, then removed them. Will add them back.

invenio/modules/relationships/engines/sqlalchemy.py
jacquerie: A `set` loses some information compared to a `list`; since users of this methods will have to do some work of their own (as I said in a previous comment), returning the most complete information will help them.

Besides, we (me and @kaplun) weren't able to come up with a real world example of a double link. Do you have one?

invenio/modules/relationships/api.py
jacquerie: Ah, yes, definitely. Will correct.

invenio/modules/relationships/engines/sqlalchemy.py
jacquerie: I'd like to propose an alternative design: instead of relying on abstract classes, rely on the tests. The `invenio/modules/relationships/testsuite/test_nodes.py` should test that the `Node` class has all the methods a `Node` should have (the test is currently testing for the presence of `_edges`, which is wrong, since it is an implementation detail).

Any new implementation will just have to run this test to know that it's doing the right thing. Perhaps the test should be renamed to `test_node_interface.py` to stress this.

invenio/modules/relationships/engines/sqlalchemy.py
kaplun: Well, the `link_type` happen to be unique, but is not necessary the case: imagine a link type: "`childrenof`". This is valid no matter of the categories involved.

Additionally it is optional in the signature of your function ;-)

invenio/modules/relationships/engines/sqlalchemy.py
MSusik: Running `IN` query can be done inside the method.

If you are worried about multiple `successors`/`predecessors` calls, you can also write a wrapper function which will use the `IN` query to return `Nodes` when tuples are given.

tiborsimko:
High-level naming comment: the module should rather be named `relations` (among objects), not `relationships` (among people).

```
From WordNet (r) 3.0 (2006) [wn]:
  relationship
      n 1: a relation between people; (`relationship' is often used
           where `relation' would serve, as in `the relationship
           between inflation and unemployment', but the preferred
           usage of `relationship' is for human relations or states of
           relatedness); "the relationship between mothers and their
           children" [syn: {relationship}, {human relationship}]
      2: a state of connectedness between people (especially an
         emotional connection); "he didn't want his wife to know of
         the relationship"
      3: a state involving mutual dealings between people or parties
         or countries
      4: (anthropology) relatedness or connection by blood or marriage
         or adoption [syn: {kinship}, {family relationship},
         {relationship}]
```

invenio/modules/relationships/engines/sqlalchemy.py
MSusik: > real world example

An author who has been working two times in a university. For example http://inspirehep.net/author/profile/J.R.Ellis.1. The point is that if the developer wants to get information about links, he uses edges method.

Anyway, returning an ordered collection is fine for me, I just wanted to make things clear. And a second question: why a list instead of a tuple?

invenio/modules/relationships/engines/sqlalchemy.py
MSusik: Does this suggestion include a base class for tests? Will it be obvious for a developer that he should do it in this way?

MSusik:
@tiborsimko 

>High-level naming comment: the module should rather be named `relations` (among objects), not `relationships` (among people).

My mistake, sorry for that.

invenio/modules/relationships/engines/sqlalchemy.py
MSusik: One more thing to add here. If we have a JSON field inside the table, we can actually get rid of any multigraph thinking, just by storing information in this field in case of doubled edges.

However, it is a design decision, and should be discussed with interested people.

jacquerie:
This approach was rejected in favor of a solution similar to https://github.com/inveniosoftware/invenio/compare/pu...jirikuncar:2573-relationships, except that it uses class decorators to work around the metaclass conflict problem I described in a previous comment.

Unfortunately, there's another problem with that approach: the proposed black magic and the one in JSONAlchemy conflict. See this branch: https://github.com/jacquerie/invenio/commit/f11ff09afa6d70aac9f7bdd583bd8eecf35f6021. The problem is that `Document.create({})` doesn't create an object of `__class__` Document, as it would be expected, but an object of class `invenio.modules.jsonalchemy.jsonext.parsers.extension_model_parser.Document`. Therefore its class doesn't get found in the class registry, and the registry becomes useless.

tiborsimko:
> My mistake, sorry for that.

No problem, but can we rename the module to "relations", and also all the PR files etc, before "relationships" takes a chance to soak in irreversibly?

jacquerie:
@tiborsimko Sure! The only thing that might still have the name "relationships" in it is the branch name, so that I can reuse this PR.

It's just that the technical/architectural issues I'm facing prioritized my attention.

tiborsimko:
@jacquerie Perfect, thanks.

jacquerie:
@jirikuncar The problem mentioned in https://github.com/inveniosoftware/invenio/pull/2903#issuecomment-86982597 is a blocker for me. Can you take a look?

kaplun:
We have to close this PR as (will be) superseded by different approach based on invenio-records and pure jsonschema.

PR #2906, opened 2015-03-20
access: usedeposit action addition

### LINKED ISSUE #2724: RFC deposit: filter depositions based on user roles
Problem
=======

Currently for all visitors of the index page of deposit ```/deposit``` all depositions are shown.

We would like to have some filtering done based on the roles of the current user so that different users can see different submissions.

Proposal
===========
https://github.com/inveniosoftware/invenio/blob/pu/invenio/modules/deposit/views/deposit.py#L113
```python
ctx = dict(
        deposition_types=DepositionType.all(),
        my_depositions=Deposition.get_depositions(current_user),
        prefill_data=draft_cache.data,
    )
```

Instead of ```DepositionType.all()``` we can have a new function ```DepositionType.all_authorized(current_user)``` that only returns deposition types the user is allowed to see.

We will need a new attribute in the ``DepositionType`` model that represents the roles that are authorized to see such Deposition type (we probably don't need a config variable for this, if a new role needs to be added to a deposition type, it can be changed in the DepositionType definition)

Then for all views that access a certain deposition type, like:

```python
@blueprint.route('/<depositions:deposition_type>')
@blueprint.route('/<depositions:deposition_type>/create',
                 methods=['POST', 'GET'])
@blueprint.route('/<depositions:deposition_type>/<int:uuid>/<draft_id>',
                 methods=['POST'])
```

There should be a protection so that, something like a 404 is returned if the user is not authorized. Should this be implemented as a decorator for those views?

What more should be taken into account? Suggestions for the implementation?

CC @lnielsen

jstypka:
* NEW Adds usedeposit action which enables per user access
  restrictions for different deposit types. (closes #2724)

* Requires running `webaccessadmin -u admin -c -a -D` command.

Signed-off-by: Jan Stypka <jan.stypka@cern.ch>

cc @jmartinm @jalavik

### DIFF
--- a/invenio/modules/access/local_config.py
+++ b/invenio/modules/access/local_config.py
@@ -133,6 +133,7 @@ DEF_ROLES = ((SUPERADMINROLE, 'superuser with all rights', 'deny any'),
              ('paperattributionlinkviewers', 'Users who can see attribution links in the search', 'allow all'),
              ('authorlistusers', 'Users who can user Authorlist tools', 'deny all'),
              ('holdingpenusers', 'Users who can view Holding Pen', 'deny all'),
+             ('depositusers', 'Users who can use a deposit type', 'allow any'),
              )
 
 
@@ -240,6 +241,7 @@ DEF_AUTHS = (('basketusers', 'usebaskets', {}),
              ('claimpaperoperators', 'claimpaper_change_own_data', {}),
              ('claimpaperoperators', 'claimpaper_change_others_data', {}),
              ('holdingpenusers', 'viewholdingpen', {}),
+             ('depositusers', 'usedeposit', {}),
              )
 
 
--- a/invenio/modules/deposit/cache.py
+++ b/invenio/modules/deposit/cache.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Implementation of allowed deposit type caching."""
+
+from .acl import usedeposit
+from .registry import deposit_types
+
+
+def get_authorized_deposition_types(user_info):
+    """Return a list of allowed deposition types for a certain user."""
+    from invenio.modules.access.engine import acc_authorize_action
+
+    return {
+        key for key in deposit_types.mapping().keys()
+        if acc_authorize_action(user_info, usedeposit.name, type=key)[0] == 0
+    }
--- a/invenio/modules/deposit/models.py
+++ b/invenio/modules/deposit/models.py
@@ -23,87 +23,114 @@ Classes for wrapping BibWorkflowObject and friends to make it easier to
 work with the data attributes.
 """
 
-from uuid import uuid4
 import json
 import os
 from datetime import datetime
+
+from uuid import uuid4
+
 from dateutil.tz import tzutc
 
-from sqlalchemy.orm.exc import NoResultFound
-from werkzeug.datastructures import MultiDict
-from werkzeug.utils import secure_filename
-from flask import redirect, render_template, flash, url_for, request, \
-    session, current_app
+from flask import current_app, flash, redirect, render_template, request, \
+    session, url_for
 from flask_login import current_user
 from flask_restful import fields, marshal
-from invenio.ext.restful import UTCISODateTime
+
 from invenio.base.helpers import unicodifier
+from invenio.ext.restful import UTCISODateTime
 
 from invenio.ext.sqlalchemy import db
-from invenio.modules.workflows.models import BibWorkflowObject, Workflow, \
-    ObjectVersion
 from invenio.modules.workflows.engine import WorkflowStatus
+from invenio.modules.workflows.models import BibWorkflowObject, ObjectVersion, \
+    Workflow
+
+from sqlalchemy.orm.exc import NoResultFound
+
+from werkzeug.datastructures import MultiDict
+from werkzeug.utils import secure_filename
 
 from .form import CFG_FIELD_FLAGS, DataExporter
 from .signals import file_uploaded
-from .storage import Storage, DepositionStorage
+from .storage import DepositionStorage, Storage
 
 
 #
 # Exceptions
 #
 class DepositionError(Exception):
+
     """Base class for deposition errors."""
+
     pass
 
 
 class InvalidDepositionType(DepositionError):
+
     """Raise when a deposition type cannot be found."""
+
     pass
 
 
 class InvalidDepositionAction(DepositionError):
+
     """Raise when deposition is in an invalid state for action."""
+
     pass
 
 
 class DepositionDoesNotExists(DepositionError):
+
     """Raise when a deposition does not exists."""
+
     pass
 
 
 class DraftDoesNotExists(DepositionError):
+
     """Raise when a draft does not exists."""
+
     pass
 
 
 class FormDoesNotExists(DepositionError):
+
     """Raise when a draft does not exists."""
+
     pass
 
 
 class FileDoesNotExists(DepositionError):
+
     """Raise when a draft does not exists."""
+
     pass
 
 
 class DepositionNotDeletable(DepositionError):
+
     """Raise when a deposition cannot be deleted."""
+
     pass
 
 
 class FilenameAlreadyExists(DepositionError):
+
     """Raise when an identical filename is already present in a deposition."""
+
     pass
 
 
 class ForbiddenAction(DepositionError):
+
     """Raise when action on a deposition, draft or file is not authorized."""
+
     pass
 
 
 class InvalidApiAction(DepositionError):
+
     """Raise when an invalid API action is requested."""
+
     pass
 
 
@@ -111,7 +138,9 @@ class InvalidApiAction(DepositionError):
 # Helpers
 #
 class FactoryMixin(object):
+
     """Mix-in class to help create objects from persisted object state."""
+
     @classmethod
     def factory(cls, state, *args, **kwargs):
         obj = cls(*args, **kwargs)
@@ -123,6 +152,7 @@ class FactoryMixin(object):
 # Primary classes
 #
 class DepositionType(object):
+
     """
     A base class for the deposition types to ensure certain
     properties are defined on each type.
@@ -133,6 +163,7 @@ class DepositionType(object):
     you can override the render_error(), render_step() and render_completed()
     methods.
     """
+
     workflow = []
     """ Workflow definition """
 
@@ -521,8 +552,22 @@ class DepositionType(object):
         """ Return a name for this class """
         return self.get_identifier()
 
+    @classmethod
+    def all_authorized(cls, user_info=None):
+        """
+        Return a dict of deposit types that the current user
+        is allowed to view.
+        """
+        user_info = user_info or current_user
+
+        all_types = cls.all()
+        auth_types = user_info.get('precached_allowed_deposition_types', set())
+
+        return {key: all_types[key] for key in auth_types if key in all_types}
+
 
 class DepositionFile(FactoryMixin):
+
     """
     Represents an uploaded file
 
@@ -565,6 +610,7 @@ class DepositionFile(FactoryMixin):
         d.delete()
 
     """
+    
     def __init__(self, uuid=None, backend=None):
         self.uuid = uuid or str(uuid4())
         self._backend = backend
--- a/invenio/modules/deposit/views/deposit.py
+++ b/invenio/modules/deposit/views/deposit.py
@@ -20,34 +20,40 @@
 """Deposit Blueprint implementation."""
 
 import json
-from functools import wraps
 import warnings
+from functools import wraps
 
-from flask import current_app, Blueprint, \
+from flask import Blueprint, abort, current_app, \
+    flash, \
+    jsonify, \
+    make_response, \
+    redirect, \
     render_template, \
     request, \
-    jsonify, \
-    redirect, \
-    url_for, \
-    flash, \
     send_file, \
-    abort, \
-    make_response
-from werkzeug.datastructures import MultiDict
-from werkzeug.utils import secure_filename
-from flask_login import current_user, login_required
+    url_for
 from flask_breadcrumbs import default_breadcrumb_root, register_breadcrumb
+
+from flask_login import current_user, login_required
+
 from flask_menu import register_menu
 
 from invenio.base.i18n import _
+from invenio.ext.principal import permission_required
+
+from werkzeug.datastructures import MultiDict
+from werkzeug.utils import secure_filename
+
+from ..acl import usedeposit
+from ..models import Deposition, DepositionDoesNotExists, \
+    DepositionDraftCacheManager, DepositionFile, DepositionNotDeletable, \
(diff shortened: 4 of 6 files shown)


invenio/modules/access/local_config.py
```diff
@@ -558,7 +558,8 @@
                ('claimpaper_change_others_data', 'Change data of any person ID', '', 'no'),
                ('runbibtasklet', 'run BibTaskLet', '', 'no'),
                ('cfgbibsched', 'configure BibSched', '', 'no'),
-               ('runinfomanager', 'run Info Space Manager', '', 'no')
+               ('runinfomanager', 'run Info Space Manager', '', 'no'),
+               ('usedeposit', 'use deposit type', 'type', 'yes'),
```
jirikuncar: this line is not needed if you define your own action in https://github.com/inveniosoftware/invenio/pull/2906/files#diff-1917d7fa51639dedef09e17079a8a7cdR25

invenio/modules/deposit/views/deposit.py
```diff
@@ -129,6 +131,10 @@ def index():
 @register_breadcrumb(blueprint, '.type', _('Select or create'))
 def deposition_type_index(deposition_type):
     """Render deposition type index."""
+
+    if not permission_required(usedeposit.name, type=deposition_type).can():
```
jirikuncar: Please use it as decorator:

```python
def deposition_type_from_request():
    """Return deposition from parsed view arguments."""
    return request.view_args['deposition_type']


@ permission_required(usedeposit, type=deposition_type_from_request)
def <action>(deposition_type=None):
    pass
```

invenio/modules/deposit/views/deposit.py
```diff
@@ -236,6 +246,10 @@ def save(deposition_type=None, uuid=None, draft_id=None):
 
     deposition = Deposition.get(uuid, current_user, type=deposition_type)
 
+    dep_type = deposition_type if deposition_type else Deposition.get_type(uuid)
```
jirikuncar:
```python
def deposition_type_from_request():
    """Return deposition from parsed view arguments."""
    return request.view_args.get(
        'deposition_type', None
    ) or (
        Deposition.get_type(request.view_args.get('uuid')) 
        if request.view_args.get('uuid', None) else None
    )
```

jirikuncar:
@jstypka can I ask you to first improve PEP8/PEP257 code style of the files you are going to edit? Thank you in advance

jstypka:
@jirikuncar I've improved style of my code and the files in general. Hope that helps.

jirikuncar:
@jstypka I think you might want to create new role and authorization similar to:
- https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/local_config.py#L472
- https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/local_config.py#L576

jstypka:
I can add it, but shouldn't the roles be specific for a particular Invenio instance?
E.g. 
`('articledeposituser', 'usedeposit', {'type' : 'article'}),`
would imply that there should be an `article` deposition type, which is not always the case.

jirikuncar:
Just use this: `('deposituser', 'usedeposit', {}),` for Invenio and you can modify it in INSPIRE overlay similarly to https://github.com/inveniosoftware/invenio-demosite/blob/master/invenio_demosite/modules/access/config.py.

jstypka:
@jirikuncar There you go!

jirikuncar:
@jstypka it's shaping well. There is one last not addressed point in https://github.com/inveniosoftware/invenio/issues/2724#issuecomment-72184717.

invenio/modules/deposit/models.py
```diff
@@ -521,8 +553,29 @@ def __unicode__(self):
         """ Return a name for this class """
         return self.get_identifier()
 
+    @classmethod
+    def all_authorized(cls, current_user):
```
jirikuncar:
```python
def all_authorized(cls, user_info=None):
    user_info = user_info or current_user
    ...
```

invenio/modules/deposit/models.py
```diff
@@ -521,8 +553,29 @@ def __unicode__(self):
...
+        all_types = cls.all()
+        action_id = acc_get_action_id('usedeposit')
+
+        avail_actions = acc_find_possible_actions_user(current_user.get_id(),
+                                                       action_id)[1:]
+        values = [s for s in itertools.chain(*avail_actions)
+                  if isinstance(s, basestring)]
+
+        # filter the all_types dict()
+        return {key: all_types[key] for key in values if key in all_types}
```
jirikuncar: What if by default all types are allowed? Maybe we should have something like:
```python
return {key: all_types[key] for key in all_types.keys() if acc_authorize_action(usedeposit, type=key)}
```

jirikuncar: @jstypka how does it look like? Do you have any troubles with it?

jstypka: @jirikuncar Sorry, but I didn't have time recently to push it forward. I will probably get back to it tomorrow or the day after and push some changes.

jstypka: @jirikuncar I changed the user_info line that you suggested. Regarding:

`return {key: all_types[key] for key in all_types.keys() if acc_authorize_action(usedeposit, type=key)}`

I think this is semantically the same as code above, but little cleaner and more readable. However it would require much more database queries, which might hurt the performance. If you think it's better I may change it.

Can you also explain me how do you calculate the `precached_allowed_deposition_types`? I didn't manage to find anything similar in the code. Is it a user property? When should it be calculated?

jirikuncar: > Can you also explain me how do you calculate the `precached_allowed_deposition_types`?

Please check: `invenio.ext.login.legacy_user`

> I think this is semantically the same as code above

You can check if one of the authorized action has `type=None` then **all** depositions should be allowed.

```python
avail_actions = acc_find_possible_actions_user(
    current_user.get_id(),
    action_id)[1:]
values = set(s for s in itertools.chain(*avail_actions) if isinstance(s, basestring))
# check 
if None in values:
    # return all actions
return # filter allowed actions
```

jstypka: Ah, I think I understand. So you want to have some kind of a wildcard - if a user is allowed to (usedeposit, type=None), this would be equivalent to a right to use all the types?

I can do that, but I think it would require additional `or s is None` in the if condition in the set comprehension.

jirikuncar: @jstypka yes, I have removed it from there

jstypka:
@jirikuncar I finally decided to go with your initial suggestion for the dict creation. The problem is that the `acc_find_possible_actions_user` function returns an empty list for the SUPERADMIN role, which would complicate the situation.

I also made an attempt to implement the cache - please have a look at it, because I'm not sure about some stuff.

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,70 @@
...
+        user_info,
+        'usedeposit',
+        batch_args=True,
+        collection=deposit_type_cacher.cache
+    )
+
+    for collection, auth in zip(deposit_type_cacher.cache, auths):
+        if auth[0] == 0:
+            ret.append(collection)
+    return ret
```
jirikuncar: Can you keep the new line character here(`\n`) (not new empty line `\n\n` or none `\0`). Thanks

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,70 @@
...
+
+    if recreate_cache_if_needed:
+        deposit_type_cacher.recreate_cache_if_needed()
+    ret = []
+
+    auths = acc_authorize_action(
+        user_info,
+        'usedeposit',
+        batch_args=True,
+        collection=deposit_type_cacher.cache
```
jirikuncar: `type=`?

invenio/modules/deposit/models.py
```diff
@@ -521,8 +552,25 @@ def __unicode__(self):
...
+        """
+        Return a dict of deposit types that the current user
+        is allowed to view.
+        """
+        user_info = user_info or current_user
+        from invenio.modules.access.engine import acc_authorize_action
+
+        all_types = cls.all()
+
+        return {key: all_types[key] for key in all_types.keys()
```
jirikuncar: are you going to use `get_authorized_deposition_types` here?

invenio/modules/deposit/cache.py
jstypka: good point, pushed the changes

jstypka:
@jirikuncar good catches - fixed!

jstypka:
@jirikuncar so how does it look? Any other remarks?

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,71 @@
...
+class DepositTypeDataCacher(DataCacher):
+    def __init__(self):
+        def cache_filler():
+            from invenio.modules.access.control import acc_get_action_id
+            from invenio.modules.access.models import AccAuthorization, \
+                AccARGUMENT
+
+            action_id = acc_get_action_id('usedeposit')
+
+            return [auth[0] for auth in AccAuthorization.query.join(
```
jirikuncar: shall we return `set` instead of `list`?

invenio/modules/deposit/models.py
```diff
@@ -521,8 +552,23 @@ def __unicode__(self):
...
+    def all_authorized(cls, user_info=None):
+        """
+        Return a dict of deposit types that the current user
+        is allowed to view.
+        """
+        user_info = user_info or current_user
+        from .cache import get_authorized_deposition_types
+
+        all_types = cls.all()
+        auth_types = get_authorized_deposition_types(user_info)
```
jirikuncar: shall we use `user_info.get('precached_...', set())`?

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,71 @@
...
+    ret = []
+
+    auths = acc_authorize_action(
+        user_info,
+        'usedeposit',
+        batch_args=True,
+        type=deposit_type_cacher.cache
+    )
+
+    for dep_type, auth in zip(deposit_type_cacher.cache, auths):
```
jirikuncar:
```python
return {
    dep_type for dep_type, auth in zip(deposit_type_cacher.cache, auths)
    if auth[0] == 0
}
```

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,71 @@
...
+            from invenio.modules.access.models import AccAuthorization, \
+                AccARGUMENT
+
+            action_id = acc_get_action_id('usedeposit')
+
+            return [auth[0] for auth in AccAuthorization.query.join(
+                AccAuthorization.argument
+            ).filter(
+                AccARGUMENT.keyword == 'type',
+                AccAuthorization.id_accACTION == action_id
```
jirikuncar: you can do one more 'join' or 'subquery' to get `action_id`.

jirikuncar:
Can you also mention in commit message that one need to run: `webaccessadmin -u admin -c -a -D`.

--
cc @kaplun can you confirm that the command arguments work well? Thanks

jstypka:
@jirikuncar done, is it good to go?

jirikuncar:
@jstypka have you talked with @kaplun about the `webaccessadmin`? I'm not sure about the arguments. Otherwise it looks ok.

kaplun:
@jstypka, @jirikuncar, indeed it's correct. We can skip the `-D` since this is for the demosite and there are no demo authorization added in @jstypka's PR.

jirikuncar:
@kaplun thanks! Please consider adding yourself there as reviewer. @jstypka please rebase the PR.

invenio/modules/deposit/cache.py
jstypka: I think it might be causing an error. I get a following type error:
```
File "/home/stpk/envs/invenio/src/invenio/invenio/ext/login/legacy_user.py", line 84, in __init__
  acc = self._precache(data, force)
File "/home/stpk/envs/invenio/src/invenio/invenio/ext/login/legacy_user.py", line 203, in _precache
  get_authorized_deposition_types(user_info)
File "/home/stpk/envs/invenio/src/invenio/invenio/modules/deposit/cache.py", line 60, in get_authorized_deposition_types
  type=deposit_type_cacher.cache
File "/home/stpk/envs/invenio/src/invenio/invenio/modules/access/engine.py", line 57, in acc_authorize_action
  roles_list = acc_find_possible_roles(name_action, always_add_superadmin=True, batch_args=batch_args, **arguments)
File "/home/stpk/envs/invenio/src/invenio/invenio/modules/access/control.py", line 1451, in acc_find_possible_roles
  elem[k] = arguments[k][i]
TypeError: 'set' object does not support indexing
```

jirikuncar:
ping @jstypka (https://github.com/inveniosoftware/invenio/pull/2906#issuecomment-90493478)

jstypka:
@jirikuncar after resolving rebase conflicts I'm testing the application again. Please have a look at:
https://github.com/inveniosoftware/invenio/pull/2906#discussion_r27883399
I think there might be a bug in the `batch_args` usage.

invenio/modules/deposit/cache.py
jirikuncar: I might need to talk with @lnielsen about https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/control.py#L1444-L1452 (from a9ca60d2). I'm not sure if I completely understand the code there.

jirikuncar: @jstypka quick fix is to pass `list` instead of `set`.

```diff
- type=deposit_type_cacher.cache
+ type=type=list(deposit_type_cacher.cache)
```

jstypka: @jirikuncar yes, I think this fix or basically returning a list instead of a set in `def cache_filler()` should solve it. Everything seems to work, apart from the fact that all users get authorized, because of this function:

https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/firerole.py#L194-L273

It authorizes everyone by default, but I understand it's configurable and it's not a problem with my code, am I right?

jirikuncar: By default all *authenticated* users should be allowed to use deposit.

invenio/modules/deposit/cache.py
```diff
@@ -0,0 +1,67 @@
...
+    from invenio.modules.access.engine import acc_authorize_action
+
+    if recreate_cache_if_needed:
+        deposit_type_cacher.recreate_cache_if_needed()
+
+    auths = acc_authorize_action(
+        user_info,
+        'usedeposit',
+        batch_args=True,
+        type=deposit_type_cacher.cache
```
jirikuncar: `type=list(deposit_type_cacher.cache)`

jirikuncar:
@jstypka please note that there is a failing test
```
======================================================================
FAIL: test_registration (invenio.modules.deposit.testsuite.test_type_simplerecord.SimpleRecordTest)
  File "/home/travis/build/inveniosoftware/invenio/invenio/modules/deposit/testsuite/test_type_simplerecord.py", line 102, in test_registration
    assert "Tests" in res.data
AssertionError

```

jirikuncar:
@jstypka you have also left out `(addresses #2724)` or `(closes #2724)` in commit message.

jstypka:
@jirikuncar I think I pushed all the changes that we discussed. Is it fine?

jirikuncar:
@jstypka the tests are still failing :disappointed:

jirikuncar:
@jstypka do you need any help with the failing test? Please let us know if you are working on it. Thanks

jstypka:
@jirikuncar Yes, I am currently trying to figure out why it fails. If I get stuck I'll let you know.

jirikuncar:
@jstypka I offered my help so please tell me where is the problem or deliver working solution. Thank you!

jstypka:
@jirikuncar I sent you an e-mail few days ago. Basically the cache at the beginning seems to be filled properly, however the UserInfo object that is checked during the test seems to be a blank one - all lists empty, all fields False etc.

jirikuncar:
@jstypka please use my updated commit from https://github.com/jirikuncar/invenio/tree/2906-fix. Moreover we need to protect also deposit REST API - most importantly https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/deposit/restful.py#L250.

jirikuncar:
Is anybody looking into this? (cc @kaplun @jmartinm @jalavik)

jirikuncar:
I have fixed the issue with tests (see https://github.com/jirikuncar/invenio/tree/2906-fix) caused by caches in user info object.

@jstypka add permission checking to REST API.

--
cc @kaplun @jmartinm @jalavik @tiborsimko

jirikuncar:
Superseded by #3172

PR #2924, opened 2015-03-24
multimedia: replace with Flask-IIIF extension

drjova:
First should be merged https://github.com/inveniosoftware/flask-iiif/pull/3


invenio/base/config.py
```diff
@@ -55,6 +55,7 @@
     'invenio.ext.jasmine',  # after assets
     'flask_breadcrumbs:Breadcrumbs',
     'invenio.modules.deposit.url_converters',
+    "invenio.ext.multimedia"
```
jirikuncar: Can you rename it to `iiif` to follow *Flask-IIIF* name?

requirements.txt
```diff
@@ -1,2 +1,3 @@
 # requires numpy before it can be installed
 #-e svn://svn.code.sf.net/p/gnuplot-py/code/trunk#egg=gnuplot-py
+git+https://github.com/drjova/flask-iiif.git@personal-fixes # :( :( :(
```
jirikuncar: It should not be here. Instead add it to https://github.com/inveniosoftware/invenio/blob/master/.travis-devel-requirements.txt.

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+"""
+    invenio.ext.iiif
```
jirikuncar: Please use new style of comments compatible with PEP257.

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+from ..cache import cache
+from invenio.modules.documents.api import Document
+from invenio.modules.documents.errors import DocumentNotFound
+
+iiif = IIIF()
+
+__all__ = ['invenio_uuid_to_path', 'setup_app']
+
+
+def invenio_uuid_to_path(uuid):
```
jirikuncar: `s/invenio_uuid_to_path/uuid_to_path/`

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    ----------------
+
+    This module provides initialization and configuration for `flask_iiif`
+    module.
+"""
+from flask_iiif import IIIF
+from flask_iiif.errors import MultimediaImageNotFound
+
+from ..cache import cache
+from invenio.modules.documents.api import Document
```
jirikuncar: Global `invenio` imports might cause troubles. I would move them inside `uuid_to_path` function that should be probably defined inside `invenio.modules.documents.utils/wrappers`.

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+    else:
+        return document.get('uri', '')
+
+
+def setup_app(app):
+    """Setup iiif extension."""
+    iiif.init_app(app)
+    iiif.init_restful(app.extensions['restful'])
+    iiif.uuid_to_path_handler(invenio_uuid_to_path)
```
jirikuncar: We shall check that `invenio.modules.documents` is in `PACKAGES` (`assert`?).

jirikuncar:
inveniosoftware/flask-iiif#3 has been merged. :shipit: 

--
ping @drjova

.travis-devel-requirements.txt
```diff
@@ -15,3 +15,5 @@
 -e git+git://github.com/inveniosoftware/flask-breadcrumbs.git#egg=flask-breadcrumbs
 -e git+git://github.com/inveniosoftware/flask-registry.git#egg=flask-registry
 -e git+git://github.com/inveniosoftware/invenio-query-parser.git#egg=invenio-query-parser
+
+-e git+https://github.com/drjova/flask-iiif.git@personal-fixes#egg=flask-iiif
```
jirikuncar: don't forget to change it to official repo

invenio/base/config.py
```diff
@@ -55,6 +55,7 @@
     'invenio.ext.jasmine',  # after assets
     'flask_breadcrumbs:Breadcrumbs',
     'invenio.modules.deposit.url_converters',
+    "invenio.ext.iiif"
```
jirikuncar: please add `,` at the end of line. thanks

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+
+def setup_app(app):
+    """Setup iiif extension."""
+    try:
+        assert 'invenio.modules.documents' not in \
+            app.config['PACKAGES_EXCLUDE']
+    except AssertionError:
+        raise MultimediaError(
+            "Could not initialize the Flask-IIIF extension because "
+            "`~invenio.modules.docuements` is missing"
```
jirikuncar: `s/docuements/documents/`

invenio/modules/documents/utils.py
```diff
@@ -17,4 +17,16 @@
...
+"""Document utils."""
+
+
+def uuid_to_path(uuid):
+    """Convert `~invenio.modules.documents.api.Document` to path."""
+    from invenio.modules.documents.api import Document
+
+    try:
+        document = Document.get_document(uuid)
+    except Exception:
```
jirikuncar: use more specific exception - e.g. `DocumentNotFound`

requirements-devel.txt
```diff
@@ -16,3 +16,5 @@
 -e git+git://github.com/inveniosoftware/flask-breadcrumbs.git#egg=flask-breadcrumbs
 -e git+git://github.com/inveniosoftware/flask-registry.git#egg=flask-registry
 -e git+git://github.com/inveniosoftware/invenio-query-parser.git#egg=invenio-query-parser
+
+-e git+git://github.com/inveniosoftware/flask-iiif.git#egg=flask-iiif
```
jirikuncar: please move it after `flask-registry`

invenio/modules/documents/utils.py
```diff
@@ -17,22 +17,19 @@
...
+        DocumentNotFound, DeletedDocument
     )
-    return _redis_server
+
+    try:
+        document = Document.get_document(uuid)
+    except (DocumentNotFound, DeletedDocument):
+        raise
+    else:
+        return document.get('uri', '')
```
jirikuncar: I'm not sure if the implementation can be so simple.

1. We need access rights that might depend on the record if the document is attached to any. (cc @lnielsen @tiborsimko)
2. The file can be on different storage e.g. S3, EOS, ... hence we should probably return something like `document.open('r')`. If we pass bytestream then we should amend Flask-IIIF so it can handle it too.

drjova: @jirikuncar about the 2nd point Flask-IIIF can handle bytestream.

jirikuncar: Then we might need to change the way how the REST API works in https://github.com/inveniosoftware/flask-iiif/blob/master/flask_iiif/restful.py#L103.

lnielsen: Re. 1: Yes, it will be a big security hole if we don't handle the access restrictions. But perhaps that can be solved via refactoring the ACL system.

jirikuncar:
Can you keep here only `invenio.modules.multimedia` removal without `Flask-IIIF` and `invenio.modules.documents` integration?

drjova:
done!

Issue #2750, opened 2015-02-06
inveniomanage demosite populate errors


hachreak:
Hi at all,
I tried to run `inveniomanage demosite populate --yes-i-know` and  I received these errors.
Someone can help me? :)

```python
--------------------------------------------------------------------------------
ERROR in wrappers [/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/ext/logging/wrappers.py:310]:

--------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibsched/bibtask.py", line 615, in task_init
    ret = _task_run(task_run_fnc)
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibsched/bibtask.py", line 1191, in _task_run
    if callable(task_run_fnc) and task_run_fnc():
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibindex/engine.py", line 2169, in task_run_core
    task_get_option("cmd") == "del"))
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibindex/engine.py", line 559, in find_affected_records_for_index
    AND (SELECT COUNT(*) FROM hstRECORD WHERE id_bibrec=id)=1""", (min_last_updated,))
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/dbquery_mysql.py", line 280, in run_sql
    rc = cur.execute(sql, param)
  File "/home/vagrant/.virtualenvs/invenio2/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
  File "/home/vagrant/.virtualenvs/invenio2/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
OperationalError: (1052, "Column 'id' in field list is ambiguous")
2015-02-06 13:20:12 -->
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibsched/bibtask.py", line 615, in task_init
    ret = _task_run(task_run_fnc)
  File "/home/vagrant/.virtualenvs/invenio2/src/invenio/invenio/legacy/bibsched/bibtask.py", line 1191, in _task_run
[...]

PR #2929, opened 2015-03-24
sorter: fix locale based comparison of lists

jirikuncar:
* FIX Comparison function of record tags uses space concatened string from list of all tags values.  (addresses #2750)

### DIFF
--- a/invenio/legacy/bibsort/engine.py
+++ b/invenio/legacy/bibsort/engine.py
@@ -1,7 +1,7 @@
 # -*- mode: python; coding: utf-8; -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2010, 2011, 2012 CERN.
+# Copyright (C) 2010, 2011, 2012, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


invenio/legacy/bibsort/engine.py
```diff
@@ -404,7 +403,12 @@ def sort_dict(dictionary, spacing=1, run_sorting_for_rnk=False, sorting_locale=N
                 locale.setlocale(locale.LC_ALL, sorting_locale + '.UTF8')
             except locale.Error:
                 write_message("Setting locale to %s is not working.. ignoring locale")
-        sorted_records_list = sorted(dictionary, key=dictionary.__getitem__, cmp=locale.strcoll, reverse=False)
+
+        def locale_cmp_list(first, second):
+            return locale.strcoll(' '.join(first), ' '.join(second))
```
jirikuncar: @ludmilamarian is there some other way how to do it? Shall we concatenate with some special `locale` word separator if exists?

jirikuncar: @tiborsimko @lnielsen can I do something like:
```python
dictionary = {key: ' '.join(value) for key, value in six.iteritems(dictionary)}
```
instead?

jirikuncar: maybe @kaplun knows more ...

jirikuncar:
@tiborsimko @kaplun can you please have a look?

invenio/legacy/bibsort/engine.py
tiborsimko: @ludmilamarian wrote that part, so she may be able to tell first hand.  Without studying the context too much, the join seems OK...

PR #2930, opened 2015-03-24
installation: WSGI user in Apache virtual hosts

tiborsimko:
* BETTER Apache virtual environments are now created with appropriate
  `WSGIDaemonProcess` user value, taken from the configuration variable
  `CFG_BIBSCHED_PROCESS_USER`, provided it is set.  This change makes it
  easier to run Invenio under non-Apache user identity.

### DIFF
--- a/modules/miscutil/lib/inveniocfg.py
+++ b/modules/miscutil/lib/inveniocfg.py
@@ -918,7 +918,7 @@ WSGIRestrictStdout Off
         AliasMatch /sitemap-(.*) %(webdir)s/sitemap-$1
         Alias /robots.txt %(webdir)s/robots.txt
         Alias /favicon.ico %(webdir)s/favicon.ico
-        WSGIDaemonProcess invenio processes=5 threads=1 display-name=%%{GROUP} inactivity-timeout=3600 maximum-requests=10000
+        WSGIDaemonProcess invenio processes=5 threads=1 display-name=%%{GROUP} inactivity-timeout=3600 maximum-requests=10000 %(wsgiuser)s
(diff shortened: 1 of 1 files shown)


modules/miscutil/lib/inveniocfg.py
```diff
@@ -918,7 +918,7 @@ def cli_cmd_create_apache_conf(conf):
         AliasMatch /sitemap-(.*) %(webdir)s/sitemap-$1
         Alias /robots.txt %(webdir)s/robots.txt
         Alias /favicon.ico %(webdir)s/favicon.ico
-        WSGIDaemonProcess invenio processes=5 threads=1 display-name=%%{GROUP} inactivity-timeout=3600 maximum-requests=10000
+        WSGIDaemonProcess invenio processes=5 threads=1 display-name=%%{GROUP} inactivity-timeout=3600 maximum-requests=10000 %(wsgiuser)s
```
dset0x: Hmm, this doesn't look right. There's `invenio`, but also `%(wsgiuser)s`. Perhaps the latter takes presedence, but the [wiki](https://code.google.com/p/modwsgi/wiki/ConfigurationDirectives#WSGIDaemonProcess) states that there is only one `name`. Note the old date of the document however.

jirikuncar: IMHO it's correct.
```
WSGIDaemonProcess invenio ... user=jkuncar
#                 ^-- process name ^-- user
```

dset0x: I didn't realize that `wsgiuser` was already prefixed with `user=`. Should be good in that case!

Issue #2862, opened 2015-03-09
global: jinja2hacks are causing troubles on Python 2.7.9


jirikuncar:
see for example https://travis-ci.org/inveniosoftware/invenio/builds/53365855

#### Workaround

Disable `invenio.ext.jinja2hacks` in `EXTENSIONS`.

--
cc @hachreak

lnielsen:
I have the same problem but doesn't seem to be solely related to ``jinja2hacks``:
https://travis-ci.org/zenodo/zenodo#L6092

Could be the encoding of one specific file that causes the problem as described on https://github.com/mitsuhiko/markupsafe/issues/19

hachreak:
```bash
file -bi invenio/base/templates/page_base.html
text/html; charset=us-ascii
```

tiborsimko:
`page_base.html` loads stuff from `_macros.html` that contains UTF-8 character "×" in the data dismiss alert.  

You can perhaps try to replace it with regular ASCII "x" too see whether it helps?

P.S. Note also that elsewhere we do stuff with this character, such as replacing it in widgets...

```console
$ git grep -Hni × invenio
invenio/base/templates/_macros.html:25:        <a class="close" data-dismiss="alert" href="#">×</a>
invenio/modules/accounts/static/js/accounts/widgets.js:108:                        title = widget.find("h4").text().replace("×", " ").trim()
invenio/modules/comments/templates/comments/add_base.html:60:      <a class="close" data-dismiss="modal">×</a>
invenio/modules/comments/templates/comments/add_review_base.html:53:      <a class="close" data-dismiss="modal">×</a>
invenio/modules/communities/templates/communities/new_base.html:32:        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">×</button>
invenio/modules/previewer/templates/previewer/bottom_base.html:27:    <button type="button" class="close" data-dismiss="modal" aria-hidden="true">×</button>
invenio/modules/search/static/js/search/facet.js:357:    close: $('<a/>', {'class': 'close', html:'&nbsp;×'}),
invenio/modules/workflows/templates/workflows/details_base.html:282:<a class="close" data-dismiss="alert">×</a>
[...]
```

jirikuncar:
```console
$ find invenio -name \*.html | xargs pcregrep "[\x80-\xFF]"
[...]

PR #2932, opened 2015-03-25
global: jinja2hacks deprecation and usage warning

jirikuncar:
* NOTE Adds deprecation warning for `invenio.ext.jinja2hacks` and all detected non-ascii strings usage in templates mainly coming from legacy (1.x) modules.  (addresses #2862)

Reviewed-by: Lars Holm Nielsen <lars.holm.nielsen@cern.ch>

### DIFF
--- a/invenio/ext/jinja2hacks.py
+++ b/invenio/ext/jinja2hacks.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # This file is part of Invenio.
-# Copyright (C) 2012, 2013 CERN.
+# Copyright (C) 2012, 2013, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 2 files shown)


invenio/ext/jinja2hacks.py
```diff
@@ -90,25 +95,25 @@ def setup_app(app):
...
 def utf8escape(s):
-    """
-    UTF8-8-bit-string-friendly replacement function for MarkupSafe/Jinja2
-    escape function.
+    """UTF8-8-bit-string-friendly replacement for MarkupSafe escape function.
 
     WARNING: Do not use this method. Use jinja2.escape() instead.
     """
     if isinstance(s, str):
+        warnings.warn("Convert string '{0}' in template to unicode.".format(s),
```
lnielsen: Perhaps we try to decode using ascii first, and if it fails, then issue the warning and decode using outfit. That give warnings only for the problematic strings.

lnielsen:
What about adding a template filter along the lines of:

```python
@app.template_filter('u')
def tounicode(val):
    if isinstance(val, str):
         return val.decode('utf8')
     return val
```

Then we can easily fix problems in the templates by:

```jinja
{{ myvar|u }}
```

lnielsen:
Looks good to me.

PR #2941, opened 2015-03-25
documents: fix Flask-Login import

jacquerie:
As discussed on Hangouts/Gitter, running `nosetests invenio/modules/documents/testsuite/test_api.py` is currently broken due to a broken import in `invenio/modules/documents/documentext/fields/base.cfg`, similar to https://github.com/inveniosoftware/invenio/commit/0875c0570b99f77af5a715ed35857c506b6e8bbf.


jacquerie:
Uh, I was under the impression that 1/2 spaces before the bullet point were mandatory. Sorry for the emails.

Issue #2945, opened 2015-03-26
RFC Removing remote debugger


jirikuncar:
Since `werkzeug` has nice interactive debugger, we can remove `invenio.utils.remote_debugger` from 2.1+.

--
PS: Ticket will be closed if there is no update for 5 days.

tiborsimko:
:+1: The remote debugger utility was born in the pre-Flask era and can be safely removed now.

PR #2946, opened 2015-03-26
global: deprecation of custom remote debugger

jirikuncar:
* NOTE Deprecates custom remote debuggers. Please use native Werkzeug
  debugger or other (*)pdb equivalents.  (addresses #2945)

### DIFF
--- a/invenio/utils/remote_debugger/__init__.py
+++ b/invenio/utils/remote_debugger/__init__.py
@@ -1,5 +1,5 @@
 # This file is part of Invenio.
-# Copyright (C) 2011, 2013 CERN.
+# Copyright (C) 2011, 2013, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


invenio/utils/remote_debugger/__init__.py
```diff
@@ -21,6 +21,13 @@
 that makes remote debugging possible and easy.
 """
 
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio21Warning
+
+warnings.warn("Remote debugger is going to be removed. "
+              "Please use native Werkzeug debugger.",
+              RemovedInInvenio21Warning)
```
lnielsen: Should in principle be ``RemovedInInvenio22Warning``, but I fully agree it's better to remove it faster. Should we amend the deprecation policy to allow for these sort of fast track deprecations.

jirikuncar: > Should we amend the deprecation policy to allow for these sort of fast track deprecations?

Yes! Especially for features that hasn't well tested and are not part of the *core* functionality.

jalavik: Agreed :+1:

lnielsen: I've opened an RFC #2955 for it :-)

Issue #2923, opened 2015-03-24
assets: enable less source maps


jmartinm:
Following up on https://github.com/inveniosoftware/invenio/issues/2921 I have come to the conclusion that our extension code could be simplified by dropping support for ``LESS_RUN_IN_DEBUG``.

This would avoid, for instance, the following 'hack' that needed to be introduced https://github.com/inveniosoftware/invenio/pull/2922/files#diff-dc717d85d0b109c334416b4b5a305f8cR136

What this variable does is enabling compilation of less files in the browser http://lesscss.org/#client-side-usage

But this does not offer any advantage as far as I can see.

By default ``webassets`` will do the right thing and if a ``less`` file on a Bundle is modified, upon page refresh, the bundle is recompiled and changes are visible immediately. If there are no changes then no compilation happens.

This is a WIP branch where I show what the change would be https://github.com/jmartinm/invenio/commit/5fdd9c0a540d2f8947dd81639acb5325dcc62e72

Proposal
=======

* [x] Drop support for LESS_RUN_IN_DEBUG (See [this commit](https://github.com/jmartinm/invenio/commit/5fdd9c0a540d2f8947dd81639acb5325dcc62e72)) (PR #3001)
* [ ] Enable less source maps (http://robdodson.me/debug-less-with-chrome-developer-tools/) - no less.js needed
* [ ] Fix ``depends=`` parameter used, for example, [here](https://github.com/inveniosoftware/invenio/blob/master/invenio/base/bundles.py#L80) so that we can have imports in the ``less`` code like [here](https://github.com/inspirehep/inspire-next/blob/master/inspire/base/static/less/inspire.less) and a change on any of the dependent files will trigger recompilation of ``less``. This is currently not happening, likely due to something missing in our [custom resolver](https://github.com/inveniosoftware/invenio/blob/master/invenio/ext/assets/extensions.py#L193)

I think this would make working with ``less`` nice while simplifying our code.

What do you think? Anybody using ``less.js`` in the browser? Or there is a good reason for keeping it that I am missing?

jacquerie:
+1 on all three points (is that a +3?). Even the author of `less.js` agrees on client side compilation: https://github.com/less/less.js/issues/1541#issuecomment-42751132.

lnielsen:
Personally I never used ``LESS_RUN_IN_DEBUG=True`` so I'm fine with the change if it implies the code.

jirikuncar:
[...]

PR #2948, opened 2015-03-26
global: post-release deprecation bump

jirikuncar:


jirikuncar:
Depends on #2923 (cc @jmartinm)

PR #2949, opened 2015-03-26
kwalitee: valid component names (legacy)

kaplun:


kaplun:
> 1: M111 unrecognized component name: kwalitee

That's not a good sign, is it? The very purpose of this PR is to introduce components for legacy branches. And kwalitee is included...

jirikuncar:
@kaplun IMHO you can remove the legacy from the commit message.

kaplun:
As you prefer :-) Done

PR #2955, opened 2015-03-26
RFC fast track deprecation policy

lnielsen:
## Problem

Some features that have barely been used may require significant effort to deprecate. The proposal is to allow getting rid of them faster.

## Proposal

Allow fast track feature deprecation (i.e. removed in current version +0.1 instead of current version +0.2) for **poorly tested, non-core and little used features for which alternatives already exists**. E.g. the remote debugging feature (#2945) was basically not used, untested, and the werkzeug debugger already does nearly the same.

### DIFF
--- a/invenio/utils/deprecation.py
+++ b/invenio/utils/deprecation.py
@@ -17,7 +17,7 @@
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
-"""Define exceptions that follows deprecation policy.
+"""Invenio deprecation policy.
 
 A minor release version (``A.B``) may deprecate certain features from previous
 releases. If a feature is deprecated in version ``A.B`` it will continue to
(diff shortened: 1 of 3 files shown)


jalavik:
Looks good :+1:

egabancho:
:+1:

kaplun:
:+1:

invenio/utils/deprecation.py
```diff
@@ -25,20 +25,54 @@
...
+be API backward compatible so that existing code does not break. Please start
+fixing all places ``old_method`` is used, so that it can be easily removed
+once Invenio v2.1 has been released.
 """
 
 
 class RemovedInInvenio22Warning(PendingDeprecationWarning):
 
-    """Mark feature that will be removed in current version +0.2."""
+    """Mark feature that will be removed in Invenio version 2.2."""
```
jirikuncar: I would try to keep everywhere `A.(B+1)` and `A.(B+2)` to simplify the version bump procedure.

--
WDYT about keeping the document neutral without mentioning Invenio?

lnielsen: My reason for making is explicit is because it goes into the documentation. 

``current`` -> ``Invenio`` is because ``current version 2.1`` doesn't makes sense. I.e. either ``Invenio version 2.1`` or ``current version +0.1``

jirikuncar: `acked`

invenio/utils/deprecation.py
```diff
@@ -25,20 +25,54 @@
...
   ``RemovedInInvenio22Warning`` is now loud by default (`DeprecationWarning`).
 - Version *2.2* will completely remove the feature.
+
+Fast track deprecation
+----------------------
+Poorly tested, non-core and little used features for which alternatives already
+exists in Invenio may opt to use fast track deprecation. If a feature is
+fast track deprecated in version ``A.B`` it will continue to work in version
+``A.B`` but raise loud warnings. It will be completely removed in version
+``A.(B+1)``.
```
tiborsimko: It may be useful to expand the description a bit, to say basically that this is an exceptional process that concerns itself with features that "nobody" uses or cares about, especially in these times of v1 -> v2 transition times. Example remote debugger.  The idea is that the services shouldn't get impression that "important user facing features" they started to rely on in their LABS instances will be randomly removed... Kind of set the expectations straight.

jirikuncar: > "nobody" uses or cares about ...

vs. "little used"?

IMHO "non-core" is pretty clear and sufficient.

jirikuncar:
@lnielsen :shipit:

invenio/utils/deprecation.py
tiborsimko: I asked to clarify the text, because "little used" and "non-core" are not clear by themselves.  There may be some *user-facing feature* that only 1 installation out of 100 uses.  Is this "little used" or not?  Globally yes, it is little used.  Locally no, the feature may be of utmost importance to that particular installation.   Using better wording in the text, with concrete example of *internal features* considered for fast deprecation, such as remote debugger in the flask world, would help to alleviate potential concerns that people may be having WRT fast deprecation of some user-facing features by the majority.

jirikuncar:
ping @lnielsen

invenio/utils/deprecation.py
lnielsen: Please see clarification added below now. I've added an escape clause that allows a fast track deprecation to be downgraded to a normal deprecation in case it has significant impact on a minority.

PR #2961, opened 2015-03-27
search: author search and collection results

jalavik:
Here is a bundle of fixes to the new search structure I did with @jirikuncar 

Basically it fixes the default author search to return the correct result if you only search for the lastname (as one word), and that records without any collections do not display in search results.

### DIFF
--- a/invenio/modules/search/searchext/units/author.py
+++ b/invenio/modules/search/searchext/units/author.py
@@ -1,0 +1,30 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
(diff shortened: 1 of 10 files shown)


invenio/modules/search/searchext/units/author.py
```diff
@@ -0,0 +1,29 @@
...
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Author search unit."""
+
+
+def search_unit(p, f, m, wl=None):
+    """Special author field handler."""
+    from invenio.legacy.bibindex.engine_utils import author_name_requires_phrase_search
```
jirikuncar: @jalavik can we move the imports on global level?

jirikuncar:
What about 'firstauthor', 'exactauthor' and other friends?

jalavik:
Added them now. They throw a PEP8 error, but we can ignore it.

Issue #2858, opened 2015-03-06
docker: prepare Dockerfile for 2.x branch


hachreak:
I tried to install invenio master from docker and there is a problem:
```bash
git clone https://github.com/inveniosoftware/invenio.git invenio2docker
cd invenio2docker
docker-compose up
```

And, the error:

```
Complete!
Loaded plugins: fastestmirror
Cleaning repos: base epel extras updates
Cleaning up Everything
Cleaning up list of fastest mirrors
 ---> b0bc273d953a
Removing intermediate container 0c0a8484f277
Step 2 : ADD requirements.txt /tmp/requirements.txt
 ---> 4f40edc1e5bc
Removing intermediate container 62d0f41a3ab1
Step 3 : ADD requirements-extras.txt /tmp/requirements-extras.txt
Service 'web' failed to build: requirements-extras.txt: no such file or directory
```

jirikuncar:
@hachreak the `Dockerfile` needs to be completely rewritten for Invenio 2.x. I have started with very simplified version in my [docker](https://github.com/jirikuncar/invenio/tree/docker) branch (https://github.com/jirikuncar/invenio/commit/568ada845d20a9594713e34644e0ccc3ed4dae6f).

tiborsimko:
In Invenio v2.0, the demo site was split into its own overlay, so I think most of Docker application building will be done in overlays.  (I had planned to set up an example for the demo site and the CERN Open Data, but I did not get to it yet.)  Inside Invenio itself, I think the Docker could be rather thin, basically just to install the software and run the tests...

jirikuncar:
> basically just to install the software and run the tests...

I would say: basically just to install the software ~~and run the tests...~~

tiborsimko:
> I would say: basically just to install the software 

(to be able to run the tests; we wouldn't run them in CMD)

jirikuncar:
> to be able to run the tests

:+1:

jirikuncar:
- [ ] add new configuration `INVENIO_ALLOWED_ENVS` with list of allowed os environment variables to ease creation of `docker-compose.yml`

**Example**
```console
export SQLALCHEMY_URI = mysql://invenio:invenio@mysql/invenio
export REDIS_HOST = redis
INVENIO_ALLOWED_ENVS=SQLALCHEMY_URI,REDIS_HOST inveniomanage runserver -h 0.0.0.0 -p 8080
```

--
[...]

PR #2965, opened 2015-03-30
docker: pythonic installation process

jirikuncar:
There is a initial Docker configuration for Invenio 2.x.  (addresses #2858)

--
cc @hachreak

### DIFF
--- a/.kwalitee.yml
+++ b/.kwalitee.yml
@@ -49,6 +49,7 @@ components:
 - dashboard
 - debug_toolbar
 - deposit
+- docker
 - docs
 - documentation
 - documents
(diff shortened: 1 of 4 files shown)


tiborsimko:
FWIW I would call this something like "docker: installation in separate containers", because it is strictly speaking not only applicable to Python apps.

tiborsimko:
> NEW Adds new configurable variable `INVENIO_ENABLE_ENV` 

Another thing: we have renamed the variable, but it stayed under the old name in the commit log message...

jirikuncar:
@tiborsimko see https://github.com/jirikuncar/invenio/commit/dcb508a590992122a5845e42f19fab5f2895de2c

PR #2966, opened 2015-03-30
knowledge: temporary knowledge creation removal

### LINKED ISSUE #698: profile bibknowledge.get_kbd_values_for_bibedit
Originally by *cdickins* on 2011-06-22

bibknowledge.get_kbd_values_for_bibedit() creates and then deletes a dynamic knowledge-base within the same request. What is the benefit of creating the knowledge-base for a single request? This should be looked into and optimized if possible.

hachreak:
* Relaxes constrictions on dynamic search function that
  force us to create temporary knowledge.
  (closes #698)

### DIFF
--- a/invenio/modules/knowledge/api.py
+++ b/invenio/modules/knowledge/api.py
@@ -574,9 +574,6 @@ def get_kbd_values(kbname, searchwith=""):
     :param kbname:     name of the knowledge base
     :param searchwith: a term to search with
     """
-    from invenio.legacy import search_engine
-
-    # first check that the kb in question is dynamic
     kb = get_kb_by_name(kbname)
     kbid = kb.id
     if not kbid:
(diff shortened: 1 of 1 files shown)


invenio/modules/knowledge/api.py
```diff
@@ -662,17 +668,10 @@ def get_kbd_values_for_bibedit(tag, collection="", searchwith="",
...
-    # in case there are concurrent calls.
-    kb_name = get_kb_name(kb_id)
-    add_kb_mapping(kb_name, tag, expression, collection)
-    # now, get stuff
-    myvalues = get_kbd_values(kb_name, searchwith)
-    # the tmp dyn kb is now useless, delete it
-    delete_kb(kb_name)
-    return myvalues
+    collection = Collection.query.filter_by(name=collection).first()
+    kbdefs = models.KnwKBDDEF(output_tag=tag, expression=expression,
```
jirikuncar: I'm not sure if this is safe. You are creating SQLAlchemy object that it's binded to a session and if it runs in request than it's auto-commited to db. Can you change `get_kbd_values_by_kbdefs` to accept `dict` with values from `kbdefs.to_dict()` instead of `KnwKBDDEF` instance? Then it should be fine to do `s/models.KnwKBDDEF/dict/`.

PR #2969, opened 2015-03-30
global: config values from environment variables

jirikuncar:
* NEW Adds new configurable variable `INVENIO_APP_CONFIG_ENVS` that can
  be set both from `invenio.cfg` and OS environment. Application factory
  iterates over comma separated list of configuration variable names and
  updates application config with equivalent OS environment value.
  (addresses #2858)

* AMENDS db8b055a6db2c01bcb174b2fee1c65723c9193bb

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,12 +17,10 @@
 
 FROM python:2.7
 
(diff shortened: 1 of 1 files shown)


tiborsimko:
+1

Issue #2967, opened 2015-03-30
tests: investigate loud warnings and errors


hachreak:
I executed tests in local and I received many warning and errors.
It's a summary:

 - [ ] `WARNING:py.warnings: /path/to/file.html RuntimeWarning: Convert string 'active' in template to unicode.`.
 - [ ] `InvenioBibFormatError: Format element SERVER_INFO could not be found.`
 - [ ] `ERROR:invenio.base:Error retrieving the blob for recid 33`
 - [ ] `AttributeError: 'NoneType' object has no attribute 'legacy_create_recstruct'`
 - [ ] `AttributeError: 'NoneType' object has no attribute 'creation_date'`
 - [ ] `KeyError: 'Default_unknown.bft'`
 - [x] `ImportError: import_string() failed for 'invenio.legacy.webcomment.webinterface'.`. See also #2958 (closed by  #2973)
 - [ ] `WorkflowError: WorkflowError(Error: ZeroDivisionError()`
 - [x] `ImportError: cannot import name InvenioWebAccessMailCookieDeletedError`. Maybe introduced when @93014d525861f278aa60acace5ec15cae106ef25 merged. (closed by  #2973)
 - [ ] `ERROR: Expected package but found module at 'invenio.modules.upgrader.engine'.`
 - [ ] `ERROR: Could not generate upgrade - /tmp/tmpwNcF3O/invenio_upgrader_test/upgrades/invenio_upgrader_test_2015_03_31_rename_me.py already exists.`
 
See all w/e: https://gist.githubusercontent.com/hachreak/2c34a5723795e0a34776/raw/master%20errors

dset0x:
> WorkflowError: WorkflowError(Error: ZeroDivisionError()

This is expected. The [test](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/workflows/testsuite/test_workflows_others.py#L78) checks if the engine allows the ZeroDivisionError to bubble up to the user. It could potentially be silenced, but it is not an actual problem with the test.

jirikuncar:
@dset0x that's fine. But the tests should not write to stderr unless they fail. If you expect something in stderr then catch it.

tiborsimko:
Superseded by Invenio 3.

PR #2971, opened 2015-03-31
global: access errors import fix

hachreak:
* FIX Fixes imports from access module.
  (addresses #2967)


tiborsimko:
> FIX Fixes imports from access module.

BTW, this change is of an internal nature, so to speak. It should not be necessary to announce it to the users in the release notes...

PR #2982, opened 2015-04-04
indexer: fix for term washing

### LINKED ISSUE #2981: indexer: exception when passing unicode to washer
In `2.1dev` there is an issue when searching for example with the crossref extension which passes Unicode string around. This happens: 

```
In [1]: from invenio.legacy.bibindex.engine_washer import lower_index_term

In [2]: lower_index_term(u"Über")
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-5310032ccacc> in <module>()
----> 1 lower_index_term(u"Über")

/home/jlavik/envs/pu/src/invenio/invenio/legacy/bibindex/engine_washer.py in lower_index_term(term)
     76     UnicodeDecodeError, so the client code may want to catch it.
     77     """
---> 78     return unicode(term, 'utf-8').lower().encode('utf-8')
     79 
     80 latex_markup_re = re.compile(r"\\begin(\[.+?\])?\{.+?\}|\\end\{.+?}|\\\w+(\[.+?\])?\{(?P<inside1>.*?)\}|\{\\\w+ (?P<inside2>.*?)\}")

TypeError: decoding Unicode is not supported
```

jalavik:
* FIX Avoids an exception from happening when passing a unicode string
  to the BibIndex engine washer. (closes #2981)

* Adds a unit test to cover this case.

Co-authored-by: Eamonn Maguire <eamonn.maguire@cern.ch>

### DIFF
--- a/invenio/modules/indexer/testsuite/test_indexer_engine.py
+++ b/invenio/modules/indexer/testsuite/test_indexer_engine.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2004, 2005, 2006, 2007, 2008, 2010, 2011, 2013, 2014 CERN.
+# Copyright (C) 2004, 2005, 2006, 2007, 2008, 2010, 2011, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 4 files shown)


invenio/modules/indexer/testsuite/test_indexer_engine_washer.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def test_lower_term_washer_encoding(self):
+        """Bibindex engine washer - lower term encoding."""
+        self.assertEqual(
+            engine_washer.lower_index_term(u"test"),
+            u"test"
+        )
+        self.assertEqual(
+            engine_washer.lower_index_term("Über"),
+            "über"
```
jirikuncar: I would rather see all as unicode.

jalavik: Right. So that would change the behavior of the original function a bit as it was always returning the bytestring. (The test above on line 45 is actually wrong(!), as it should not be `u"test"`, but simply `"test"`.).

The question is, should we allow the unicode string to survive through this function. The obvious answer is yes, but I wonder what it will break further down the line. I could give it a go and see.

jirikuncar: IMHO the strategy is *all in unicode*. Let's change it as see what gets broken.

jalavik:
Pushed an update that will make it Unicode. Prepare to open a can of worms :boom:

invenio/modules/indexer/tokenizers/BibIndexCJKTokenizer.py
```diff
@@ -93,15 +99,15 @@ def tokenize_for_words(self, phrase):
...
-            pre_tokenized = self.tokenize_for_words_default(phrase)
-            #list for keeping CJK chars and non-CJK words
+            # first, we split our phrase with default word tokenizer to make it
+            # easier later
+            words = self.tokenize_for_words_default(phrase)
+            # list for keeping CJK chars and non-CJK words
             chars = []
-            #every CJK word splits into a set of single characters
-            #for example: "春眠暁覚" into ['春','眠','暁','覚']
-            words = [ word.decode("utf8") for word in pre_tokenized]
```
jalavik: This caused an exception as it was now already Unicode, so I removed this part.

invenio/modules/indexer/testsuite/test_indexer_engine.py
```diff
@@ -19,65 +19,72 @@
...
     def test_wash_index_term_unicode(self):
-        """bibindex engine - wash index term, unicode"""
+        """bibindex engine - wash index term, unicode."""
         self.assertEqual("ελληνικό αλφάβητο",
-          bibindex_engine.wash_index_term("Ελληνικό αλφάβητο"))
+                         bibindex_engine.wash_index_term("Ελληνικό αλφάβητο"))
+
+    def test_wash_index_term_unicode_for_real(self):
+        """bibindex engine - wash index term, unicode string."""
+        self.assertEqual("ellis",
```
jalavik: Here I added a new test, as the washer did not like to be given Unicode suddenly :boom:

PR #2983, opened 2015-04-07
travis: unittest2 requirements fix

### LINKED ISSUE #2862: global: jinja2hacks are causing troubles on Python 2.7.9

hachreak:
* FIX Fixes TypeError error on markupsafe library. (closes #2862)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,6 +50,7 @@ install:
   - "sudo a2enmod rewrite"
   - "sudo mkdir /etc/apache2/ssl"
   - "sudo /usr/sbin/make-ssl-cert generate-default-snakeoil /etc/apache2/ssl/apache.pem"
+  - "travis_retry pip install unittest2"
   - "travis_retry pip install -r requirements.txt --quiet"
   - "travis_retry pip install -r .travis-$REQUIREMENTS-requirements.txt --allow-all-external --quiet"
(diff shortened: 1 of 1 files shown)


jirikuncar:
@hachreak shall we integrate it to maint-2.0 too?

hachreak:
ping @jirikuncar

tiborsimko:
BTW this commit adds `unittest2` only to Travis, while developers using Python 2.7.9 locally may also be bitten by this error the first time they run the tests?  It may be better to add a hack to `setup.py` too, though `unittest2` is already in `tests_require` there...

hachreak:
@tiborsimko The error was not correlated with the python version.
See my [last comment](https://github.com/inveniosoftware/invenio/issues/2862#issuecomment-90508434) (and maybe the issue that I linked).

tiborsimko:
> The error was not correlated with the python version.

OK.  We started to see it the problem at the time of the Python version upgrade, I think.  

Anyway, my main point was that this branch fixes Travis configuration only, while non-Travis users may also see the problem if they don't have `unittest2` previously installed...

PR #2986, opened 2015-04-08
docker: make configuration more useful for developers

crepererum:
Currently docker does not seem very helpful for developers. The following patches will simplify and speed up the process of setting up virtual containers for development, including source code sharing with the host. Each patch can be applied as a single atomic change, but keep in mind that some of them depend on previous ones. They are rebased against `maint-2.0` + #2991.


scripts/setup_devmode.sh
```diff
@@ -0,0 +1,45 @@
...
+# Sets the configuration of Invenio to defaults which are sutable for
+# developers.
+#
+# !!!                  WARNING                      !!!
+# !!! Do NOT use these configurations in production !!!
+#
+
+echo "Load developer configuration. Do NOT use this in production mode!"
+inveniomanage config create secret-key
+inveniomanage config set CFG_EMAIL_BACKEND flask_email.backends.console.Mail
```
lnielsen: +1 for making these changes. While you improving it, it might be worthwhile reducing the number of calls to ``inveniomanage config`` since each call takes a couple of seconds and all they do is to write to a text file. Something along the lines of:

```
$ cat <<EOF >> var/invenio.base-instance/invenio.cfg
PACKAGES_EXCLUDE = []
...
EOF
```

jirikuncar: :+1:

crepererum:
@jirikuncar last action this week, I consider this PR finished. You might want to suggest commit squashing (if so, please inform me how many commits you prefer). I'll then address your comments on Monday.

Dockerfile
```diff
@@ -15,16 +15,54 @@
...
+
+
+###############################################################################
+## 2. Code (changing)                                                        ##
+###############################################################################
+
+# select proper requierments level
+#ENV REQUIREMENTS lowest
+#ENV REQUIREMENTS release
+ENV REQUIREMENTS devel
```
lnielsen: Perhaps use release requirements by default, and then we just let Travis test development versions?

crepererum: Then we use 2 Dockerfiles for 2 different images. One for production and one for devs. Would that be an option?

lnielsen: For developers I would still use the latest released pypi package (which is likely newer than what is installed on production - see how packages are managed on Zenodo [here](http://zenodo.readthedocs.org/en/latest/development.html#python-requirements)).

The devel requirements are only used to detect possible problems with upcoming versions of dependent packages. These devel packages might have big fluctuations, and thus quite unreliable to dependent on for developers.

crepererum: After investigating your concern: It is currently NOT possible two have 2 Dockerfiles until docker/compose#1075 is in place, sorry. So can we build a big image until this part is fixed and separate dev and production images (and Dockerfiles) after the referred PR is merged? There are things we need to do to slim down our production image anyway (e.g. https://intercityup.com/blog/downsizing-docker-containers.html) so this would make a great new PR.

lnielsen: My concern is not about the production image. It's only about that in my opinion the dev image should be using the release requirements instead of dev requirements. 

Note that:
- Devel requirements != requirements for developers
- Release requirements != requirements for production sites

rather:
- Devel requirements == Latest *unreleased* version of the requirements
- Release requirements != Latest released version of package on pypi.

crepererum:
Ready to go

lnielsen:
```
* NOTE Be careful when modifying the Invenio configuration file
  directly because the file is parsed as python and therefore a
  mistake will lead to a non-starting Invenio setup.
```
I'm not sure what you mean with this? Is it just that ``invenio.cfg`` is a Python file? If it shouldn't be in the commit message but rather added in the documentation.

lnielsen:
``/Therefor/Therefore/`` in:
```
* Sets correct redis connection in compose file. The required config
  flags for the worker and for the web part are different. Therefor
  these changes are only required for the worker node.
```

lnielsen:
```* Adds more file types to docker ignore ... ``` is missing a dot in the end.

**Comment by lnielsen** [comment]
Otherwise looks good to me. :shipit: Thanks for writing documentation! :-)

**Comment by crepererum** [comment]
Addressed your comments, thank for the feedback :)

**Comment by crepererum** [comment]
Rebased against `maint-2.0` again.

I added a new commit without touching the first 3 ones to not force you review them again. I can squash the last one if you want to. Using this stage (first 3 commits + the last new one) gives us a docker environment that can also be used for overlay development (tested using the demosite). I will add a PR for the demosite later, so we have a complete example.

Just to give you an idea of the my personal TODOs (aka wish list for further work):
 - slim down the docker image to make it deployment ready (first tests show size shrinking from 1300MB to around 800MB)
 - add the docker image to the official [Docker Hub](https://hub.docker.com/) (that would simplify overlay development even more)

File: Dockerfile [hunk 3]
```diff
@@ -15,16 +15,73 @@
...
+###############################################################################
+
+# select proper requierments level
+#ENV REQUIREMENTS lowest
+ENV REQUIREMENTS release
+#ENV REQUIREMENTS devel
+
+# add current directory as `/code`.
 ADD . /code
 WORKDIR /code
```
tiborsimko: Adding whole code base at this point seems too soon.  It would invalidate the docker build cache even if only docs files change in Invenio.  It is better to copy only pip requirements files and friends at this point, and install those, as we do in other packages.  This has a consequence that, unless requirements change, the docker can reuse prior layers, and so the new docker image will be built in no time.  (without running pip repetitively)

Dockerfile
```diff
@@ -15,16 +15,73 @@
...
+    pip install mock && \
+    pip install unittest2 && \
+    pip install watchdog
+
+# install nodejs requirements
+RUN npm update && \
+    npm install --silent -g bower less clean-css uglify-js requirejs
+
+# add invenio user
+RUN useradd --home-dir /home/invenio --create-home --shell /bin/bash invenio
```
tiborsimko: I'd also force UID to 1000, as we do in other packages, which is practical for volume mounting.

crepererum:
All issues addressed, ready for merge (finally).

crepererum:
Closed because of merge conflicts (wrong branch, my fault). See #3032 for the up-to-date version.

PR #2987, opened 2015-04-08
WIP search: query enhancement

marioskogias:
* Implements a decorator function that wraps the SearchEngine's
  query property. This decorator causes the enhancer functions
  specified in the cfg['SEARCH_QUERY_ENHANCERS'] to be applied to
  the query AST one after the other.
  The initial query AST is also kept.


invenio/modules/search/api.py
```diff
@@ -47,6 +48,7 @@ def parser(self):
         return query_parser
 
     @cached_property
+    @enhance_query()
```
jirikuncar: @marioskogias please move it inside `search` method as the query enhancers might need access to it's keyword arguments.

invenio/modules/search/enhancers/enhancer_decorator.py
```diff
@@ -0,0 +1,47 @@
...
+        search_obj._initial_query
+    """
+    def decorator(decorated_fn):
+        def wrapper(search_obj):
+            tree = decorated_fn(search_obj)
+            search_obj.query = tree
+            for enhancer in cfg['SEARCH_QUERY_ENHANCERS']:
+                if isinstance(enhancer, six.string_types):
+                    enhancer = import_string(enhancer)
+                    enhanced_query = enhancer(search_obj, *args, **kwargs)
```
jirikuncar: -4 spaces indentation

jirikuncar:
@marioskogias please try to start with implementing the actual query enhancers and we can come with configuration later.

invenio/modules/search/enhancers/collection_filter.py
```diff
@@ -0,0 +1,96 @@
...
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+from invenio.base.globals import cfg
+from invenio_query_parser.ast import (
+        AndOp, OrOp, NotOp, KeywordOp, Keyword, Value
+)
+
+def kw_formatter(val):
+    return KeywordOp(Keyword("collection"), Value(val))
```
jirikuncar: shouldn't it be `collection:"value"` hence `KeywordOp(Keyword("collection"), DoubleQuotedValue(val))`?

invenio/modules/search/searchext/engines/native.py
```diff
@@ -48,7 +48,7 @@ def search(self, user_info=None, collection=None):
     from invenio.modules.search.walkers.search_unit import SearchUnit
     results = get_results_cache(self._query, collection)
     if results is None:
-        results = self.query.accept(SearchUnit())
+        results = self.enhanced_query.accept(SearchUnit())
```
jirikuncar: all of this should be moved back to `api.py`

invenio/modules/search/api.py
```diff
@@ -59,6 +59,13 @@ def query(self):
     def search(self, user_info=None, collection=None):
         """Search records."""
         user_info = user_info or current_user
+        # Enhance query first
+        self.enhanced_query = self.query
+        for enhancer in cfg['SEARCH_QUERY_ENHANCERS']:
```
jirikuncar: you can cache the list of query enhances, but I would not add anything to `self.enhanced_query` because it depends on arguments the `search` method gets.

invenio/modules/search/enhancers/basic_enhancers.py
```diff
@@ -0,0 +1,29 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Very simple enhancer functions used as proof of concept
+
+    Every enhancer function accepts the SearchEngine instance as the first
```
jirikuncar:
```diff
-    Every ...
+Every ...
```

invenio/modules/search/enhancers/basic_enhancers.py
```diff
@@ -0,0 +1,29 @@
...
+    parameter and an arbitrary number of parameters and keyword values defined
+    in the decorator. It should return the new enhanced query AST.
+"""
+
+
+def identical_enhancer(search_obj, *args, **kwargs):
+    return search_obj.query
```
jirikuncar: if it's "identity" on first argument then we should return `search_obj`.

--
PS: `s/search_obj/query/`

invenio/modules/search/enhancers/collection_filter.py
```diff
@@ -0,0 +1,96 @@
...
+    Policy ELSE:
+    cc AND NOT r'
+    """
+
+    r_prime = set(restricted_cols) - set(permitted_restricted_cols)
+
+    r_prime_list = list(r_prime)
+    r_prime_tree = None
+
+    def _format_terms(term_list):
```
jirikuncar:
```python
return reduce(
    OrOp, [KeywordOp(Keyword("collection"), DoubleQuotedValue(val))
           for val in term_list]
) if len(term_list) else None
```

invenio/modules/search/enhancers/collection_filter.py
```diff
@@ -0,0 +1,96 @@
...
+                kw = format_vals(val)
+                or_term = OrOp(or_term, kw)
+            return or_term
+        elif len(term_list) == 1:
+            return term_list[0]
+        else:
+            return None
+
+    r_prime = set(restricted_cols) - set(permitted_restricted_cols)
+    r_prime_list = list(r_prime)
```
jirikuncar: you don't need to make `list` for `set` if you don't use `__getitem__` ;)

invenio/modules/search/enhancers/facet_filters.py
```diff
@@ -0,0 +1,109 @@
...
+    return out
+
+def format_facet_tree_nodes(facet_dict):
+    """Create the extra nodes to add to the AST
+
+    :param facet_fict: the grouped facets dictionary
+    :param query: the query AST so far
+    :return: the enhanced AST
+    """
+    def _format_terms(key, vals):
```
jirikuncar: see the solution with `reduce` above ...

invenio/modules/search/searchext/engines/native.py
marioskogias: But get_results_cache is native-indexer specific, isn't it? If this is moved to the api then what will the search() function from searchext/engines/native.py do?

invenio/modules/search/api.py
jirikuncar: @marioskogias do you have any updates on this?

PR #2989, opened 2015-04-09
config: add inveniomanage config set_from_file

crepererum:
Adds the ability to set multiple instance variables with one call to inveniomanage. The effect equals multiple calls to `inveniomanage config set`, but because Invenio is only loaded once it is way faster when setting many variables.

This is a preparation for #2986.


crepererum:
Will be changed to another approach using another PR.

jirikuncar:
### Proposal

`inveniomanage config locate`

PR #2991, opened 2015-04-09
global: new `inveniomanage config locate` command

crepererum:
Adds a command to request the location of the instance config file. Proper error handling (message + return code) is implemented in case the config file does not exist.

Rebased against: `maint-2.0`
Replacement for: #2989
Prepares: #2986

### DIFF
--- a/invenio/base/scripts/config.py
+++ b/invenio/base/scripts/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2013, 2014 CERN.
+# Copyright (C) 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -22,12 +22,15 @@
 from __future__ import print_function
 
 import ast
-import imp
-import sys
 import errno
+import imp
+import os.path
+import sys
 
 from pprint import pformat
+
 from flask import current_app
+
 from invenio.ext.script import Manager, change_command_name, \
     generate_secret_key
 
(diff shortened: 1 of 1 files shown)


tiborsimko:
> Adds a command to request the location of the instance config file.

Besides getting location information, one may perhaps want to print the raw content of the file immediately.  E.g. `inveniomanage config show` or something.

invenio/base/scripts/config.py
```diff
@@ -119,6 +122,17 @@ def set_(name, value, filename='invenio.cfg'):
...
 
 
+@manager.command
+def locate(filename='invenio.cfg'):
+    """Print the location of the configuration file."""
+    try:
+        with current_app.open_instance_resource(filename, 'r') as config_file:
+            print(os.path.abspath(config_file.name))
+    except IOError:
+        print("ERROR: configuration file does not exist.")
```
jirikuncar: it should probably go to `stderr`

jirikuncar:
> Besides getting location information, one may perhaps want to print the raw content of the file immediately.

Maybe we can extend `inveniomanage config get \*` or `inveniomanage config get '^SEARCH_'`. Shall we create new issue?

tiborsimko:
> Maybe we can extend `inveniomanage config get \*`

We could get into output format change troubles here.  Currently, getting single variable returns directly value as a string:

```console
$ inveniomanage config get CFG_SITE_URL
http://localhost:4000
```

which seems alright.

Now if we accept `CFG_FOO_*`, what would we return?  Python code excerpt always?  What if the wildcard matches only one variable?

If we want to keep "get" for getting simple values, then it seems better to me to introduce another operation for "getting many"...

PR #2993, opened 2015-04-09
oairepository: GetRecord fix

### LINKED ISSUE #2990: oairepository: GetRecord does not find record via identifier
See zenodo/zenodo#172

Problem is that when a record is part of a restricted collection (which allows guest access) it will be filtered out of the final list of recids.

Fix incoming by @bouzlibop

bouzlibop:
* Correctly handles getting single record from
  restricted collection. (closes #2990)

### DIFF
--- a/invenio/legacy/oairepository/server.py
+++ b/invenio/legacy/oairepository/server.py
@@ -1,5 +1,5 @@
 # This file is part of Invenio.
-# Copyright (C) 2009, 2010, 2011, 2014 CERN.
+# Copyright (C) 2009, 2010, 2011, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -62,10 +62,13 @@ from invenio.config import \
 from intbitset import intbitset
 from invenio.utils.html import X, EscapedXMLString
 from invenio.legacy.dbquery import run_sql, wash_table_column_name
-from invenio.legacy.search_engine import record_exists, get_all_restricted_recids, get_all_field_values, search_unit_in_bibxxx, get_record, search_pattern
+from invenio.legacy.search_engine import record_exists, get_all_restricted_recids, \
+    get_all_field_values, search_unit_in_bibxxx, get_record, search_pattern, \
+    get_records_that_can_be_displayed
 from invenio.modules.formatter import format_record
 from invenio.legacy.bibrecord import record_get_field_instances
 from invenio.ext.logging import register_exception
+from flask_login import current_user
 from invenio.legacy.oairepository.config import CFG_OAI_REPOSITORY_GLOBAL_SET_SPEC
 from invenio.utils.date import localtime_to_utc, utc_to_localtime
 from invenio.base.globals import cfg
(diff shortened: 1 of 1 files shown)


bouzlibop:
@lnielsen please take a look at these changes. I tested it locally and now it is returning me the correct result, however I'm not sure how to be sure with this legacy code if this change doesn't break something?

invenio/legacy/oairepository/server.py
```diff
@@ -575,12 +578,10 @@ def oai_get_recid(identifier):
     if identifier:
         recids = search_pattern(p=identifier, f=CFG_OAI_ID_FIELD, m='e', ap=-9)
         if recids:
-            restricted_recids = get_all_restricted_recids()
+            displayable_recids = get_records_that_can_be_displayed(UserInfo(), recids)
```
jirikuncar: maybe we should use `current_user` (`from flask_login import current_user`).

invenio/legacy/oairepository/server.py
```diff
@@ -575,12 +578,10 @@ def oai_get_recid(identifier):
...
             for recid in recids:
```
lnielsen: ``for recid in displayable_recids:``

invenio/legacy/oairepository/server.py
```diff
@@ -575,12 +578,10 @@ def oai_get_recid(identifier):
...
-                if record_exists(recid) > 0 and recid not in restricted_recids:
+                if record_exists(recid) > 0 and recid in displayable_recids:
```
lnielsen: ``if record_exists(recid) > 0:``

bouzlibop:
Thanks for the comments :smile: I just pushed updated version of this PR.

invenio/legacy/oairepository/server.py
```diff
@@ -62,10 +62,13 @@
...
 from invenio.utils.html import X, EscapedXMLString
 from invenio.legacy.dbquery import run_sql, wash_table_column_name
-from invenio.legacy.search_engine import record_exists, get_all_restricted_recids, get_all_field_values, search_unit_in_bibxxx, get_record, search_pattern
+from invenio.legacy.search_engine import record_exists, get_all_restricted_recids, \
+    get_all_field_values, search_unit_in_bibxxx, get_record, search_pattern, \
+    get_records_that_can_be_displayed
 from invenio.modules.formatter import format_record
 from invenio.legacy.bibrecord import record_get_field_instances
 from invenio.ext.logging import register_exception
+from flask_login import current_user
```
jirikuncar: please check with @lnielsen about the `current_user` usage and in case it's fine please re-order the imports.

lnielsen:
Let's leave it with ``current_user``, and if needed I'll apply a patch on Zenodo. The module needs major rewrite anyway.

Just to log what the problem concerns:
- Current ``maint-2.0`` behaviour is that only public records are exposed via OAI-PMH independent of if you are logged in or not.
- If ``current_user`` is used, logged in users can access private records as well and thus we don't maintain current behaviour which may impact ease of debugging problems. I don't think any harvesters are logged in, so this will not impact them. 
- If we require user not to be logged in when accessing OAI-PMH then we also break current behaviour which allow exploring OAI-PMH even when you are logged in.
- Using ``UserInfo()`` instead might have side effects which are not intended.

jirikuncar:
@lnielsen @bouzlibop please add NOTE to commit message describing the problem (https://github.com/inveniosoftware/invenio/pull/2993#issuecomment-92338962) and then :shipit: . Thanks

bouzlibop:
Alright, I have just added a NOTE with the list of concerns you provided @lnielsen. Let me know in case there should be some changes made there.

lnielsen:
Sorry, I was too slow. Can you change the extra bullet to (not bullet list within bullet list I don't think work well with the automatic release notes generation):

```
* NOTE Changes current behavior of OAI-PMH server for logged in users to take
  into account all records a user can view and not only public records.
``

I.e. just to notify people that behaviour is changing. The implementation issue can be looked up here.

bouzlibop:
Sure, thanks for the clarification - I wasn't sure how to indicate this problem in NOTE. I rebased and repushed with new commit message.

Issue #2958, opened 2015-03-27
search: remove legacy webinterface


jirikuncar:
- [x] remove `invenio/legacy/websearch/webinterface.py`
- [x] check all imports
- [x] implement `/unapi` url handler (@tiborsimko is it used somewhere?)

jmartinm:
will this be removed as well https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/formatter/engine.py#L519-L545 ??

Its quite horrible, for example, I am trying to create a brief record template and every time there is a Jinja exception instead of seeing the real error this fallback happens and some default Invenio template is being rendered :scream: 

Shall I create separate issue?

PR #2997, opened 2015-04-10
formatter: exception fallback removal

jirikuncar:
* NOTE Fallback template rendering has been removed and standard
  exception logging is in place.  (addresses #2958)

### DIFF
--- a/invenio/modules/formatter/engine.py
+++ b/invenio/modules/formatter/engine.py
@@ -516,33 +516,12 @@ def format_record_1st_pass(recID, of, ln=CFG_SITE_LANG, verbose=0,
                                                        needs_2nd_pass)
 
         return out, needs_2nd_pass
-    except Exception, e:
-        register_exception(prefix="An error occured while formatting record %s in %s" %
-                           (recID, of),
-                           alert_admin=True)
-        #Failsafe execution mode
-        import invenio.legacy.template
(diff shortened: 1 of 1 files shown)


jirikuncar:
@jmartinm can you please review it? Thanks

jmartinm:
This now works as expected and an exception is raised when there is a problem in a template.

The only issue I see is that, for example, for testing I created an error in a template called Author_HTML_Brief.tpl and the following traceback is generated:

https://gist.github.com/anonymous/5f0e0266d8d389a71973

The error is correct but there is no reference of the template that actually caused it, which is not very nice.

invenio/modules/formatter/engine.py
```diff
@@ -516,33 +516,12 @@ def format_record_1st_pass(recID, of, ln=CFG_SITE_LANG, verbose=0,
...
-            </span><br/>""" % recID
-        return out + websearch_templates.tmpl_print_record_brief(ln=ln,
-                                                                 recID=recID,
-                                                                 ), False
+    except Exception:
+        current_app.logger.exception(
+            "An error occured while formatting record {recid} in {of}".format(
+                recid=recID, of=of
+            ))
+        raise
```
jirikuncar: @jmartinm I have added `raise` to propagate the exception. I hope soon we can re-factor whole *formatter engine*.

PR #3005, opened 2015-04-13
redis: legacy module deprecation

### LINKED ISSUE #2885: utils: deprecate invenio.utils.redis in favor of invenio.ext.cache
Currently redis host must be set both with ``CACHE_REDIS_URL`` (invenio.ext.cache) and ``CFG_REDIS_HOSTS`` (invenio.utils.redis). This is confusing especially when redis is not running on localhost (which is the default value).

* ``CFG_REDIS_HOSTS`` comes from Invenio legacy code, and uses Nydus package with consistent hashing to spread load over several caches.
* ``CACHE_REDIS_URL`` uses Flask-Cache and can be configured with many different cache backends.

Following has to be done:
- [x] Deprecate use of ``CFG_REDIS_HOSTS`` in favor of invenio.ext.cache
- [ ] Add support for using Nydus with consistent hashing in Flask-Cache.

lnielsen:
* INCOMPATIBLE Deprecates use of invenio.utils.redis in favor of
  invenio.ext.cache (closes #2885).


jirikuncar:
@lnielsen `git commit --amend` & `s/redis:/global:/`

PR #3009, opened 2015-04-13
celery: only accept msgpack content

### LINKED ISSUE #3003: celery: deprecation warnings issue
## Abstract
We use pickle as the serialization method when communicating with celery. This method will be disabled in future standard configurations because it has security issues. We should decide how to move on.

## Problem description

Celery starts to emit the following warning:
 ```
Starting from version 3.2 Celery will refuse to accept pickle by default.

The pickle serializer is a security concern as it may give attackers
the ability to execute any command.  It's important to secure
your broker from unauthorized access when using pickle, so we think
that enabling pickle should require a deliberate action and not be
the default choice.

If you depend on pickle then you should set a setting to disable this
warning and to be sure that everything will continue working
when you upgrade to Celery 3.2::

    CELERY_ACCEPT_CONTENT = ['pickle', 'json', 'msgpack', 'yaml']

You must only enable the serializers that you will actually use.


  warnings.warn(CDeprecationWarning(W_PICKLE_DEPRECATED))
```

## Possible solutions
We should solve this by implementing one of the following solutions.

### a) pickle
Follow the given advice and set the `CELERY_ACCEPT_CONTENT` config variable. If so, we should update the documentation to reflect this and we must mention that the broker should be secured on the network side. The pickle method has some serious security problems because it enables remote code execution.

### b) something else
Switch over to a slower but more secure method, e.g. `msgpack` (which is kind of JSON but in binary form and faster, see [msgpack.org](http://msgpack.org/)). This would solve the security problem but we have to evaluate the performance impact.

## Importance and suggestion
It does not seem that celery 3.2 is close yet. The broker should be secured anyway, but "executing any task" and "executing any code" is a quite different security level. So we might want at least go for solution a). The documentation does not describe the security impact of unsecured worker configurations right now (or I didn't see it).

### PR DESCRIPTION
**Pull request by crepererum**
SECURITY Forces Celery to only accept msgpack content when using standard configuration. This disallows pickle messages which can be used for remote code execution. (closes #3003)

I didn't write extra documentation as mentioned in #3003, because our standard configuration is secure now and you deliberately change the configuration to something insecure now. If you select `pickle` as your config then it's clearly your fault.

Rebased against: `maint-2.0`

### DIFF
--- a/.kwalitee.yml
+++ b/.kwalitee.yml
@@ -37,6 +37,7 @@ components:
 - bulletin
 - cache
 - checker
+- celery
 - circulation
 - classifier
 - cloudconnector
--- a/invenio/celery/config.py
+++ b/invenio/celery/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2013, 2014 CERN.
+# Copyright (C) 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -23,9 +23,10 @@
 
 def default_config(config):
     """Default configuration."""
-    ## Broker settings
-    ## ---------------
+    # Broker settings
+    # ---------------
     config.setdefault("BROKER_URL", "redis://localhost:6379/1")
+    config.setdefault("CELERY_ACCEPT_CONTENT", ['msgpack'])
 
     # Extra modules with tasks which should be loaded
     # The Invenio Celery loader automatically takes care of loading tasks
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by lnielsen** [comment]
Looks good. Can I get you to fix kwalitee issues from http://kwalitee.invenio-software.org/inveniosoftware/invenio/branches/e5d101fd020a8582680ab4c533a4e12f0485ec76/crepererum:fix/celery_pickle2 (sorry, I know it's annoying to fix other old code formatting problems :-) Thanks a lot.

**Comment by lnielsen** [comment]
Just to check, you want this to go in ``maint-2.0`` right and not master?

**Comment by crepererum** [comment]
So there are pros and cons for the `main-2.0`.

Pro:
 - It's a security patch.
 - None of our parts uses the other possible serialization methods.

Con:
 - It's a breaking change in case someone uses external software with our broker.

The decision part is the following: Is the broker an API (documented or undocumented)? How high is the probability that someone uses it with another serialization method? And finally: is this amount high enough to satisfy that we postpone a security patch?

My own opinion is that admins or devs who are using our broker with external software (and other serialization methods) will understand the clearly marked change and can just set the right config flag. For installations which use the standard way, this is clearly a standard security update.

(kwalitee fix will follow)

**Comment by jirikuncar** [comment]
:shipit:

**Comment by lnielsen** [comment]
``maint-2.0`` it is. Last change, can you add the following the commit message (just so that admins know what to do if they change something):

```
* INCOMPATIBILITY If you use any Celery serializer other than msgpack, you must
  update configuration variable CELERY_ACCEPT_CONTENT to include that 
  serializer.
```

crepererum:
Ready to go

PR #3028, opened 2015-04-16
records: clean CFG_CERN_SITE, activate refs count

### LINKED ISSUE #3026: records: reference count is wrong
Reference count is wrong

### LINKED ISSUE #3027: global: CFG_CERN_SITE still in code
In utils.py of record module

PXke:
* Removes a CFG_CERN_SITE occurence. (closes #3027)

* Fixes the way references are counted. (closes #3026)

### DIFF
--- a/invenio/modules/records/utils.py
+++ b/invenio/modules/records/utils.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2014 CERN.
+# Copyright (C) 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -29,7 +29,7 @@ from flask import g, request
 from werkzeug.utils import cached_property, import_string
 
 from invenio.base.globals import cfg
-from invenio.config import (CFG_CERN_SITE, CFG_INSPIRE_SITE,
+from invenio.config import (CFG_INSPIRE_SITE,
                             CFG_BIBRANK_SHOW_CITATION_LINKS)
 from invenio.ext.cache import cache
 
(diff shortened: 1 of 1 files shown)


PXke:
2.0.2 ? :smile:

jirikuncar:
@tiborsimko I don't have enough knowledge about reference counting. Can you please have a look? Thanks

tiborsimko:
- We don't really need the `get_field_tags("reference")` etc dance anymore, we can simply introduce "number_of_references" in "atlantis.cfg" as a derived field, like we have "number_of_authors".  This would make nice analogy with "number_of_citations" and friends.   This would simplify things.
(E.g. notice your code does not even use `reftag` counting anymore, since not needed.)

- Do you need to remove `CFG_CERN_SITE`?  Currently it would affect CDSLABS.  (CC @ludmilamarian @egabancho @kasioumis)

- Generally speaking, It is good to go for removal of `CFG_CERN_SITE`, because looking into the history, the original intention was to show or not to show the reference counts, which can be better done via views.  However, the task #3027 says "global" removal, while this PR would only remove one occurrence.  Two options: (a) use "addresses" instead of "closes",  or (b) change title of #3027.  The latter seems preferred, unless you or somebody else plans to work on removing the remaining `CFG_CERN_SITE`? occurrences?  

- BTW  citation counts in `records.utils` have the same view-related issue, see `CFG_BIBRANK_SHOW_CITATION_LINKS` in `records.utils`. Need to refactor this one as well? Would be nice to do it along... Could affect INSPIRELABS (CC @kaplun @jmartinm)

- (Note that some of my above comments might call for more extensive `master` refactoring, not `maint-2.0` strictly speaking.  Depending on your needs of the moment vs nicer future longer cleaning goals...)

jirikuncar:
This can not be safely applied to `maint-2.x` branches. If you need it for future release please re-submit the patch to *Invenio-Records*. Thanks

PR #3032, opened 2015-04-17
docker: make configuration more useful for developers

crepererum:
Replaces #2986


requirements-docs.txt
```diff
@@ -1,3 +1,8 @@
+# !WARNING! !WARNING! !WARNING! !WARNING! !WARNING! !WARNING! !WARNING!
+#      Do NOT delete this file, even when it seems to be useless!
+#      It used by read Read the Docs to build our documentation!
+# !WARNING! !WARNING! !WARNING! !WARNING! !WARNING! !WARNING! !WARNING!
+
 -r requirements.txt
```
jirikuncar: you have removed `requirements.txt` hence remove this line too. Thanks

crepererum:
OK.
Please wait until travis switches to greenlight, because of the new requirements.txt generator.

crepererum:
Mmh, the new generator seems to discover some problems with our minimal (or unspecified) requirements. I hope after multiple re-pushs we're back to normal. Let's see...

crepererum:
Last commit contains some Friday afternoon typos 😀 I'll fix them on Monday.

crepererum:
OK, I'm finally very very satisfied. I hope that all wished, comments, critics and dreams are addressed now. I rate it as "ready to go".

requirements.py
```diff
@@ -16,29 +16,162 @@
...
+        if not pkg.extras:
+            include = True
+        else:
+            for e in pkg.extras:
+                if e in args.extras:
+                    include = True
+
+        if include:
+            # remove extra marker
+            pkg.extras = set()
```
jirikuncar: This might be important for certain packages. (e.g. `SQLAlchemy-Utils[encrypted]`).

crepererum: `This` = what? The `FIXME`? No, it is not. it is only important in case you specify a requirement at multiple places and using different dependency ranges, e.g. for `docs` you specify `Sphinx` to be between 2.0 and 2.2 and for `funny_feature` you specify `Sphinx` to be between 1.9 and 2.1. In this case we don't do any smart interception of ranges. And I doubt that we should do this here. Instead we should find a way better installation method than pip, which sucks when it comes to requirement calculation.

PR #3060, opened 2015-04-28
docker: image slim-down

crepererum:
* Slims down docker image by building on top of less bloated base image
  and only install what is really required. Also purges unneeded
  packages, flushes caches and clean temporary files. All these parts
  should not be in a production image and are also not required by
  developers. You can still install components when extending the
  Invenio base image.

* Removes dpkg package docs from the production image because admins
  should not use a slim docker image as a knowledge base.

* Sets correct MySQL username and db variable in boot script since we
  rely on them and Invenio might change these defaults.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,17 +19,41 @@
 ## 1. Base (stable)                                                          ##
 ###############################################################################
 
-FROM python:2.7
+FROM python:2.7-slim
 MAINTAINER CERN <info@invenio-software.org>
 
 # nodejs repo
 # detects distribution and adds the right repo
 # See: https://github.com/joyent/node/wiki/Installing-Node.js-via-package-manager#debian-and-ubuntu-based-linux-distributions
-RUN curl -sL https://deb.nodesource.com/setup | bash -
+RUN apt-get update && \
+    apt-get -qy install --fix-missing --no-install-recommends \
+        curl \
+        && \
+    curl -sL https://deb.nodesource.com/setup | bash -
 
 # install requirements from repos
+# also clean up apt
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,17 +19,41 @@
...
+        mysql-client \
+        nodejs \
+        poppler-utils \
+        subversion \
+        sudo \
+        && \
+    apt-get clean autoclean && \
+    apt-get autoremove -y && \
+    rm -rf /var/lib/{apt,dpkg}/ && \
+    (find /usr/share/doc -depth -type f ! -name copyright|xargs rm || true) && \
```
dset0x: I would consider `-delete` instead of passing to `xargs`, which can run into troubles if the script hits the arg count limit.

jirikuncar: :+1:

crepererum:
Should be ready now

scripts/docker_devboot.sh
```diff
@@ -71,6 +71,8 @@ run_master() {
 CFG_SITE_URL = u'http://localhost:5000'
 CFG_SITE_SECURE_URL = u'http://localhost:5000'
 CFG_REDIS_HOSTS = {'default': [{'db': 0, 'host': 'cache', 'port': 6379}]}
+CFG_DATABASE_NAME = u'invenio'
```
jirikuncar: aren't these defaults?

crepererum: Yes, but some overlays tend to overwrite these. And because we want to reuse the script for as many cases as possible, I enforce that default here.

jirikuncar: If they do they are on their own. Probably they are going to use external database anyway so the docker will just complicate their lives.

PR #3062, opened 2015-04-28
global: fix for multiple binding problems

crepererum:
This reworks the way we're binding to network interfaces and ports in different situation. See commits for details.

Rebased against: `maint-2.0`

### DIFF
--- a/docs/getting-started/overlay.rst
+++ b/docs/getting-started/overlay.rst
@@ -223,8 +223,8 @@ the application).
     (myoverlay)$ inveniomanage config set CFG_DATABASE_NAME mysql-database
     (myoverlay)$ inveniomanage config set CFG_DATABASE_USER mysql-user
     # HOST configuration (for redirects, etc.)
-    (myoverlay)$ inveniomanage config set CFG_SITE_URL http://0.0.0.0:4000
-    (myoverlay)$ inveniomanage config set CFG_SITE_SECURE_URL https://0.0.0.0:4000
+    (myoverlay)$ inveniomanage config set CFG_SITE_URL http://invenio.example.com
+    (myoverlay)$ inveniomanage config set CFG_SITE_SECURE_URL https://invenio.example.com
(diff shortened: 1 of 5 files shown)


invenio/ext/script/__init__.py
```diff
@@ -184,8 +186,14 @@ def register_manager(manager):
     manager.add_command("show-urls", ShowUrls())
     manager.add_command("shell", Shell())
     parsed_url = urlparse(manager.app.config.get('CFG_SITE_URL'))
-    port = parsed_url.port or 80
-    host = parsed_url.hostname or '127.0.0.1'
+    port = manager.app.config.get(
+        'CFG_BIND_PORT',
```
jirikuncar: What about `BIND_SERVER_PORT` and `BIND_SERVER_ADDRESS`? We should also document it in https://github.com/inveniosoftware/invenio/pull/3062/files#diff-dc22af6b54d602ea39f4e47197d8731dR20.

docs/getting-started/overlay.rst
```diff
@@ -223,8 +223,8 @@ the application).
     (myoverlay)$ inveniomanage config set CFG_DATABASE_NAME mysql-database
     (myoverlay)$ inveniomanage config set CFG_DATABASE_USER mysql-user
     # HOST configuration (for redirects, etc.)
-    (myoverlay)$ inveniomanage config set CFG_SITE_URL http://0.0.0.0:4000
-    (myoverlay)$ inveniomanage config set CFG_SITE_SECURE_URL https://0.0.0.0:4000
+    (myoverlay)$ inveniomanage config set CFG_SITE_URL http://localhost:4000
+    (myoverlay)$ inveniomanage config set CFG_SITE_SECURE_URL https://localhost:4000
```
jirikuncar: this might be confusing ... shall we completely remove it or show `http://invenio.example.com` and `https://invenio.example.com`?

PR #3068, opened 2015-04-28
search: initial implementation of query enhancers

jirikuncar:
* Implements a mechanism that enhances users' queries. The enhancer
  functions are specified in the 'SEARCH_QUERY_ENHANCERS' and later they
  are applied to the query AST one after the other in the search method.
  (closes #2987)


jirikuncar:
@tiborsimko can you have a look at logic in https://github.com/inveniosoftware/invenio/pull/3068/files#diff-0cb1b69656f0e417efea971fb077465f? Thanks

invenio/modules/search/api.py
```diff
@@ -22,13 +22,26 @@
...
 from invenio.base.globals import cfg
+from werkzeug.utils import cached_property, import_string
 
-from .walkers.terms import Terms
 from .walkers.match_unit import MatchUnit
+from .walkers.terms import Terms
+
+
+def query_enhancers():
+    """Return lilst of query enhancers."""
```
kaplun: typo: lilst

invenio/modules/search/enhancers/cache_results.py
```diff
@@ -0,0 +1,49 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Query results cacher."""
+
+from invenio.modules.search.cache import get_results_cache, set_results_cache
+
+
+class CacheOp(object):
+
+    """Store results in cache."""
```
kaplun: How does the cache works? For how long the results are retained? Can this be added to the docstring?

invenio/modules/search/api.py
jirikuncar: Thanks :)

invenio/modules/search/enhancers/cache_results.py
jirikuncar: For now it's using same functions `(get|set)_results_cache` like in 2.0 (https://github.com/inveniosoftware/invenio/pull/3068#discussion-diff-29278110R22).

jirikuncar:
@kaplun have you looked also at https://github.com/inveniosoftware/invenio/pull/3068/files#diff-0cb1b69656f0e417efea971fb077465f?

tiborsimko:
The restricted record logic looks good.

PR #3077, opened 2015-04-29
script: TLS encryption

crepererum:
* NEW Implements optional TLS encryption directly by Werkzeug. Adds
  many configuration variables (`SERVER_TLS_*`) to control the
  behaviour.

### DIFF
--- a/invenio/ext/script/__init__.py
+++ b/invenio/ext/script/__init__.py
@@ -46,6 +46,8 @@ import functools
 
 import re
 
+import ssl
+
 from types import FunctionType
 
 from flask import current_app, flash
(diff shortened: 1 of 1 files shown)


invenio/ext/script/__init__.py
```diff
@@ -188,6 +190,92 @@ def set_serve_static_files(sender, *args, **kwargs):
...
+                                 `ssl.PROTOCOL_*`. Defaults to `SSLv23`. See
+                                 :mod:`ssl` for details.
+    `SERVER_TLS_CIPHERS`         String that selects possible ciphers according
+                                 to the `OpenSSL cipher list format
+                                 <https://www.openssl.org/docs/apps/
+                                 ciphers.html>`_
+    `SERVER_TLS_DHPARAMS`        Filepath (string) to parameters for
+                                 Diffie-Helman key exchange. If not set the
+                                 built-in parameters are used.
+    `SERVER_TLS_ECHDCURVE`       Curve (string) that should be used for
```
jirikuncar: `ECHD` vs. `ECDH`?

invenio/ext/script/__init__.py
```diff
@@ -188,6 +190,92 @@ def set_serve_static_files(sender, *args, **kwargs):
...
+                                 Elliptic Curve-based Diffie-Helman key
+                                 exchange. If not set, the defaults provided by
+                                 OpenSSL are used.
+    ============================ ==============================================
+
+    .. note:: In case `None` is returned because of a non-enabling
+        configuration, TLS will be disabled. It is **not** possible to have a
+        TLS and non-TLS configuration at the same time. So if TLS is activated,
+        no non-TLS connection are accepted.
```
jirikuncar: :memo: nice double negation ... :smile:

jirikuncar:
:shipit: with `Reviewed-by: me` :+1:

invenio/ext/script/__init__.py
crepererum: ECDH (Elliptic Curve-based Diffie Hellman)

jirikuncar: so there is a typo `SERVER_TLS_ECHDCURVE` ;)

crepererum:
Ready. Did a last small change, that uses python to create the context with some intelligent results. This includes the deactivation of compression, which prevents us from being a victim of [CRIME](https://en.wikipedia.org/wiki/CRIME)

crepererum:
Now rebased against `master`.

dset0x:
Do the SERVER_TLS_* variables refer to the ones that will also be used by the webserver? (WRT #2645)

jirikuncar:
@dset0x you can try to re-use them where it makes sense. #2645 was stalled for long time so we have proceed without waiting.

dset0x:
In that case is `SERVER_TLS_ENABLE` somehow inferred from `CFG_SITE_SECURE_URL` and `CFG_SITE_URL` as expected in `apache.py`? How do all the other variables cooperate with the current path auto-detection mechanism that is used in [`apache.py`](https://github.com/inveniosoftware/invenio/blob/master/invenio/base/scripts/apache.py#L70)?

dset0x:
Ping @crepererum @jirikuncar

jirikuncar:
`apache.py` and `SERVER_TSL_*` are kind of independent.

crepererum:
Right now the `SERVER_TLS_*` variables are only used by this patch. You might want to extract the auto-detection (includes inference from `CFG_SITE_[_SECURE]_URL`) to a single method so your can reuse it for other parts of the project. You can also integrate the TLS config system into the Apache config generator if you like. Be aware that Python and Apache might offer different configuration options for their TLS modules and that is is kind of difficult to guess the available options, esp. for Apache. Wrong or unkown TLS parameters then might lead to an unusable Apache config.

dset0x:
@crepererum 

> You can also integrate the TLS config system into the Apache config generator if you like.

The issue is that we are not moving to such a variable for Apache yet. My question is whether it makes sense for a user to run one of the following:

* `SERVER_TLS_ENABLE = True` && `CFG_SITE{_,SECURE}_URL="http://foo"`
* `SERVER_TLS_ENABLE = False` && `CFG_SITE{_,SECURE}_URL="https://foo"`


> Be aware that Python and Apache might offer different configuration options for their TLS modules and that is is kind of difficult to guess the available options, esp. for Apache. Wrong or unkown TLS parameters then might lead to an unusable Apache config.

Indeed, this can be tricky, but if both are to use the same certificates, should both not also support eg `SERVER_TLS_KEYPASS`?

jirikuncar:
@dset0x @crepererum can you create an RFC from this discussion?

PR #3083, opened 2015-05-04
global: url_for unicode compatibility fix

hachreak:
* INCOMPATIBLE Changes url_for behaviour to return always a unicode
  string. (addresses #2967)

* INCOMPATIBLE Deprecates invenio.config hack for legacy code.
  (addresses #3106)

### DIFF
--- a/invenio/base/wrappers.py
+++ b/invenio/base/wrappers.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
-
+#
 # This file is part of Invenio.
-# Copyright (C) 2011, 2012, 2013, 2014 CERN.
+# Copyright (C) 2011, 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -62,10 +62,16 @@ same.
 
 """
 
-from functools import wraps
+from __future__ import unicode_literals
+
 from flask import Flask as FlaskBase, current_app
-from werkzeug.utils import import_string
+
+from functools import wraps
+
+from six import text_type
+
 from werkzeug.local import LocalProxy
+from werkzeug.utils import import_string
 
 from .signals import before_handle_user_exception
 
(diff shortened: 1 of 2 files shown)


invenio/base/wrappers.py
```diff
@@ -62,10 +62,14 @@
 
 """
```
jirikuncar: @hachreak we might need to import `unicode_literals` ...

invenio/base/wrappers.py
```diff
@@ -92,7 +96,7 @@ def decorator(*args, **kwargs):
         url = wrapped(*args, **kwargs)
```
jirikuncar: as `url_scheme_prefixes` should be already in `unicode` we should *unicodify* only the result of `wrapped` function.

invenio/base/wrappers.py
hachreak: :+1:

invenio/ext/confighacks.py
```diff
@@ -77,9 +79,8 @@ def __getattr__(self, name):
             elif name == '__path__':
                 return os.path.dirname(__file__)
             try:
-                from invenio.base.helpers import utf8ifier
```
jirikuncar: This hack is important for *legacy* pages. We might however raise deprecation warning for Invenio 2.3.

invenio/base/wrappers.py
```diff
@@ -90,9 +96,11 @@ def decorator(*args, **kwargs):
         url_scheme = getattr(wrapped.im_self, 'url_scheme', 'http')
         kwargs['force_external'] = False
         url = wrapped(*args, **kwargs)
+        if isinstance(url, string_types):
+            url = url.decode('utf-8')
         if force_external:
             url = url_scheme_prefixes.get(url_scheme) + url
-        return url
+        return text_type(url)
```
jirikuncar: is it still needed?

jirikuncar: ping

invenio/utils/deprecation.py
```diff
@@ -73,6 +73,11 @@ def old_method(a, b):
 """
 
 
+class RemovedInInvenio24Warning(PendingDeprecationWarning):
```
jirikuncar: IMHO It should be 2.3. (=2.1 + 0.2)

invenio/ext/confighacks.py
```diff
@@ -78,11 +87,15 @@ def __getattr__(self, name):
                 return os.path.dirname(__file__)
             try:
                 from invenio.base.helpers import utf8ifier
+                warnings.warn(
+                    "usage of invenio.config.{0} is deprecated".format(name),
```
jirikuncar: Usage ...

invenio/utils/deprecation.py
jirikuncar: you don't need this change for now ...

hachreak: ok, I'm removing this change..

PR #3084, opened 2015-05-04
global: enable Flask-IIIF extension

drjova:
Signed-off-by: Harris Tzovanakis <me@drjova.com>

### DIFF
--- a/invenio/ext/iiif/__init__.py
+++ b/invenio/ext/iiif/__init__.py
@@ -1,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Flask-IIIF extension.
+
+.. py:data:: IIIF_IMAGE_OPENER
+
+    There are two ways to initialize an IIIF image object, by ``fullpath`` and
+    ``bytestream``. Can be an ``import string`` or a ``callback function``.
+    By default ``identifier_to_path`` return the ``fullpath`` of
+    :class:`~invenio.modules.documents.api:Document`.
+
+    default: :func:`~invenio.modules.documents.utis.identifier_to_path`
+"""
+
+from flask_iiif import IIIF
+from flask_iiif.errors import MultimediaError
+
+from six import string_types
+
+from werkzeug.utils import import_string
+
+from .utils import api_file_permission_check
+
+__all__ = ('setup_app', )
+
+iiif = IIIF()
+
+
+def setup_app(app):
+    """Setup Flask-IIIF extension."""
+    if 'invenio.modules.documents' in app.config.get('PACKAGES_EXCLUDE'):
+        raise MultimediaError(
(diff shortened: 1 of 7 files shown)


invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+)
+
+__all__ = ('setup_app', )
+
+iiif = IIIF()
+
+
+def setup_app(app):
+    """Setup Flask-IIIF extension."""
+    try:
```
jirikuncar: you can just simplify it to:
```python
if ('invenio.modules.documents' not in
    app.config.get('PACKAGES_EXCLUDE'):
    raise MultimediaError(...)
```

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+        iiif.init_app(app)
+        iiif.init_restful(app.extensions['restful'])
+        iiif.uuid_to_path_handler(identifier_to_path)
+        app.config['IIIF_CACHE_HANDLER'] = 'invenio.ext.cache:cache'
+
+        # Make sure that permissions are applied on cached images as well.
+
+        @app.before_request
+        def check_cached_permissions():
+            if 'iiifimageapi' == request.endpoint:
```
jirikuncar: can you get it from `iiif` object?

invenio/ext/iiif/__init__.py
```diff
@@ -0,0 +1,69 @@
...
+                uuid = request.view_args.get('uuid')
+                try:
+                    identifier_to_path_and_permissions(uuid)
+                except (Forbidden, NotFound):
+                    raise
+                else:
```
jirikuncar: this is not needed

drjova: if the image is cached then it's not passing through the `uuid_to_path` callback it takes it directly from the cache and exposes it

invenio/modules/documents/utils.py
```diff
@@ -0,0 +1,130 @@
...
+    from invenio.modules.documents.errors import (
+        DocumentNotFound, DeletedDocument
+    )
+
+    try:
+        document = Document.get_document(uuid)
+    except (DocumentNotFound, DeletedDocument):
+        path = _simulate_file_not_found()
+    else:
+        path = document.get('uri', ''), document.is_authorized()
```
jirikuncar: Please note that this will work only for local filesystem. We should assert the opener type until you update the *Flask-IIIF* to accept bytestream.

invenio/ext/iiif/__init__.py
jirikuncar: but it would `return` anyway even without the `else` branch

drjova:
Depends on https://github.com/inveniosoftware/flask-iiif/pull/8

drjova:
Depends on  https://github.com/inveniosoftware/flask-iiif/pull/11

invenio/ext/iiif/utils.py
```diff
@@ -0,0 +1,35 @@
...
+from werkzeug.exceptions import Forbidden
+
+__all__ = ('api_file_permission_check', )
+
+
+def api_file_permission_check(*args, **kwargs):
+    """IIIF API file permission check."""
+    try:
+        identifier_to_path_and_permissions(kwargs.get('uuid', ''))
+    except Forbidden:
```
jirikuncar: I'm not sure if I understand this exception handling. Is it really needed?

drjova: :/ oups it shouldn't be here thanks!

tiborsimko:
> Depends on inveniosoftware/flask-iiif#11

Merged and released.

tiborsimko:
> * NEW Initialize the Flask-IIIF extension.

Since this will be announced in the release notes under the "global" category  (as instructed by the commit log headline), can you please expand the phrase a bit?  For example:

    * NEW Uses Flask-IIIF extension providing various image manipulation capabilities.

> * NEW Adds callback function to return the `fullpath` from a bibdoc
  file or a document based to a special syntax `recid:{recid}` or
  `recid:{recid}-{filename}` or `uuid` from document and check for
  permissions.

This description sounds a bit too technical for a "global" release notes category.  Can you please rephrase the message from the user perspective?  For example:

    * NEW  Adds possibility to refer to documents via special path such as
      `/foo/bar/uuid` (...) with proper permission checking.

drjova:
@tiborsimko I have amended the message thanks!

PR #3092, opened 2015-05-05
oauthclient: exception handling during response

### LINKED ISSUE #3029: oauthclient: Protect against exceptions when handling oauth2 responses
While testing a new oauth2 server I got an exception from the oauthclient module. The server I'm testing  fails to return the authorization token and instead returns a 500 HTTP error with XML content but without setting the content-type. The flask oauthlib assumes it's json and tries to read it, thus triggering an exception. I think it doesn't make sense to change the flask library, but Invenio should protect against this case.

TL;DR: display a nice error message when this line
https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/oauthclient/views/client.py#L182
throws an exception

jirikuncar:
* FIX Addresses problem with exception caused by malformed response
  comming from an external OAuth2 server.  (closes #3029)

### DIFF
--- a/invenio/modules/oauthclient/testsuite/test_views.py
+++ b/invenio/modules/oauthclient/testsuite/test_views.py
@@ -215,6 +215,40 @@ class RemoteAccountTestCase(OAuth2ClientTestCase):
                 )
             )
 
+    def test_invalid_authorized_response(self):
+        from simplejson import JSONDecodeError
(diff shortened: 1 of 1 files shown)


lnielsen:
See comment in #3029

lnielsen:
Looks good to me

PR #3103, opened 2015-05-06
search: proper Cache and TTL info in HTTP headers

### LINKED ISSUE #2302: search: set proper Cache and TTL information in HTTP headers
The `/search` pages use Redis as a backend to store user-selected facets.  The Redis cache for these facets expires after N minutes.  The search results pages should therefore emit proper Cache and TTL information in its HTTP headers, so that any eventual external cachers (such as varnish) could act accordingly to invalidate their caches automatically, without any configuration.

jirikuncar:
* BETTER The search results pages emits proper Cache and TTL
  information in its HTTP headers, so that any eventual external
  cachers (such as varnish) could act accordingly to invalidate their
  caches automatically, without any configuration.  (closes #2302)

Reviewed-by: Tibor Simko <tibor.simko@cern.ch>

### DIFF
--- a/invenio/modules/search/views/search.py
+++ b/invenio/modules/search/views/search.py
@@ -41,7 +41,10 @@ Template hierarchy.
 """
 
 import cStringIO
+import datetime
 import string
+import time
+
 from math import ceil
 
 from flask import (Blueprint, abort, current_app, flash, g, jsonify,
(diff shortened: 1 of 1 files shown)


invenio/modules/search/views/search.py
```diff
@@ -101,12 +105,34 @@ def checker(value):
...
+
+    if expires is None:
+        response.headers['Cache-Control'] = (
+            'no-store, no-cache, must-revalidate, '
+            'post-check=0, pre-check=0, max-age=0'
+        )
+        response.headers['Expires'] = '-1'
+    else:
+        expires_time = current_time + datetime.timedelta(seconds=expires)
+        response.headers['Cache-Control'] = 'public'
```
jirikuncar: @tiborsimko @lnielsen shall we do something special if `not current_user.is_guest()`?

lnielsen: I would assume so. Part of it can be controlled with a "vary on".

Does it already sent a vary on for the language?

Issue #3089, opened 2015-05-05
legacy: run_sql problems with quotes


jalavik:
After 79741888d8e71b22e704a30038f9857d90a38590 there are several problems with running programs such as bibsched:

```shell
$ bibsched
...
File "/home/jlavik/envs/pu/src/invenio/invenio/legacy/dbquery_mysql.py", line 277, in run_sql
    except (OperationalError, InterfaceError):
NameError: global name 'OperationalError' is not defined
```
PS: Perhaps pylint -E could have catched this initial error?

I fixed the import, but for bibsched troubles were not over:

```
  File "/home/jlavik/envs/pu/src/invenio/invenio/legacy/bibsched/cli.py", line 378, in fetch_debug_mode
    r = run_sql('SELECT value FROM schSTATUS WHERE name = "debug_mode"')
  File "/home/jlavik/envs/pu/src/invenio/invenio/legacy/dbquery_mysql.py", line 277, in run_sql
    rc = cur.execute(sql, param)
  File "/home/jlavik/envs/pu/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
  File "/home/jlavik/envs/pu/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
_mysql_exceptions.OperationalError: (1054, "Unknown column 'debug_mode' in 'where clause'")
```

I tried fixing these, but its simply too much. Probably due to ansi mode quotes? I suggest in the meantime to revert 79741888d8e71b22e704a30038f9857d90a38590

@hachreak @jirikuncar

hachreak:
Sorry, I replaced import in any files but I forgot to leave necessary in dbquery... T_T
I'll fix it immediately! :)

hachreak:
for the `ansi_quote`: I think it's necessary to be compatible with PostgreSQL. T_T
Tomorrow I'll work on replace `"` with `'` in raw query on legacy code to solve the 2° problem...

jalavik:
BTW, visiting the `/admin/` page gives the following error:

```
[...]

PR #3104, opened 2015-05-06
access: fix model upgrade

hachreak:
* Adds a recipe for upgrade the model. (addresses #3089)

### DIFF
--- a/invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
+++ b/invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Upgrade recipe."""
+
+import warnings
+
+from invenio.ext.sqlalchemy import db
+from invenio.legacy.dbquery import run_sql
+from invenio.modules.upgrader.api import op
+
+from sqlalchemy.engine import reflection
+
+depends_on = ['invenio_release_1_2_0']
+
+
+def exists_id_column():
+    """Check if id column already exists."""
+    insp = reflection.Inspector.from_engine(db.engine)
+    columns = insp.get_columns('accROLE_accACTION_accARGUMENT')
+
+    return any([column['name'] == 'id' for column in columns])
+
+
+def info():
+    """Info."""
+    return "Add new column id to accROLE_accACTION_accARGUMENT table."""
+
+
+def do_upgrade():
+    """Implement your upgrades here."""
+    if exists_id_column():
+        warnings.warn(
+            """Upgrade skipped. """
+            """Column 'id' already exists on accROLE_accACTION_accARGUMENT.""")
+        return
+    if op.impl.dialect.name != 'mysql':
+        warnings.warn("""This upgrade supports only MySQL.""")
+        return
+
+    # table accROLE_accACTION_accARGUMENT
+
+    # - drop primary key
+    # - add "id" column int(15) unsigned
+    # - set "id" as primary key, autoincrement
+    # - column id_accROLE, id_accACTION, id_accARGUMENT, argumentlistid server
+    #   default = None
+    op.execute(
+        """
+        SET SESSION sql_mode = ANSI_QUOTES;
+        ALTER TABLE "accROLE_accACTION_accARGUMENT"
+        CHANGE COLUMN "id_accROLE" "id_accROLE" INT(15) UNSIGNED NULL ,
+        CHANGE COLUMN "id_accACTION" "id_accACTION" INT(15) UNSIGNED NULL ,
+        CHANGE COLUMN "id_accARGUMENT" "id_accARGUMENT" INT(15) NULL ,
+        CHANGE COLUMN "argumentlistid" "argumentlistid" MEDIUMINT(8) NULL ,
+        ADD COLUMN "id" INT(15) UNSIGNED NOT NULL AUTO_INCREMENT,
+        DROP PRIMARY KEY,
+        ADD PRIMARY KEY ("id");
+        """)
+
+
+def estimate():
+    """Estimate running time of upgrade in seconds (optional)."""
+    total = run_sql(
+        """SELECT count(*) FROM "accROLE_accACTION_accARGUMENT" """)
+    return int(float(int(total[0][0])) / 1000) + 1
+
+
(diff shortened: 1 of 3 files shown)


invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
```diff
@@ -0,0 +1,103 @@
...
+from invenio.modules.upgrader.api import op
+
+from sqlalchemy.exc import OperationalError
+
+depends_on = ['invenio_release_1_2_0']
+
+
+def info():
+    """Info."""
+    return "Add a autoincrement id in accROLE_accACTION_accARGUMENT table"""
```
jirikuncar: What about: `Add new column id to accROLE_accACTION_accARGUMENT table.` ?

invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
```diff
@@ -0,0 +1,103 @@
...
+        run_sql("""UPDATE accROLE_accACTION_accARGUMENT
+                   SET id = %s
+                   WHERE id_accROLE = %s AND
+                         id_accACTION = %s AND
+                         id_accARGUMENT = %s AND
+                         argumentlistid = %s """,
+                (index + 1, rec[0], rec[1], rec[2], rec[3]))
+    # remove primary key
+    try:
+        op.drop_constraint(None, 'accROLE_accACTION_accARGUMENT',
```
jirikuncar: I would start by dropping the constraint and then adding new `AUTO_INCREMENT` column.

> When you add an AUTO_INCREMENT column, column values are filled in with sequence numbers automatically. For MyISAM tables, you can set the first sequence number by executing SET INSERT_ID=value before ALTER TABLE or by using the AUTO_INCREMENT=value table option. 

https://dev.mysql.com/doc/refman/5.5/en/alter-table-examples.html

There must an easy way for PostgreSQL too, but you can fallback to the current solution ...

hachreak: if you do:
```python
op.drop_constraint(None, 'accROLE_accACTION_accARGUMENT',type_='primary')
op.add_column('accROLE_accACTION_accARGUMENT', db.Column('id', db.Integer(15, unsigned=True), nullable=False, autoincrement=True))
```
Mysql create a id column by initialized as zero.

p.s. If I try (after drop the constraint) to create the id column also as primary key:

```python
op.add_column('accROLE_accACTION_accARGUMENT', db.Column('id', db.Integer(15, unsigned=True), nullable=False, autoincrement=True, primary_key=True))
```

Mysql returns error:
```
OperationalError: (_mysql_exceptions.OperationalError) (1075, 'Incorrect table definition; there can be only one auto column and it must be defined as a key') [SQL: u'ALTER TABLE `accROLE_accACTION_accARGUMENT` ADD COLUMN id INTEGER(15) UNSIGNED NOT NULL AUTO_INCREMENT']
```

hachreak: execute manually sql, mysql automatically set id:

```sql
CREATE TABLE `invenio`.`test_increment` (
  `foo` INT NOT NULL AUTO_INCREMENT,
  `test` VARCHAR(45) NULL,
  PRIMARY KEY (`foo`));

INSERT INTO `invenio`.`test_increment` (`test`) VALUES ('one');
INSERT INTO `invenio`.`test_increment` (`test`) VALUES ('two');
INSERT INTO `invenio`.`test_increment` (`test`) VALUES ('three');
INSERT INTO `invenio`.`test_increment` (`test`) VALUES ('four');

ALTER TABLE `invenio`.`test_increment` 
DROP COLUMN `foo`,
ADD COLUMN `id` INT NOT NULL AUTO_INCREMENT AFTER `test`,
DROP PRIMARY KEY,
ADD PRIMARY KEY (`id`);
```

hachreak: @jirikuncar I'll use `run_sql` and run this `ALTER TABLE` query instead of use alembic? :)

jirikuncar:
```python
op.execute('ALTER TABLE "test_increment" ADD COLUMN "id" INT NOT NULL AUTO_INCREMENT AFTER "test", ADD PRIMARY KEY ("id");')
```
works for me

invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def info():
+    """Info."""
+    return "Add new column id in accROLE_accACTION_accARGUMENT table."""
+
+
+def do_upgrade():
+    """Implement your upgrades here."""
+    # table accROLE_accACTION_accARGUMENT
```
jirikuncar: please check if you already have `id` column.

hachreak: exists a "standard" way to check? I think i need to create a mysql specific query

jirikuncar: try http://docs.sqlalchemy.org/en/latest/core/reflection.html and `hasattr(table.c, 'id')`

hachreak: it should reflect the database and not the code.
can I load SQLAlchemy table class from the database?

invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
```diff
@@ -0,0 +1,75 @@
...
+    op.execute(
+        """
+        %s
+        ALTER TABLE "accROLE_accACTION_accARGUMENT"
+        CHANGE COLUMN "id_accROLE" "id_accROLE" INT(15) UNSIGNED NULL ,
+        CHANGE COLUMN "id_accACTION" "id_accACTION" INT(15) UNSIGNED NULL ,
+        CHANGE COLUMN "id_accARGUMENT" "id_accARGUMENT" INT(15) NULL ,
+        CHANGE COLUMN "argumentlistid" "argumentlistid" MEDIUMINT(8) NULL ,
+        ADD COLUMN "id" INT(15) UNSIGNED NOT NULL
+            AUTO_INCREMENT,
```
jirikuncar: lines 54 and 55 can be joined ...

invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
jirikuncar: yes, reflect db not python model.

hachreak:
ping @jirikuncar

invenio/modules/access/upgrades/access_2015_05_06_accROLE_accACTION_accARGUMENT_id.py
```diff
@@ -0,0 +1,95 @@
...
+    # - drop primary key
+    # - add "id" column int(15) unsigned
+    # - set "id" as primary key, autoincrement
+    # - column id_accROLE, id_accACTION, id_accARGUMENT, argumentlistid server
+    #   default = None
+    op.execute(
+        """
+        {0}
+        ALTER TABLE "accROLE_accACTION_accARGUMENT"
+        CHANGE COLUMN "id_accROLE" "id_accROLE" INT(15) UNSIGNED NULL ,
```
jirikuncar: these 4 lines can be done with alembic directly, can't they?

PR #3121, opened 2015-05-12
Personalization fixes and improvements

kasioumis:


egabancho:
@kasioumis can you rebase it against latest legacy?

PR #3122, opened 2015-05-12
oauthclient: validate oauth users' email

### LINKED ISSUE #2739: oauthclient: validate email address on user sign-up
For oauth services where the email address is not provided upon first sign up (such as ORCID) we currently show a sign-up form with an input for the email address but that email address is not being validated.

The email address should be validated but ideally the flow of the user wanting, for example, to submit should not be blocked until he/she validates email address.

So the proposal would be to do something similar to what https://balsamiq.com/ is doing. Upon sign-up you can start using the app right away but the validation email states:

> To start your trial, you need to verify your email address. Please go to the link below and log in:

> https://inspirehep.mybalsamiq.com/confirm/ac62ef1a8ce02277c901b61d86690e7fb3d1e1fb

> If you don't confirm within the next 2 days, your site will be suspended. If you don't confirm within a month, the site will be deleted.

> Almost there!

So we could allow for some days to access without validation and after that, block the account.

nharraud:
* FIX users creating their account with oauth process have to validate
  their email too. Previously it was marked as valid from the start.

* FIX a validation email is now sent when users create their account
  with oauth and their email address is not provided by the oauth
  service. (closes #2739)

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/invenio/modules/oauthclient/handlers.py
+++ b/invenio/modules/oauthclient/handlers.py
@@ -26,10 +26,12 @@ from functools import partial, wraps
 from flask import current_app, flash, redirect, render_template, \
     request, session, url_for
 from flask_login import current_user
-import six
-from werkzeug.utils import import_string
 
 from invenio.base.globals import cfg
+
+import six
+
+from werkzeug.utils import import_string
 
 from .client import oauth, signup_handlers
 from .errors import OAuthClientError, OAuthError, \
--- a/invenio/modules/oauthclient/testsuite/test_contrib_orcid.py
+++ b/invenio/modules/oauthclient/testsuite/test_contrib_orcid.py
@@ -23,6 +23,8 @@ from __future__ import absolute_import
 
 from flask import session, url_for
 
+from flask_email.backends import locmem
+
 import httpretty
 
 from invenio.ext.sqlalchemy import db
@@ -66,6 +68,8 @@ class OrcidTestCase(OAuth2ClientTestCase):
                 consumer_key='changeme',
                 consumer_secret='changeme',
             ),
+            # use local memory mailbox
+            EMAIL_BACKEND='flask_email.backends.locmem.Mail',
         ))
         return app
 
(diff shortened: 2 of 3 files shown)


jmartinm:
@nharraud thanks for the update. On http://labs.inspirehep.net we currently only use the ORCID login so this will affect us directly.

- Can you clarify what ``CFG_ACCESS_CONTROL_NOTIFY_USER_ABOUT_NEW_ACCOUNT`` does?

Once the user clicks on the validation link on his email, does it get directed to the page he/she was trying to visit in the first place? This is important to minimally disrupt what she was trying to do in the first place when the login page was encountered.

lnielsen:
> Can you clarify what CFG_ACCESS_CONTROL_NOTIFY_USER_ABOUT_NEW_ACCOUNT does?

This is the old variable which is currently to control if a user receives email validation message upon registration - see https://github.com/inveniosoftware/invenio/blob/master/invenio/legacy/webuser.py#L490

nharraud:
@jmartinm This modification doesn't change the way email validation works. The user can use his account right away without clicking the link in the email. There is just a flash message informing the user that he should validate it.

jmartinm:
Nice. So if ``CFG_ACCESS_CONTROL_NOTIFY_USER_ABOUT_NEW_ACCOUNT == False`` there would be absolutely no change in the way the login workflow works at the moment right?

Is there any consequence on not validating the email address?

It could be nice to allow the user to continue straight away, but say, after a couple of days if the email has not been validated they would be forced to validate to continue on a page that needs login.

That way we don't interrupt the person that wants to get something done the first time and give them some time to validate but we get the email validated at some point.

Does it make sense?

nharraud:
@jmartinm `CFG_ACCESS_CONTROL_NOTIFY_USER_ABOUT_NEW_ACCOUNT == False` would just prevent the email to be sent after registration.
There is always a popup on the website informing the user that he has to validate his email address.
Right now the user can do any action he wants without validating it. Later on we plan to add some limitations like, for example, enabling the user to create a deposition but forbidding him from submitting it.
As you say, we don't want to interrupt the website usage just after the registration.

jmartinm:
Thanks for clarification @nharraud 

Just had a look in Zenodo to see how the `Verify your email` works. Looks good to me.

jirikuncar:
@nharraud it looks good to me. @lnielsen have you reviewed it?

lnielsen:
Yes. :shipit:

PR #3128, opened 2015-05-13
global: pagination compontent enhancement

bouzlibop:
* Makes diabled links non-clickable.

### DIFF
--- a/invenio/base/templates/paginate_base.html
+++ b/invenio/base/templates/paginate_base.html
@@ -39,11 +39,19 @@ It assumes that page exists in the current context.
     <li {% if not obj.has_prev -%} class="disabled" {%- endif %}}>
       {%- set new_args = args.copy() -%}
       {%- do new_args.update({'page': 1}) -%}
+      {%- if not obj.has_prev %}
+      <span title="first">&laquo;</span>
+      {% else %}
       <a title="first" href="{{ url_for(endpoint, **new_args) if obj.has_prev}}">&laquo;</a>
+      {%- endif %}
     </li>
     <li {% if not obj.has_prev -%} class="disabled" {%- endif %}}>
       {%- do args.update({'page': obj.page-1 }) -%}
+      {%- if not obj.has_prev %}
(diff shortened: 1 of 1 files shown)


invenio/base/templates/_formhelpers_base.html
```diff
@@ -144,3 +146,17 @@
...
 {% endblock %}{% endmacro %}
+
+{% macro render_radio_field(field) -%}
+<div class="form-group">
+  <label class="control-label" for="name">{{ field.label }}</label>
+  {% for value, label, checked in field.iter_choices() %}
+  <div class="radio">
+    <label>
+      <input type="radio" name="{{ field.id }}" id="{{ field.id }}" value="{{ value }}" {{ "checked" if checked }}>{{ label }}
+      <small class="text-muted">{{ field.description[value] }}</small>
```
bouzlibop: @lnielsen, can you take a look at it?
Here I used a little trick to get description per bullet.
Here is how I'm using it: https://github.com/inveniosoftware/invenio-groups/pull/1/files#diff-53e9e6c85499508d2e7be4767059f1daR149
I'm not sure if this could be done any better?

lnielsen: I think it's better that we create a new widget which is capable of rendering a description text. See e.g. https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/deposit/field_widgets.py#L598

invenio/base/templates/paginate_base.html
```diff
@@ -39,11 +39,19 @@
     <li {% if not obj.has_prev -%} class="disabled" {%- endif %}}>
       {%- set new_args = args.copy() -%}
       {%- do new_args.update({'page': 1}) -%}
+      {%- if not obj.has_prev %}
+      <span title="first" href="{{ url_for(endpoint, **new_args) if obj.has_prev}}">&laquo;</span>
```
lnielsen: I don't think the ``href`` should be in the ``span` tag?

PR #3129, opened 2015-05-13
records: revision API

egabancho:


invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
```diff
@@ -0,0 +1,70 @@
...
+    if not op.has_table("record_json_history"):
+        op.create_table(
+            'history_record_json',
+            sa.Column(
+                'id', db.MediumInteger(8, unsigned=True), nullable=False),
+            sa.Column('revision', db.DateTime, nullable=False),
+            sa.Column('json', db.JSON, nullable=False),
+            sa.PrimaryKeyConstraint('id', 'revision'),
+            mysql_charset='utf8',
+            mysql_engine='MyISAM'
```
jmartinm: Any advantage of using MyISAM over InnoDB?

invenio/modules/records/models.py
```diff
@@ -159,6 +159,31 @@ class RecordMetadata(db.Model):
...
+    """Represents the revisions of a record if any."""
+
+    __tablename__ = 'record_json_history'
+
+    id = db.Column(
+        db.MediumInteger(8, unsigned=True),
+        db.ForeignKey(Record.id),
+        primary_key=True,
+        nullable=False,
+        autoincrement=True
```
jmartinm: I think `autoincrement` is `True` by default? http://docs.sqlalchemy.org/en/improve_toc/dialects/mysql.html#auto-increment-behavior

Different question is, do we want to have autoincrement in this column? Or if it should contain a `recid` autoincrementing is not desirable? Just thinking out loud..

egabancho: You are right!, it shouldn't be `autoincrement`

invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
egabancho: We can discuss about MyISAM vs InnoDB, but I think the point here is that I copy paste from other upgrade recipe, see `git grep 'op\.create_table'`

I'm not sure whether is necessary or not to specify here the engine, anyhow this is the one used by [default](https://github.com/inveniosoftware/invenio/blob/master/invenio/ext/sqlalchemy/__init__.py#L142) in Invenio.

cc @lnielsen

lnielsen: Currently Invenio does not support InnoDB AFAIK, hence it should be MyISAM. InnoDB support comes together with PostgreSQL support (again AFAIK).

invenio/modules/records/api.py
```diff
@@ -58,11 +60,30 @@ def create_many(cls, blobs, master_format, **kwargs):
...
     @classmethod
-    def get_record(cls, recid, reset_cache=False):
+    def get_record(cls, recid, reset_cache=False, revision=None):
         """Get one record from the DB.
 
         :param reset_cache: If set to `True` it creates the JSON again.
+        :param revision: If not None and revision exists it tries to retrieve
+            it from `history_record_json`.
         """
+        from invenio.ext.sqlalchemy import db
```
lnielsen: Does the import not work in the top?

lnielsen: Ah, sorry just saw that you just moved an existing import. Anyway, perhaps it still works if the import is in the top.

invenio/modules/records/api.py
```diff
@@ -58,11 +60,30 @@ def create_many(cls, blobs, master_format, **kwargs):
...
+        if revision is not None:
+            try:
+                timestamp = datetime.datetime.strptime(
+                    revision[:14], '%Y%m%d%H%M%S')
+            except:
```
lnielsen: ``except ValueError:`` ?

invenio/modules/records/api.py
```diff
@@ -58,11 +60,30 @@ def create_many(cls, blobs, master_format, **kwargs):
...
+                raise RecordError('Wrong revision: {0}'.format(revision))
+            try:
```
lnielsen: Perhaps collapse the two try/except clauses into one?

invenio/modules/records/models.py
```diff
@@ -159,6 +159,30 @@ class RecordMetadata(db.Model):
     record = db.relationship(Record, backref='record_json')
 
 
+class RecordMetadataHistory(db.Model):
```
lnielsen: What about just ``RecordHistory `` instead of ``RecordMetadataHistory``?

invenio/modules/records/models.py
```diff
@@ -159,6 +159,30 @@ class RecordMetadata(db.Model):
...
+
+    revision = db.Column(
+        db.DateTime,
+        primary_key=True,
+        nullable=False
+    )
+
+    json = db.Column(db.JSON, nullable=False)
+
+    record = db.relationship(Record, backref='record_json_history')
```
lnielsen: If we go with ``RecordHistory`` then perhaps we can call the back reference just ``history``?

invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
```diff
@@ -0,0 +1,70 @@
...
+
+def do_upgrade():
+    """Create new table for the record history.
+
+    To fill it up please use:
+    `python scripts/history_record_migration_kit.py run`.
+    """
+    if not op.has_table("record_json_history"):
+        op.create_table(
+            'history_record_json',
```
lnielsen: ``history_record_json`` -> ``record_json_history``

invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
```diff
@@ -0,0 +1,70 @@
...
+                'id', db.MediumInteger(8, unsigned=True), nullable=False),
+            sa.Column('revision', db.DateTime, nullable=False),
+            sa.Column('json', db.JSON, nullable=False),
+            sa.PrimaryKeyConstraint('id', 'revision'),
+            mysql_charset='utf8',
+            mysql_engine='MyISAM'
+        )
+    else:
+        warnings.warn(
+            "*** Creation of 'history_record_json' table skipped! ***")
```
lnielsen: ditto

invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def estimate():
+    """Estimate running time of upgrade in seconds (optional)."""
+    return 1
+
+
+def pre_upgrade():
+    """Run pre-upgrade checks (optional)."""
+    table = "history_record_json"
```
lnielsen: ditto

scripts/history_record_migration_kit.py
```diff
@@ -0,0 +1,65 @@
...
+    from invenio.modules.records.models import Record, HistoryRecordMetadata
+    from invenio.legacy.bibedit.utils import \
+        get_marcxml_of_revision, get_record_revisions
+
+    for recid in db.session.query(Record.id).yield_per(1000):
+        recid = recid[0]
+        for dummy, revision in get_record_revisions(recid):
+            marc_xml = get_marcxml_of_revision(recid, revision)
+            record = create_record(marc_xml, 'marc', schema='xml')
+            hst_record = HistoryRecordMetadata(
```
lnielsen: ``HistoryRecordMetadata`` -> ``RecordMetadataHistory`` (or whichever name we decide upon).

invenio/modules/records/upgrades/records_2015_05_06_recjson_history_table.py
kaplun: ? It's not forbidden to mix up InnoDB and MyISAM tables. E.g. in INSPIRE we have ported several tables to InnoDB to avoid table locking. The only issue is whether this specific table requires functionalities that are available only in MyISAM, which should not be the case for any new table, right? Especially since the final goal is to be ANSI and hence PostgreSQL compatible.

The basic limitation of InnoDB WRT MyISAM is the missing support for fulltext search, which should not be needed in this context and which is now implemented as of MySQL 5.6.4 :smile: 
https://dev.mysql.com/doc/refman/5.6/en/innodb-fulltext-index.html

On the other hand, since here we are in the context of an upgrade recipe, which is used to migrate live instances of Invenio, since we always have recommended a certain set of optimization to the MySQL configuration which were ignoring completely the usage of InnoDB, I believe it might make sense, in the upgrade recipe to use MyISAM.

We should though create an RFC about our recommended database configuration.

tiborsimko: I think @lnielsen may have alluded to the fact that we had agreed in the past that instead of moving to InnoDB engine (and updating old SQL client code accordingly to take advantage of transactions etc), we simply aim at moving fully to PostgreSQL (and SQLAlchemy) in general.  From this point of view, while it would be very well possible to mix MyISAM and InnoDB engines, it does not seem advantageous to start introducing InnoDB tables now, in view of this longer goal (that is progressing well).  

(Side advantage: regular Invenio clients can keep using the same DB setup with ISAM-driven DB configuration for optimisation and backups etc.)

jirikuncar:
Is revision identifier resolution in seconds enough? Shall we include milliseconds (and/or record sha1)?

egabancho:
So far the milliseconds were not needed because BibUpload works at the level of seconds.
But now that you mentioned it, if we are going to allow concurrent upload/modify tasks, this is something we should be looking at.

About sha1, I think it's a nice way of naming versions, should we start from now on (meaning whenever we use the new history table) or should we transform all of them?

There is a thing which comes to the top of my head now, what will happen with the modification date as it is now? should we move it to a different field? is it a calculated field calculated from the history table?

egabancho:
see inveniosoftware/invenio-records#10

PR #3130, opened 2015-05-13
global: groups module externalization

bouzlibop:


lnielsen:
You might need to add ``invenio-groups`` in ``setup.py``

lnielsen:
Merge needs to be aligned with pypi release of invenio-groups.

invenio/base/config.py
```diff
@@ -65,6 +65,7 @@
 PACKAGES = [
     'invenio.modules.*',
     'invenio.base',
+    'invenio_groups',
```
bouzlibop: Shall we by default list `invenio_groups` in `PACKAGES`?

lnielsen: Yes, I think so. As long as core invenio modules depends on groups we need to have it installed by default.

jirikuncar: later on we can use setup entry points.

invenio/modules/baskets/upgrades/baskets_2015_05_21_groups_module_externalization.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            op.rename_table(
+                old_table_name='usergroup_bskBASKET',
+                new_table_name='group_bskBASKET',
+            )
+            op.alter_column(
+                table_name='group_bskBASKET', column_name='id_usergroup',
+                new_column_name='id_group',
+                existing_type=db.Integer(15, unsigned=True),
+                existing_nullable=False, existing_server_default='0',
```
jirikuncar: shall we allow null values and add foreign key?

bouzlibop: I took what there already was in [invenio/modules/baskets/models.py](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/baskets/models.py#L160-162) and added `existing_*` according to [this](http://alembic.readthedocs.org/en/latest/ops.html#alembic.operations.Operations.alter_column.params.existing_server_default). Do we want to change this original behaviour along with the table name change?

jirikuncar: If we are doing upgrade recipe I would say we should make it "future" Postgresql friendly ;)

jirikuncar:
@bouzlibop what is the status of *Invenio-Groups* package and this PR?

bouzlibop:
@jirikuncar groups package looks pretty ready to me, can you take a look at this PR?

bouzlibop:
After disabling this one:

`invenio/base/config.py`
```diff
     'invenio.ext.jasmine',  # after assets
     'flask_breadcrumbs:Breadcrumbs',
     'invenio.modules.deposit.url_converters',
-    'invenio.ext.iiif',
+    # TODO 'invenio.ext.iiif',
     'invenio.ext.es',
 ]
 
 PACKAGES = [
     'invenio_records',
     'invenio_comments',
+    # TODO 'invenio_comments',
     # TODO 'invenio_documents',
     'invenio_pidstore',
     'invenio.modules.*',
     'invenio.base',
+    'invenio_groups',
 ]
 
 PACKAGES_EXCLUDE = [
 @@ -75,6 +77,7 @@
     'invenio.modules.comments',
     'invenio.modules.communities',  # remove with invenio/modules/communities
     'invenio.modules.documents',
+    'invenio.modules.groups',
     'invenio.modules.multimedia',
     'invenio.modules.pages',
     'invenio.modules.pidstore',
```

and removing the version from `invenio-groups` in `setup.py`

I have the same result of running the tests with this PR as in running tests master - only 3 errors with msg `No module named invenio_documents.api`

Issue #3042, opened 2015-04-21
scripts: manager database dump not implemented


blixhavn:
https://github.com/inveniosoftware/invenio/blob/master/invenio/base/scripts/database.py#L220

This function should at least contain a FIXME, or cast a not_yet_implemented exception.

jirikuncar:
### Proposal

Investigate usage of an existing tool e.g. https://github.com/cuducos/alchemydumps.

tiborsimko:
Superseded by Invenio 3.

PR #3131, opened 2015-05-15
global: manager command clean-up

lnielsen:
* FIX Removes not implemented database dump command. (addresses #3042)


egabancho:
I think this one, or at least the signal that it generates, is used by the mixer to create the JSON files, see https://github.com/inveniosoftware/invenio/blob/master/invenio/ext/mixer/__init__.py#L227

lnielsen:
Ah, got it. It looks very suspicious that there are now implementation in the method :-) I'll close the PR then.

PR #3133, opened 2015-05-15
Invenio v2.0.3

tiborsimko:
- bbb1ad3983b7fd44e4e6dd58b0e121bdd087e26c Invenio v2.0.3
- 04a3a815b8b4f228c2476de58e0f75f3724ffb9d matcher: fix argument types in jaro_fctn
- 7a122dab6847b1823d3248d2856cc06dd5e2496a legacy: lazy stopwords knowledge base


lnielsen:
:shipit: (``legacy`` is not recognised as component name, but not sure what to use instead).

tiborsimko:
7a122da just amends the commit message of c6478755110d96229bb4726dd423291efb822484 so that the change would appear in the generated release notes.  I have kept the original commit log headline to make the link obvious.  (hence "legacy" as the leading component name in the commit log headline)

PR #3136, opened 2015-05-18
workflows: Holding Pen list view

jalavik:
New Holding Pen list view based on small flightJS components instead of DataTables.

PR is ready for integration.


invenio/modules/workflows/utils.py
```diff
@@ -175,29 +177,20 @@ def _sorter(item):
...
-        bwolist.sort(key=lambda x: x.data_type, reverse=should_we_reverse)
-    elif iSortCol_0 == 7:
-        bwolist.sort(key=lambda x: x.version, reverse=should_we_reverse)
-    elif iSortCol_0 == 8:
-        bwolist.sort(key=lambda x: x.version, reverse=should_we_reverse)
+def sort_bwolist(bwolist, sort_key):
+    """Sort a list of workflow objects for the list."""
+    if sort_key == "newest":
+        bwolist.sort(key=attrgetter("created"), reverse=True)
+    if sort_key == "oldest":
```
dset0x: This looks like it should be an `elif`.

jalavik: Oh, indeed. Thanks.

invenio/modules/oaiharvester/config.py
```diff
@@ -21,6 +21,17 @@
...
+import os
+import sys
+
+
 OAIHARVESTER_DEFAULT_NAMESPACE_MAP = {
     None: "http://www.openarchives.org/OAI/2.0/",
 }
+"""The default namespace used when handling OAI-PMH results."""
+
+OAIHARVESTER_STORAGEDIR = os.path.join(sys.prefix, "var/data/oaiharvester/storage")
```
dset0x: We now have [CFG_DATADIR](https://github.com/inveniosoftware/invenio/blob/master/invenio/base/config.py#L93). It would also be best not to hardcode the slashes:

```python
OAIHARVESTER_STORAGEDIR = os.path.join(cfg['CFG_DATADIR'], 'oaiharvester', 'storage')
```

Naturally, `oaiharvester` could declare a base directory based on `CFG_DATADIR`, say,

```python
OAIHARVESTER_BASEDIR = os.path.join(cfg['CFG_DATADIR'], 'oaiharvester')
OAIHARVESTER_STORAGEDIR = os.path.join(OAIHARVESTER_BASEDIR, 'storage')
```

invenio/modules/oaiharvester/tasks/postprocess.py
```diff
@@ -297,53 +243,49 @@ def refextract(obj, eng):
...
-            obj.add_task_result("References",
-                                new_dict_representation['reference'],
-                                "workflows/results/refextract.html")
+            obj.data["reference"] = new_dict_representation["reference"]
+            obj.log.info("Extracted {0} references".format(len(obj.data["reference"])))
+            obj.update_task_results(
+                "References",
+                [{"name": "References",
+                  "result": new_dict_representation['reference'],
+                  "template": "workflows/results/refextract.html"}]
```
dset0x: Perhaps room for another call to `os.path.join` here.

invenio/utils/plotextractor/api.py
```diff
@@ -0,0 +1,110 @@
...
+    dummy, image_list, tex_files = untar(
+        tarball,
+        sub_dir
+    )
+
+    converted_image_list = convert_images(image_list)
+    extracted_image_data = []
+    if tex_files == [] or tex_files is None:
+        # Its not a tarball
+        run_shell_command('rm -r %s', (sub_dir,))
```
dset0x:
```python
import shutil
shutil.rmtree(sub_dir)
```
should give the advantage of exceptions being raised at the cost of speed.

invenio/utils/plotextractor/api.py
```diff
@@ -0,0 +1,110 @@
...
+        os.makedirs(save_to_folder)
+    tarball, dummy = harvest_single(
+        arxiv_id,
+        save_to_folder,
+        ["tarball"]
+    )
+    return tarball
+
+
+def get_pdf_from_arxiv(arxiv_id, save_to_folder):
```
dset0x: I would see to merge this and the above function. It looks like the only new required parameter would be `pdf/tarball` (at the cost of accidentally passing a string that is neither of these of course!).

invenio/modules/oaiharvester/config.py
jalavik: :+1:

invenio/modules/oaiharvester/tasks/postprocess.py
jalavik: We could, I guess. However, throughout Invenio 2 referencing jinja templates have usually been written this way everywhere. We should then change it everywhere which is probably out of scope for this update.

jirikuncar:
As I mentioned in https://github.com/inveniosoftware/invenio/pull/2628#issuecomment-130555086, please move it to *Invenio-Workflows* package. I can help you with conflict resolution in `*.py` files and the rest should be mergeable without major conflicts.

PR #3141, opened 2015-05-19
BibAuthorID: safer URL redirecting

tiborsimko:
* SECURITY Improves URL redirecting by properly quoting all URL parts,
  in order to better protect against possible XSS attacks.

Reviewed-by: Samuele Kaplun <samuele.kaplun@cern.ch>

### DIFF
--- a/modules/bibauthorid/lib/bibauthorid_webinterface.py
+++ b/modules/bibauthorid/lib/bibauthorid_webinterface.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2011, 2012 CERN.
+# Copyright (C) 2011, 2012, 2015 CERN.
 #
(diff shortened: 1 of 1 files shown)


modules/bibauthorid/lib/bibauthorid_webinterface.py
```diff
@@ -3724,7 +3726,7 @@ def __call__(self, req, form):
             try:
                 pid = int(self.path)
             except ValueError:
-                redirect_to_url(req, "%s/author/search?q=%s" % (CFG_BASE_URL, self.path))
+                redirect_to_url(req, "%s/author/search?q=%s" % (CFG_BASE_URL, urllib.quote(self.path)))
```
kaplun: I have double checked this one and `self.path` is populated internally by BibAuthorID and is always indeed a string or `None` but in this latter case it is captured earlier in the function.

kaplun:
Good for me! :shipit:

PR #3144, opened 2015-05-19
global: deprecation warnings logging and new config options

### LINKED ISSUE #3138: general: legacy config variables deprecation message improvement
Since the update to latest ``master`` the following errors appear many times in the development console.

```python
/Users/jmartinm/.virtualenvs/invenio-inspire/lib/python2.7/site-packages/werkzeug/local.py:338: RemovedInInvenio22Warning: Use a more specific variable from invenio/base/config.py instead
  return getattr(self._get_current_object(), name)
```

Problem here is that no reference to the actual problem is made, only a pointer to a line in the ``werkzeug`` library. This message should be improved if the aim is for the developer to make some fixes before Invenio 2.2 is released.

dset0x:
* BETTER Makes the `deprecated` decorator write a full stacktrace to the
  main log to cater for indirect invocations. (closes #3138)

Reported-by: Javier Martin Montull <javier.martin.montull@cern.ch>


invenio/utils/deprecation.py
```diff
@@ -87,17 +89,49 @@ class RemovedInInvenio22Warning(DeprecationWarning):
...
+    # Debug
+    try:
+        current_app.logger.debug(formatted_warning)
+        current_app.logger.debug(''.join(traceback.format_stack()))
+    except RuntimeError:
+        # App not yet initialized; see no evil, hear no evil.
+        pass
+
+
+original_showwarning = warnings.showwarning
```
crepererum: Should this backup better made in line 132 instead? Otherwise the original state is not restored when `warnings.showwarning` is changed somewhere else after 112 is executed and therefore the decorator has unwanted side effects.

dset0x: @crepererum You're right, I hadn't thought about it being changed elsewhere in the code; thanks for going through the code.

jirikuncar:
@dset0x please ask @jmartinm for review.

--
PS: `Reported-by` is usually used for mentioning people outside the developers team.

invenio/utils/deprecation.py
```diff
@@ -87,17 +89,49 @@ class RemovedInInvenio22Warning(DeprecationWarning):
...
+def warn_with_traceback(message, category, filename, lineno, file=None,
+                        line=None):
+    """Log deprecation warnings to debug logger."""
+    # Format
+    message.args += 'Full backtrace in the logfile if DEBUG is enabled',
+    formatted_warning = warnings.formatwarning(message, category, filename,
+                                               lineno, line)
+
+    # Warn
+    log = file if hasattr(file, 'write') else sys.stderr
```
jirikuncar: IMHO we should check that `if file is not None` otherwise "`if not hasattr(file, 'write')`" indicates runtime error.

dset0x: You're right, thanks!

invenio/utils/deprecation.py
```diff
@@ -87,17 +90,61 @@ class RemovedInInvenio22Warning(DeprecationWarning):
...
+    """
     def wrap(func=None):
-        """Decorator which can be used to mark functions as deprecated.
-
-        :param message: text to include in the warning
-        :param category: warning category
-        """
         @wraps(func)
         def new_func(*args, **kwargs):
+            original_showwarning = warnings.showwarning
```
jirikuncar: it would be nice to change this only in `current_app.debug` mode.

dset0x: The reasoning of always doing this so that we can always print print "Full backtrace in the logfile if DEBUG is enabled" is to let users know that they can get more information on these warnings by enabling DEBUG.

jirikuncar: I'm still not convinced. Adding "Full backtrace in the logfile if DEBUG is enabled" to all deprecation messages only makes logs bigger. I would rather see adding optional `stacklevel` argument to `def deprecated(message, category):` that allows developer to correctly specify the deprecated()‘s caller.

dset0x: > I would rather see adding optional stacklevel argument to def deprecated(message, category): that allows developer to correctly specify the deprecated()‘s caller.

The *developer* could have an argument for this, but the stacklevel is always going to be 3, since this is a decorator. How would the *caller* of a function or user of a property pass an argument here, especially if they are implicitly making a call (ie property)?

I can remove the expansion of `message.args`, but this would leave everyone wondering how to deal with future warnings that point to the wrong line again.

jirikuncar: Maybe you should consider using `inspect` module to get correct filename and line number of the `func` instead and use [`warn_explicit`](https://docs.python.org/2/library/warnings.html#warnings.warn_explicit).

```python
warnings.warn_explicit(
    message,
    category,
    func.im_func.func_code.co_filename, 
    func.im_func.func_code.co_firstlineno
)
```

dset0x: > Maybe you should consider using inspect module to get correct filename and line number of the func instead and use warn_explicit.

This was in one of my proposals, but it's going to be guesswork. My idea regarding that was to go through the stacktrace, top to bottom, and grab the first that points to invenio. However, that line can easily be unrelated.

jirikuncar: Honestly, I would rather see removing this decorator and using `warnings.warn` directly.

dset0x: Can't `warnings.warn` be equally underinforming without a stacktrace?

jirikuncar: It can. Hence, I would like to come back to my first proposal in https://github.com/inveniosoftware/invenio/pull/3144#discussion_r31118866. Do expensive operations only in debug mode.

dset0x: Changing the value of showwarning is not expensive. Resolving and writing the stacktrace is and we only do that with debug mode. The question is how much of the stacktrace we write and what the contents of the logged message are.

jirikuncar: @dset0x we can discuss in wider circle with @jalavik. I'm mostly worried about https://github.com/inveniosoftware/invenio/pull/3144/files#diff-c3d7fb1347aeebd12332c5272e0f7051R101 being called on "production" system.

dset0x: @jirikuncar This is understandable. It can perhaps be reduced to something faster with a single multiline regex. Either way, we can completely leave `showwarning` alone since the length of the messages is also a concern, but we'll have to keep in mind that the line that is blamed can be far off, both with `deprecated` and manual invocation of `warning`.

I understand that there is also an issue where @jalavik is involved that concerns deprecation warnings. Let's try to tackle both problems.

dset0x: After a short chat before Jan left, we are leaning towards this solution:

* Do not manipulate the message logged to the console.
* Introduce a variable for enabling logging stacktraces of deprecationwarnings to the logfiles, for those interested in tackling them.

The problem of reproducibility remains, but we can live with it, since these stacktraces can make the logfile blow up in size tremendously.

dset0x: @jirikuncar What do you think about the above?

jirikuncar: @dset0x I'm fine with the implementation if it's guarded with:
```python
if current_app.debug or current_app.config.get('CHOOSE_SOME_NICE_NAME', False):
```

dset0x: @jirikuncar All developers have debug on. Do you mean `and` instead of `or`?

jirikuncar: But you don't have debug on in production ... you can of course say `and` instead `or` but I'm not sure if it's what you want. (cc @jalavik)

invenio/utils/deprecation.py
```diff
@@ -87,14 +115,61 @@ class RemovedInInvenio22Warning(DeprecationWarning):
...
+        not re.search('^  File .*('+'|'.join(ignore_path_matches)+')',
+                      formatted_stack, flags=re.MULTILINE)
+    )
+
+    # Debug
+    if log_this:
+        try:
+            current_app.logger.debug(formatted_warning)
+            current_app.logger.debug(formatted_stack)
+        except RuntimeError:
```
dset0x: We've already guarded against this on 129. TODO

dset0x:
@jmartinm Feel free to give this a go. It now affects deprecation warnings in general and introduces some new variables to tune what gets shown and where. If you do, let me know how you like the variables so that we can merge it in time for 2.1.1.

invenio/utils/deprecation.py
```diff
@@ -70,12 +70,47 @@ def old_method(a, b):
 be API backward compatible so that existing code does not break. Please start
 fixing all places ``old_method`` is used, so that it can be easily removed
 once Invenio v2.1 has been released.
+
+Configuration Settings
+----------------------
+The verbosity and logging of deprecation warnings rely on these variables:
+
+===================================== ==========================================
+`CFG_DEPRECATION_DBG_IGNORED_REGEXES` Regular expressions that are matched
```
jirikuncar: please remove `CFG_` prefixes to follow new naming scheme

dset0x: Done, thanks. Where did I miss this change?

tiborsimko:
@jalavik @jmartinm Still wanted for say `maint-2.1` or can we close this as per Invenio 3?

jalavik:
@tiborsimko I think we can close as per Invenio 3.

tiborsimko:
Thanks, closing.

PR #3148, opened 2015-05-20
WebSearch: no special behaviour for subject index

tiborsimko:
* FIX Removes special behaviour of the "subject" index that was
  hard-coded based on the index name.  Installations should rather
  specify wanted behaviour by means of configurable tokeniser instead.

Reviewed-by: Jan Aage Lavik <jan.age.lavik@cern.ch>


jalavik:
:shipit:

PR #3158, opened 2015-05-26
RFC global: sessions 2.0

crepererum:
* Upgrades session serialization to LZMA.

* Implements session signing. This avoids cache request for invalid
  sessions and reduces the DDoS attack surface.

* Removes IP address storage+checks. This avoids data privacy issues and
  enables users with multiple connections (e.g. WIFI+LTE, multiple WIFI
  connections on trains+stations) to stay signed in.

* INCOMPATIBLE Drops all active sessions during upgrade. Might result in
  log entries about non-restorable sessions.

### DIFF
--- a/invenio/ext/session/interface.py
+++ b/invenio/ext/session/interface.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2011, 2012, 2013, 2014 CERN.
+# Copyright (C) 2011, 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


invenio/ext/session/interface.py
```diff
@@ -69,6 +74,20 @@ def backend(self):
...
 
+    @locked_cached_property
+    def sign_sid(self):
+        """Return ``True`` if cookie session id is signed."""
+        return current_app.config.get('SESSION_SIGN', True)
+
+    @locked_cached_property
+    def sign_salt(self):
+        """Return salt for session id signing."""
+        return current_app.config.get('SESSION_SALT', 'm4jySOhiq9wLYXxLMIIsbfKLHwsbC6e2')
```
jirikuncar: `inveniosalt`

PR #3162, opened 2015-05-27
records: fix type consistency of record

### LINKED ISSUE #2772: jsonalchemy: `Record.get_record` is inconsistent regarding `str` and `unicode` types
`invenio.modules.records.api:Record.get_record` returns a dict with keys and values of type `str` when it is first called.

After the first call, and once the result has been cached in `invenio.modules.records.models:RecordMetadata` (retrieved via `invenio.modules.jsonalchemy.jsonext.engines.sqlalchemy.SQLAlchemyStorage` in the case of `get_record`), both keys and values are now `unicode` instead of `str`.

A simple fix would be to amend its [last `return` statement](https://github.com/dset0x/invenio/blob/pu/invenio/modules/records/api.py#L91) to
```python
return cls.storage_engine.get_one(recid)
```

This may add some overhead in the function call, but since this interfaces for the database, it would probably be better to be as consistent as possible with the result.

jirikuncar:
* FIX Improves type consistency of keys and values in JSON record
  created from MARC and retrieved from storage engine.  (closes #2772)

### DIFF
--- a/invenio/modules/records/api.py
+++ b/invenio/modules/records/api.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2013, 2014 CERN.
+# Copyright (C) 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


lnielsen:
:shipit:

PR #3163, opened 2015-05-27
accounts: verification email is not being sent fix

bouzlibop:
* Corrects conditions on when to sent an email. (addresses
  zenodo/zenodo#275)

### DIFF
--- a/invenio/modules/accounts/models.py
+++ b/invenio/modules/accounts/models.py
@@ -70,7 +70,7 @@ class User(db.Model):
     password_salt = db.Column(db.String(255))
     password_scheme = db.Column(db.String(50), nullable=False, index=True)
 
-    note = db.Column(db.String(255), nullable=True)
+    _note = db.Column(db.String(255), name="note", nullable=True)
     given_names = db.Column(db.String(255), nullable=False, server_default='')
     family_name = db.Column(db.String(255), nullable=False, server_default='')
     settings = db.Column(db.MutableDict.as_mutable(db.MarshalBinary(
@@ -83,6 +83,16 @@ class User(db.Model):
 
     PROFILE_FIELDS = ['nickname', 'email', 'family_name', 'given_names']
     """List of fields that can be updated with update_profile."""
+
+    @hybrid_property
+    def note(self):
+        """Return the note."""
+        return self._note
+
+    @note.setter
+    def note(self, note):
+        """Set the note."""
+        self._note = str(note)
 
     @hybrid_property
     def password(self):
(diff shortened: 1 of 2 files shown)


bouzlibop:
\cc @lnielsen

invenio/modules/accounts/models.py
```diff
@@ -135,8 +135,8 @@ def verify_password(self, password, migrate=False):
 
     def verify_email(self, force=False):
         """Verify email address."""
-        if force or self.note == "2":
-            if self.note != "2":
+        if force or str(self.note) == "2":
```
jirikuncar: can you rather fix the setter of `self.note` to always convert everything to `text_type`?

bouzlibop: :+1:

invenio/modules/accounts/models.py
```diff
@@ -84,6 +84,16 @@ def __str__(self):
     PROFILE_FIELDS = ['nickname', 'email', 'family_name', 'given_names']
     """List of fields that can be updated with update_profile."""
 
+    @property
```
jirikuncar: `@db.hybrid_property`

bouzlibop: I've just changed that :) What is the gain of having `hybrid_property`? (I read [here](http://docs.sqlalchemy.org/en/latest/orm/extensions/hybrid.html) that it can be used both at the class level and instance level, but here I guess we are only using it for instance level?)

jirikuncar:
@bouzlibop can you also include a simple test case to cover this issue? Thanks

bouzlibop:
I've just added simple test case for `note` field. I've overwritten existing (but empty) test suit file - I'm not sure if I'm not breaking anything?

bouzlibop:
build is failing, but the error is not connected with this PR - @hachreak already created an issue in https://github.com/kvesteri/sqlalchemy-utils to resolve the problem

invenio/modules/accounts/testsuite/test_accounts_user.py
```diff
@@ -19,15 +19,45 @@
...
-from invenio.testsuite import make_test_suite, run_test_suite  # , InvenioTestCase
 
-TEST_SUITE = make_test_suite()
+class BaseTestCase(InvenioTestCase):
+    """Base test case."""
+
+    def setUp(self):
+        """Clear tables."""
+        from invenio.modules.accounts.models import User
+        User.query.delete()
```
jirikuncar: do you really have to do this? `admin` user has to be always present.

bouzlibop: Yeah, I can go without clearing this table.

bouzlibop:
[lastest](https://travis-ci.org/inveniosoftware/invenio/jobs/64414568) is passing, `release`, `devel` need fixes mentioned by @hachreak

PR #3164, opened 2015-05-27
installation: database init for any host

jirikuncar:
* BETTER Adds default priviledges for database user to access from any
  host.

--
cc @hachreak @crepererum 

PS: It's going to `maint-2.0`.

### DIFF
--- a/invenio/base/scripts/database.py
+++ b/invenio/base/scripts/database.py
@@ -86,6 +86,9 @@ def init(user='root', password='', yes_i_know=False):
             (cmd_prefix + '-e "GRANT ALL PRIVILEGES ON '
              '{CFG_DATABASE_NAME}.* TO {CFG_DATABASE_USER}@localhost '
              'IDENTIFIED BY {CFG_DATABASE_PASS}"'),
+            (cmd_prefix + '-e "GRANT ALL PRIVILEGES ON '
+             '{CFG_DATABASE_NAME}.* TO {CFG_DATABASE_USER}@\'%\' '
+             'IDENTIFIED BY {CFG_DATABASE_PASS}"'),
             cmd_admin_prefix + 'flush-privileges'
         ]
         for cmd in cmds:


invenio/base/scripts/database.py
```diff
@@ -86,6 +86,9 @@ def init(user='root', password='', yes_i_know=False):
             (cmd_prefix + '-e "GRANT ALL PRIVILEGES ON '
              '{CFG_DATABASE_NAME}.* TO {CFG_DATABASE_USER}@localhost '
              'IDENTIFIED BY {CFG_DATABASE_PASS}"'),
+            (cmd_prefix + '-e "GRANT ALL PRIVILEGES ON '
+             '{CFG_DATABASE_NAME}.* TO {CFG_DATABASE_USER}@\'%\' '
+             'IDENTIFIED BY {CFG_DATABASE_PASS}"'),
```
kaplun: This is quite open. We should document that this is recommended only for development instances, and possibly print a warning...

crepererum: You still have a username+password. So as an admin you should secure your MySQL instance, which includes port blocking and optional (auto-)blacklisting. I would also say that the new config is more modern because it does not assume any more that MySQL and Invenio run on the same system (even in the same container/chroot/jail/...)

tiborsimko: >  it does not assume any more that MySQL and Invenio run on the same system

FWIW we have been recommending to run Apache and MySQL on separate hosts since many, many years... The GRANT instructions were simply printed out and people were left to set them up themselves, using real IP addresses.  The "easy-db-init" way was used mostly on local developer boxes only.

As for the "opening up the barriers",  it is true that allowing `%` widens possibilities for an accidental break-in or a mishap, e.g. in case people use the same credentials locally and on a remote system.  However the DB is usually protected by another level, e.g. Invenio Puppet module updates DB OS firewall for each declared Web and Worker nodes, hence the situation should not be too bad.

Anyway, if the main motivation was to allow connectivity from Docker containers, one could in theory test whether the init command is run from within Docker, and allow `%` then and only then.  E.g. by checking `$(cat /proc/1/cgroup | grep -c docker)`.

(Another, even simpler possibility to address "opening up the barrier" concerns would be to document and print out the information that `%` was used, so that the admin would be informed first hand.)

crepererum:
Working on docker.

invenio/base/scripts/database.py
crepererum: We should avoid putting Docker-specific code into the Invenio core (aka Python code) and should stay with separate Dockerfile + scripts. There are plenty other deployment methods out there that might get used by our users. Ref. [Hollywood principle](https://en.wikipedia.org/wiki/Hollywood_principle)

PR #3166, opened 2015-05-27
RFC global: namespace package for modules

crepererum:
* Adds `invenio.modules` as a namespace packages. This simplifies the
  transition for external modules until we have proper entry points. The
  setuptools hack is required until we use a python version with PEP
  0420 (3.3). This might break module autocompletion for some tools,
  e.g. iPython.

See https://github.com/inveniosoftware/cookiecutter-invenio-module/pull/12 for a usage example.

### DIFF
--- a/invenio/__init__.py
+++ b/invenio/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # This file is part of Invenio.
-# Copyright (C) 2014 CERN.
+# Copyright (C) 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -28,3 +28,6 @@ protocol and uses MARC21 as its underlying bibliographic standard.
 
 # Version information
 from .version import __version__
+
+# namespace package
+__import__("pkg_resources").declare_namespace(__name__)
--- a/invenio/modules/__init__.py
+++ b/invenio/modules/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2013 CERN.
+# Copyright (C) 2013, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -16,3 +16,8 @@
 # You should have received a copy of the GNU General Public License
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Modules that are used to compose an Invenio installation."""
+
+# namespace package
+__import__("pkg_resources").declare_namespace(__name__)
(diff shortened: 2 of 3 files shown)


jirikuncar:
WDYT @tiborsimko @lnielsen @jalavik @egabancho?

jalavik:
Sounds reasonable. I read up about it here: https://pythonhosted.org/setuptools/setuptools.html#namespace-packages

And I have a question regarding `You must NOT include any other code and data in a namespace package’s __init__.py`. Is this only at toplevel or every level of the module tree? As we do sometimes have code/data in __init__.py.

crepererum:
- `invenio/__init__.py`: not loaded anymore (so no `invenio.__version__`, use `invenio.version.__version__` instead)
- `invenio/modules/__init__.py`: not loaded
- `invenio/*.py` (except `__init__.py`) no change
- `invenio/modules/*.py` (except `__init__.py`) no change
- `invenio/**` (except `modules`) no change
- `invenio/modules/**/...` no change, also `invenio/modules/foo/__init__.py` gets loaded as before

lnielsen:
First of all, long-term I'm against using namespace packages. Short-term, I see both pros and cons of temporarily introducing them. 

Technically I don't think anything that prevents us from making ``invenio.modules`` into a namespace package. Only ``invenio.modules.__init__`` needs to be empty (i.e with the namespace package declaration), which it currently is.

Cons:
- Namespace packages have lots of installation pitfalls (speaking by experience having managed a large application using namespace packages). Our installation procedure will likely guard against this, but if a user installs it any other way you risk ending up with a pretty messed up ``site-packages``.
- We're just postponing the pain - the work of fixing imports has to be done at some point.
- Risk of staying with namespace packages.

Pros:
- Faster migrationer due to less up-front work.
- Possible to perhaps externalise a module, and let it die (i.e. not having to do the migration).

> invenio/__init__.py: not loaded anymore (so no invenio.__version__, use invenio.version.__version__ instead

Just a side comment, that I don't agree with this. Current practice is that ``invenio.__init__`` contains a line ``from .version import __version__`` so that version can be accessed by ``from invenio import __version__``. The version is in a separate file, so that it can be easily read and parsed without loading the entire project by e.g. ``setup.py``.

crepererum:
Loading `invenio.__init__` is technically not possible with the *old* Python version we're currently using. If `invenio.modules` should be a namespace package, `invenio` also have to be one.  Plan B would be introducing `invenio_modules` as a namespace. Plan C would be introducing entry points now. What scares me from both is the fact that Invenio + external modules is not installable right now and what we rip out more and more modules without having a solution for the transition plan.

>  but if a user installs it any other way you risk ending up with a pretty messed up site-packages

What other ways/methods do you have in mind?

tiborsimko:
> What scares me from both is the fact that Invenio + external modules is not installable right now 

If we proceed with separating "important" modules out, and if this takes time, then this would be indeed not good.  We can complete our overall migration plans denoted here:

- http://simko.home.cern.ch/simko/tmp/invenio-3-roadmap-2015-04-01/invenio-3-legacy-renaissance.png

which may guide as to whether the temporary is interesting in place of permament.

(E.g. it may be that we can perhaps speed up the work on creating an ultra small Invenio core package with which the newly separated modules will work well...)

jirikuncar:
We continue with current package naming conventions.

- `invenio-<modulename>` for package name
- `invenio_<modulename>` for module sources

PR #3172, opened 2015-05-28
access: usedeposit action addition

### LINKED ISSUE #2724: RFC deposit: filter depositions based on user roles

jirikuncar:
* NEW Adds usedeposit action which enables per user access
  restrictions for different deposit types.  (closes #2724)

* Requires running `webaccessadmin -u admin -c -a -D` command.

Signed-off-by: Jan Stypka <jan.stypka@cern.ch>

### DIFF
--- a/invenio/ext/login/legacy_user.py
+++ b/invenio/ext/login/legacy_user.py
@@ -44,6 +44,7 @@ CFG_USER_DEFAULT_INFO = {
     'guest': '1',
     'session': None,
     'precached_permitted_restricted_collections': [],
+    'precached_allowed_deposition_types': [],
     'precached_usebaskets': False,
     'precached_useloans': False,
     'precached_usegroups': False,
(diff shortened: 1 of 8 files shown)


lnielsen:
:shipit:

PR #3173, opened 2015-05-29
records: port of BibDocFile serving

### LINKED ISSUE #3160: bibdocfile: 404 on file URLs
How to reproduce
=============

For example, in the Invenio demo site, click on any file (with URL like `http://localhost:8888/record/59/files/0204143.pdf`)

to get:

screenshot 2015-05-26 22 18 05

Same happens if a new file is uploaded through FFT and then tries to be accessed. The metadata reflects the change as field 856 is created (and also the file is listed in `bibdocfile` cli) but it is not accessible in the URL.

jirikuncar:
* NOTE Ports basic BibDocFile serving including access right checks.
  (closes #3160)

--
This commit brings back only limited support for BibDocFile serving. Do you have any critical use-cases that are missing?

cc @lnielsen @tiborsimko @egabancho @kaplun

### DIFF
--- a/invenio/modules/documents/utils.py
+++ b/invenio/modules/documents/utils.py
@@ -78,8 +78,8 @@ def _get_document(uuid):
         return path
 
 
-def _get_legacy_bibdoc(recid, filename=None):
-    """Get the the fullpath of legacy bibdocfile.
+def _get_legacy_bibdocs(recid, filename=None):
+    """Get all fullpaths of legacy bibdocfile.
 
     :param int recid: The record id
     :param str filename: A specific filename
@@ -88,12 +88,23 @@ def _get_legacy_bibdoc(recid, filename=None):
     """
     from invenio.ext.login import current_user
     from invenio.legacy.bibdocfile.api import BibRecDocs
-    paths = [
+    return [
         (bibdoc.fullpath, bibdoc.is_restricted(current_user))
         for bibdoc in BibRecDocs(recid).list_latest_files(list_hidden=False)
         if not bibdoc.subformat and not filename or
         bibdoc.name + bibdoc.superformat == filename
     ]
+
+
+def _get_legacy_bibdoc(recid, filename=None):
+    """Get the fullpath of legacy bibdocfile.
+
+    :param int recid: The record id
+    :param str filename: A specific filename
+    :returns: bibdocfile full path and access rights
+    :rtype: tuple
+    """
+    paths = _get_legacy_bibdocs(recid, filename=filename)
     try:
         path = paths[0]
     except IndexError:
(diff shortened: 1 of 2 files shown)


invenio/modules/documents/utils.py
```diff
@@ -88,12 +88,23 @@ def _get_legacy_bibdoc(recid, filename=None):
     """
     from invenio.ext.login import current_user
     from invenio.legacy.bibdocfile.api import BibRecDocs
-    paths = [
+    return [
         (bibdoc.fullpath, bibdoc.is_restricted(current_user))
         for bibdoc in BibRecDocs(recid).list_latest_files(list_hidden=False)
         if not bibdoc.subformat and not filename or
         bibdoc.name + bibdoc.superformat == filename
```
kaplun: Sorry seeing this for the first time: How can one retrieve a file having a subformat? e.g. 
`foo.pdf;pdfa`? Note also that what you obtain from `list_latest_files` are `BibDocFile` objects not `BibDoc` objects.

Can this filter be changed e.g. to:
`if not filename or bibdoc.get_fullname() == fullname`?

invenio/modules/records/views.py
```diff
@@ -231,6 +231,14 @@ def file(recid, filename):
             return send_file(file_, mimetype='application/octet-stream',
```
kaplun: Why the previous lines are reading the whole file into memory :scream: ?

invenio/modules/records/views.py
```diff
@@ -231,6 +231,14 @@ def file(recid, filename):
...
                              attachment_filename=filename)
         return send_file(document['uri'])
+
+    from invenio.modules.documents.utils import _get_legacy_bibdocs
+    for fullpath, permission in _get_legacy_bibdocs(recid, filename=filename):
+        if not permission:
+            error = 401
+            continue
+        return send_file(fullpath)
```
kaplun: Here we are not passing the mimetype... which is not going to be guessable from the fullpath which will look something like: .../foo.pdf;pdfa;1

BibDocFile will give you the correct mimetype.

kaplun:
I note that the `file` view does not support the `subformat` and `version` arguments. At least for `subformat` these are available in existing MARCXML and need to be correctly handled (e.g. for CDS icon handling etc.)

invenio/modules/documents/utils.py
jirikuncar: > How can one retrieve a file having a subformat?

I don't know.

> Can this filter be changed e.g. to:
> `if not filename or bibdoc.get_fullname() == fullname`?

Probably it can. I took a work from @drjova

invenio/modules/records/views.py
jirikuncar: see https://github.com/inveniosoftware/invenio/pull/3173#discussion-diff-31309477R227

PR #3178, opened 2015-05-29
base: apache 2.4 config syntax fix

lnielsen:
* Fixes syntax error in generated Apache virtual host
  configuration.


lnielsen:
Ups, wrong branch

PR #3179, opened 2015-05-29
base: apache 2.4 config syntax fix

lnielsen:
* Fixes syntax error in generated Apache virtual host
  configuration.


lnielsen:
Ping @crepererum 

@tiborsimko should go in 2.1 as well, otherwise the generated apache configuration is broken.

jalavik:
@lnielsen Indeed! This we noticed too. You need to add it also on line 122.

lnielsen:
@jalavik Thanks. Fixed line 122 as well.

crepererum:
:+1:

PR #3186, opened 2015-05-29
jinja2hacks: only warnings in DEBUG mode

jalavik:
* BETTER Changes the hacks to only triggers warnings from non-unicode
  strings in jinja templates when in DEBUG mode to avoid taking a
  performance hit in production. This is important until the data model
  is only represented in Unicode.

CC @lnielsen

### DIFF
--- a/invenio/ext/jinja2hacks.py
+++ b/invenio/ext/jinja2hacks.py
@@ -113,7 +113,7 @@ def utf8escape(s):
     """
     if isinstance(s, str):
         warnings.warn("Convert string '{0}' in template to unicode.".format(s),
-                      RuntimeWarning, stacklevel=3)
+                      DeprecationWarning, stacklevel=3)
         return jinja2_escape(s.decode('utf8'))
     return jinja2_escape(s)
 # Ensure function name is identical to replaced function.
@@ -136,6 +136,6 @@ class Markup(jinja2_Markup):
             encoding = 'utf8'
             warnings.warn(
                 "Convert string '{0}' in template to unicode.".format(base),
-                RuntimeWarning, stacklevel=3)
+                DeprecationWarning, stacklevel=3)
         return jinja2_Markup.__new__(cls, base=base, encoding=encoding,
                                      errors=errors)


kaplun:
I am not yet skilled to be able to review your meta-code but I fully support the idea. These types of warnings are for staging areas and testing infrastructure, on production we can live without and simply live with regular exceptions.

invenio/ext/jinja2hacks.py
```diff
@@ -106,18 +109,21 @@ def setup_app(app):
...
-        warnings.warn("Convert string '{0}' in template to unicode.".format(s),
-                      RuntimeWarning, stacklevel=3)
-        return jinja2_escape(s.decode('utf8'))
-    return jinja2_escape(s)
-# Ensure function name is identical to replaced function.
-utf8escape.__name__ = jinja2_escape.__name__
+        WARNING: Do not use this method. Use jinja2.escape() instead.
+        """
+        if isinstance(s, str):
+            if app.debug:
```
lnielsen: Not sure what would be the best approach 1) debug as you've done here, or 2) a special config variable to switch on/off (i.e. it might be useful at some later point to switch it on in production).

jirikuncar:
> This is important until the data model is only represented in Unicode.

It has been done in dcac164e3b436b1406a89d5b2a3d6273c91f84a0. I'm not a big fan of this change. You can completely disable deprecation messages in production.

invenio/ext/jinja2hacks.py
jalavik: Yeah, we can change it to 2) if preferred, however this flag should **by default** be turned off on production systems. It was due to this production vs. development I took advantage of the `app.debug` setting.

jirikuncar:
@jalavik alternatively we can change `RuntimeWarning` by `DeprecationWarning` (or `RemovedInInvenio22Warning`).

jalavik:
As far as I have seen, the `warnings.warn` call is still producing overhead when rendering templates on production. We should at least provide good defaults and not raise any deprecation warnings in production systems by default.

invenio/ext/jinja2hacks.py
```diff
@@ -106,18 +109,21 @@ def setup_app(app):
...
+                warnings.warn("Convert string '{0}' in template to unicode.".format(s),
+                              RuntimeWarning, stacklevel=3)
```
jirikuncar: The problem is in using `RuntimeWarning` instead of `DeprecationWarning`.

jalavik: Aight. I can set it to `DeprecationWarning` and see. Won't it still be raised by default and generate overhead unless we set some ignore deprecation warning flag somewhere?

lnielsen:
I can see pros/cons of completely disabling all deprecation warnings in production. If you have a QA system you ought to catch them there and not in production, on the other hand you might catch things in production which didn't appear in QA.

Anyway, in addition to changing behaviour of this warning, then we might consider adding a feature to silence the warnings in https://github.com/inveniosoftware/invenio/blob/master/invenio/base/factory.py#L155

jirikuncar:
off-topic: @jalavik what is the most common cause `utf8` errors in INSPIRE templates?

jalavik:
@lnielsen @jirikuncar I just want to stress that we should disable these in the *default* provided production system. It will simply slow down Invenio, give bad impressions to people etc, developers and admins can catch and deal with the deprecation on *development* level. Let's not be silly and make everyone having to set yet another config variable to get a proper running production system.

kaplun:
@jirikuncar although this might be minor for developers but it is major for the stake holders of Invenio i.e. final users and Invenio installation admins such as INSPIRE. 
Even if the solution at hand might not be perfect, doesn't make the issue less important.

jalavik:
Following your proposals, what about the following:

- [ ] Indeed change `RuntimeWarning` to `DeprecationWarning`
- [ ] By default, disable the raising of DeprecationWarning in https://github.com/inveniosoftware/invenio/blob/master/invenio/base/factory.py#L155 unless `app.debug or SOME_OVERRIDE_FLAG`.

jirikuncar:
I agree with both points in https://github.com/inveniosoftware/invenio/pull/3186#issuecomment-107341835.

lnielsen:
I agree with https://github.com/inveniosoftware/invenio/pull/3186#issuecomment-107341835 as well.

jalavik:
Cool. Thanks guys. I think this should do it.

> what is the most common cause utf8 errors in INSPIRE templates?

I dunno all places affect right now, but one place is the Holding Pen/workflows where we store the SmartJSON objects from the current jsonalchemy that unfortunately is not returning Unicode yet. This will be addressed as well.

invenio/base/factory.py
```diff
@@ -155,6 +152,12 @@ def configure_warnings():
         warnings.simplefilter("default", DeprecationWarning)
 
 
+def configure_deprecation_warnings(app):
+    """Configure deprecation warnings if `DEBUG=True` or ."""
+    if not app.debug and not app.config.get("DEPRECATION_WARNINGS_PRODUCTION_ENABLED"):
```
jirikuncar: Can we move the condition before `warnings.simplefilter("default", DeprecationWarning)`?

jirikuncar:
@jalavik we can merge the fix in `jinja2hacks` and the rest can be done in your overlay by simply adding `configure_deprecation_warnings` in your `EXTENSIONS`.

`invenio/ext/warnings.py`
```python
import warnings

def disable_deprecation_warnings(app):
    warnings.filterwarnings("ignore", category=DeprecationWarning)
```

`inspire/base/config.py`
```python
EXTENSIONS += _EXTENSIONS + ['invenio.ext.warnings:disable_deprecation_warnings']
```

invenio/base/factory.py
jalavik: Not easily as we do not have the app object and configuration loaded at that point.

jalavik:
I disagree that this should be an overlay specific thing. This issue is important enough to cover any Invenio instance. Do we **really** want any generic Invenio production site to be slowed down by deprecation alerts?

jirikuncar:
> I disagree that this should be an overlay specific thing. This issue is important enough to cover any Invenio instance. Do we really want any generic Invenio production site to be slowed down by deprecation alerts?

Deprecation alerts should be solved in QA. We have all agreed on [deprecation policy](http://invenio.readthedocs.org/en/latest/deprecationpolicy.html). Normally each deprecation warning is shown only once per running process, hence the performance hit should not be too high.

kaplun:
> Deprecation alerts should be solved in QA. We have all agreed on deprecation policy. Normally each deprecation warning is shown only once per running process, hence the performance hit should not be too high.

To think we can address all the deprecation warning on QA is a bit delusional as we don't have the resources to fix all them (not even Invenio has existing replacement for all the technologies we are deprecating(!)).

Note also that the process is restarted every given amount HTTP requests, and therefore the warning is fired many times per hour.

tiborsimko:
> Do we really want any generic Invenio production site to be slowed down by deprecation alerts?

> [...]

> we don't have the resources to fix all them (not even Invenio has existing replacement for all the technologies we are deprecating(!))

Here are my 2 cents on this topic.

As we are splitting Invenio into many packages, and as Invenio v2.1+ is becoming literally a framework of independent components, it will often happen that some Invenio v1.x facility does not have its counterpart in v2.1 and above yet.  Installations needing these features can run Invenio v2.0 which contains fully all the legacy modules.  Installations not needing these features can progressively upgrade to v2.1, v2.2, v2.3, etc, as we shall be cleaning the Invenio core and progressively releasing modules as separate packages.  This may bring many, many deprecation warnings along the way in the coming months, before v3.0 will be finally out.

From the point of view of Invenio being more and more a Pythonic library, instead of a monolithic ILS application, it seems natural to emit deprecation warnings by default, and relying on people switching them out on their production system.  After all, we don't recommend to run Invenio v2.x in production yet; and people testing it out always start on their developer instances anyway.  So, if we provide clear documentation on how to run Invenio in production -- which necessarily passes by creating an overlay -- it seems enough to me to simply document in big fat letters how people can switch various warnings off on their production systems, if they want to.  But, since Invenio v2.x is marketed as a developer version, it seems consistent to me to emit as much warnings as possible, so that services are informed early and can catch and fix any troubles as soon as possible. When the time comes that we shall start recommending Invenio v2.N for production one day, we could perhaps consider switching the default the other way...  The question is, has the time come already to flip the switch that way?  Or not yet?

P.S. It may be interesting to note that Python itself was showing deprecation warnings by default in 2.6 and earlier versions, but as of 2.7 it is not showing them anymore, unless people explicitly ask for them via CLI switch. https://docs.python.org/2/using/cmdline.html#cmdoption-W

jalavik:
> But, since Invenio v2.x is marketed as a developer version, it seems consistent to me to emit as much warnings as possible, so that services are informed early and can catch and fix any troubles as soon as possible.

Several of the most influential services are going for running a Labs site in production (e.g. INSPIRE runs https://labs.inspirehep.net in production today) and thus I believe sensible defaults are important already now.

I still see no benefit for spitting out these things by default *in production*, don't get me wrong, I welcome the fact that these warnings are shown by default to developers locally and to admins on staging areas.

Anyway, what if we keep this patch as is, but I set `DEPRECATION_WARNINGS_PRODUCTION_ENABLED=True` by default?

lnielsen:
IMHO it's perhaps not worth discussing this too much. It's an issue in production for INSPIRE [1], and the patch doesn't really  have any impact on other developers who are not running production sites - i.e. let's integrate the patch and satisfy the needs of the production sites we do have running on 2.x. Flask-AppFactory will require each instance to create their own factory method (a 1-liner), where this can easily be customised as desired.

[1] For Zenodo I'm not sure, but when I enabled Sentry I saw significant increase in response time which could be related.

jalavik:
@lnielsen Thank you :+1:  

Regarding the sentry issue, we noticed it as well. Does the sentry logging run in the same process? It could be better to run it as a separate process if not (to not delay the main process due to slow sending to Sentry).

jalavik:
@jirikuncar Ready to go.

jalavik:
@jirikuncar Right, done. BTW, the overlay stuff worked nicely :+1:

PR #3190, opened 2015-06-01
docker: container usability

crepererum:
This PR improves container usability by adding better boot functionality and persistence. This might be helpful for CI testing with Jenkins.

CC @tiborsimko


tiborsimko:
I may check tomorrow, but I think we'd better commit any docker related improvements to maint-2.0 first, for two reasons: (i) the implementations are largely the same, so it is easy to do and propagate upstream; (ii) having the same Docker setup in all the various v2.x branches can come handy in possible future debugging.

Can you please also look at using non-conflicting ports for MySQL and Redis in `docker-compose` configuration, so that people who *also* run these services on their host OS system would not have to shut them down?

crepererum:
Ready, rebased against `maint-2.0`.

PR #3191, opened 2015-06-01
legacy: ANSI_QUOTES setting on second try

jalavik:
* FIX Sets the ANSI_QUOTES setting also on the second attempt
  to connect to the database in case of a lost connection.

### DIFF
--- a/invenio/legacy/dbquery_mysql.py
+++ b/invenio/legacy/dbquery_mysql.py
@@ -286,6 +286,7 @@ def run_sql(sql, param=None, n=0, with_desc=False, with_dict=False,
         try:
             db = _db_login(dbhost, relogin=1)
             cur = db.cursor()
+            cur.execute("SET SESSION sql_mode = %s", ['ANSI_QUOTES'])
             gc.disable()
             rc = cur.execute(sql, param)
             gc.enable()


jalavik:
Just discovered this issue. Added it to the current milestone due to its small size.

jirikuncar:
:shipit: 

--
fyi @hachreak

jirikuncar:
@jalavik since it has been introduced in 79741888d8e71b22e704a30038f9857d90a38590 I would remove `FIX` label from the commit message. Thanks a lot for yet another catch :+1:

tiborsimko:
> @jalavik since it has been introduced in 7974188 I would remove FIX label from the commit message. 

... or we can amend the original commit message here (via AMENDS 7974188) and also fix the (invalid) COMPATIBILITY label at the same time?

jirikuncar:
>  fix the (invalid) COMPATIBILITY label at the same time?

:+1:

jalavik:
So how would the commit message look with these considerations in mind?

```
legacy: ANSI_QUOTES setting on second try

* BETTER Supports database creation on PostgreSQL server.

* Enables 'sql_mode' as 'ansi_quotes' for quotes compatibility for MySQL.

* AMENDS 7974188
```

jirikuncar:
IMHO @tiborsimko meant `s/COMPATIBILITY/INCOMPATIBLE/`.

jalavik:
So, something like this?

jirikuncar:
:+1:

PR #3195, opened 2015-06-02
BibFormat: correct MARC field usage in bfe_doi

egabancho:
* BETTER Corrects subfield usage when displaying the DOI qualifying
  information, changing subfield `y` (not valid) by `q`.


egabancho:
Running on CDS

tiborsimko:
Thanks, merged as 730468f.

Issue #3015, opened 2015-04-14
RFC security policy


crepererum:
After fixing #3003 and the more or less uncontrolled disclosure of security information, I sat down to write the following RFC. It is not urgent (especially because it is a slightly long text) and might only be a start of a discussion. Feel free to add your comments or criticize it. At a certain point of consensus, I can create a PR out of the material.

# Security Policy (draft)

Invenio reached a state where the impact of security holes and possible exploits
is not only a problem for smaller project, but for bigger installations
worldwide and therefore the organizations using them. This RFC makes
suggestions on

 - how to handle security bugs of a certain impact level
 - which developers should have which permissions regarding the information
   about unpatched security bugs
 - how users, administrators and external developers should report certain types
   of bugs or the usage of exploits
 - how we ensure to be an open project even when dealing with classified
   information

## Severity Classification
There exist several classifications from different organizations, e.g.
[CVSS](https://en.wikipedia.org/wiki/CVSS). These might sued the needs of
bigger organizations. For Invenio, the following simplified model might work
better.

### High Severity
**Description:** A security hole that impacts running installations of released
versions of Invenio (e.g. no pre-releases or developer configurations) in a way
that it is possible to gain unprivileged access to data, to destroy data, to
leak user information, or to render a system unusable with a significant lower
effort than usual; all of that on layer that is reachable by endusers.

**Examples:** bypassing user authentication, remote code execution, XSS, SQL
injections, producing very high server load by malformed inputs.

**Permissions:** Only a small dedicated group of core developers is allowed to
retrieve this information. Other people should only be involved if it is crucial
for fixing the problem.

**How to report:** The bug or usage of certain exploits should only be reported
to a closed mailing list which is only readable by people of the given
permission level.

**Disclosure:** The patch nodes must contain a remark that it is a security
patch for a `High Severity` problem. There must not be a public disclosure on
mailing lists, forum entries, issue tracker or version control system before
a certain level of installations is patched. After it is considered safe to
release the information, it must be added to the public repositories and an
issue with a problem description and solution must be created.

### Low Severity
**Description:** A security problem that does not affect installations directly.
This might be the case because it is not the first layer but a backbone system,
because it requires a misconfiguration or `High Severity` bug to be used or
because it is a theoretical bug, e.g. when using problems with certain
encryption or signature algorithms.

**Exmamples:** Weak file system permissions in Docker images (requires already a
certain level of system access), low quality password hashing (requires database
[...]

PR #3201, opened 2015-06-02
celery: message signing

crepererum:
Adds optional setup of celery message signing and makes docker to use this technology by default.

Rebased against: `maint-2.0`


invenio/celery/__init__.py
```diff
@@ -143,15 +146,28 @@ def import_default_modules(self):
...
             mod.__name__ for mod in registry.tasks or ())
 
 
+# celery security config
+CELERY_SECURITY_SERIALIZER = 'json'
+CELERY_SECURITY_DIGEST = 'sha1'
+
+# try to load config before we have a valid application context
+try:
+    import invenio.base.config as conf
```
lnielsen: I'm not sure you should be using invenio.base.config here. It's only default values and not the user configured ones.

invenio/celery/__init__.py
```diff
@@ -143,15 +146,28 @@ def import_default_modules(self):
...
-        """
+        """Discover task modules named `invenio.modules.*.tasks`."""
         from invenio.celery import tasks
         self.task_modules.update(tasks.__name__)
         self.task_modules.update(
             mod.__name__ for mod in registry.tasks or ())
 
 
+# celery security config
+CELERY_SECURITY_SERIALIZER = 'json'
```
lnielsen: Should this go to Celery's default configuration?

invenio/celery/__init__.py
```diff
@@ -159,3 +175,7 @@ def autodiscover(self):
     'invenio',
     loader=InvenioLoader,
 )
+celery.setup_security(
```
lnielsen: Can't this be done in the InvenioLoader init (first parameter is the Celery app)?

scripts/docker_boot.sh
```diff
@@ -77,6 +91,13 @@ CFG_WEBSUBMIT_BIBCONVERTCONFIGDIR = u'/home/invenio/etc/bibconvert/config'
 CFG_WEBSUBMIT_COUNTERSDIR = u'/home/invenio/var/data/submit/counters'
 CFG_WEBSUBMIT_STORAGEDIR = u'/home/invenio/var/data/submit/storage'
 
+CELERY_SECURITY_KEY = u'/home/invenio/ssl/celery/client.key'
+CELERY_SECURITY_CERTIFICATE = u'/home/invenio/ssl/celery/client.pem'
+CELERY_SECURITY_CERT_STORE = u'/home/invenio/ssl/celery/ca.pem'
+CELERY_TASK_SERIALIZER = u'auth'
+CELERY_SECURITY_SERIALIZER = u'msgpack'
```
lnielsen: I think this is already the default value.

crepererum: Nope, you don't want to do this for all configs. `CELERY_TASK_SERIALIZER = u'auth'` will activate message signing. The paths are installation specific and the `CELERY_SECURITY_SERIALIZER` is `json` by default.

invenio/celery/__init__.py
crepererum: It works, I've tested it ;)

invenio/celery/__init__.py
crepererum: Patch in progress...

lnielsen:
I'm not sure I see the added security benefit of adding message signing in our setup. I can see that message signing in a more open system (i.e. one where clients/workers/brokers doesn't intimately know each other), can be used to add extra trust. However in our production setups RabbitMQ has a virtual host per instance, requires authentication, and firewall is blocking connections from anything but clients and workers.

That said, it's not I'm against adding message signing, but I would favour using a solution that uses our already existing shared secret key. I think that using public key signing just adds extra complexity in our setup due to the need for private/public keys and sharing the public keys (and if the same private/public key's are used for all involved parties, there's little difference to a shared secret key).

I aware that you are just using features already in Celery, but perhaps it's not too much work writing a task serializer that uses our secret key to sign messages.

crepererum:
It is optional and only activated if `CELERY_TASK_SERIALIZER = u'auth'`

crepererum:
## Longer answer:

`RabbitMQ` is not the standard way proposing in our docs. We use Redis, for different purposes, without write protection.

Our "secret key" is NOT a state-of-the-art key. It is just a secret. If exposed, it is possible to use it for different API and entry points. We do not even have a good rejection mechanism which tells us which effects a change of the secret will have.

Using the same key for different purposes is a bad idea.

Arguing that we already have a firewall does not help. Should you somehow get write access to the network (in case the firewall fails) you can execute every possible command our workers (remote code execution). Adding additional layers of security is good practice and is expected by project from our size.

The size of our codebase is so enormous that I would even bet that it contains at least one critical security bug.

A last word on firewalls: it might protect one specific way to enter your systems (like a wall), but it won't protect you from other (more creative) ways, e.g. when other systems behind it have vulnerabilities.

## Alternative proposal for easier setup:
Dump the `secret_key` and use a proper CA architecture by default to generate all secrets for all subsystems.

<br/>

**ref** #3015

scripts/docker_boot.sh
lnielsen: Got it. I mistaked ``CELERY_SECURITY_SERIALIZER`` for ``CELERY_TASK_SERIALIZER``

lnielsen:
Thanks for answer. Please note, I'm not arguing against adding support for messaging signing. Also, I do recognise that the most easy implementation is to use what Celery already have (and it means we  have less code to maintain). 

My concern is purely related to trade off between complexity and security. Security is also a matter of simplicity and easily manageable solutions. If I understand correct, you would need one private/public key per node, and share all the public keys between the machines to have any benefit over a solution based on a shared secret (which doesn't have to be the one in ``SECRET_KEY``)? 
In the end, I'm concerned that adding something which is hard to manage in a production system won't be used, and thus have no effect, whereas if we enable message signing by default using a shared secret we already have, then it adds security by default. 

**Comments on specific points:**

> RabbitMQ is not the standard way proposing in our docs. We use Redis, for different purposes, without write protection.

For all current production system RabbitMQ is the defacto standard. The fact that only Redis is mentioned in the docs is a bug, and an artefact of that we keep dev setups as simple as possible. Also, our docs basically lacks any kind of production setup guide (the little that is here http://invenio.readthedocs.org/en/latest/getting-started/overlay.html was never running on an actual production system)

> The size of our codebase is so enormous that I would even bet that it contains at least one critical security bug.

Yes, I'm quite sure about that as well, and I don't think it's only 1. Please do go hunt for them. A big reason here I believe is the complexity fo the code base, which is being addressed by our road map for Invenio 3.0

> Our "secret key" is NOT a state-of-the-art key. It is just a secret. If exposed, it is possible to use it for different API and entry points. 

Isn't the same true for the private key which could get expose as easily as our secret key? If our secret key is weak then please ticketize so we can fix it.

> We do not even have a good rejection mechanism which tells us which effects a change of the secret will have.

Agree. Will you ticketize so that we can start working on it? Somekind of deprecation mechanism that would allow us to easily change our secret key on a regular basis would be great. 

> Using the same key for different purposes is a bad idea.

For my understanding, do you have some examples why it would be bad in this case?

> Arguing that we already have a firewall does not help. Should you somehow get write access to the network (in case the firewall fails) you can execute every possible command our workers (remote code execution).

I see this as an argument against using Redis as the broker. RabbitMQ requires authentication (which should naturally happen over an encrypted connection).

> Adding additional layers of security is good practice and is expected by project from our size.

Yes, and I'm also not arguing against adding message signing or the extra security. My concern is tradeoffs between two different solutions to adding extra security.

> A last word on firewalls: it might protect one specific way to enter your systems (like a wall), but it won't protect you from other (more creative) ways, e.g. when other systems behind it have vulnerabilities.

Naturally not. Gaining access to any of the nodes where the private key is stored, would also allow you to do remote code execution.

crepererum:
## What is `SECRET_KEY` good for?
> If I understand correct, you would need one private/public key per node, and share all the public keys between the machines to have any benefit over a solution based on a shared secret (which doesn't have to be the one in SECRET_KEY)? 

All nodes that submit tasks to the worker should ideally have their own `client.{pem,crt}`. These are signed by the common `ca.pem` and therefore can verified using `ca.crt`. The CA data should be protected as good as possible. The client data should be protected as well, but in case of leakage they can be revoked using the CA.

>> Our "secret key" is NOT a state-of-the-art key. It is just a secret. If exposed, it is possible to use it for different API and entry points. 

> Isn't the same true for the private key which could get expose as easily as our secret key? If our secret key is weak then please ticketize so we can fix it.

The secret can not be used for asymmetric cryptography because it does not contain the entropy to generate the required keys. The signing procedure with this secret is only be possible when using an HMAC and therefore the secret itself is required to verify the signatures. This is OK for some solutions (oauth, signed cookies) but is unacceptable when our clients want to verify that they got valid data from us.

## Broker questions (aka move to RabbitMQ)
> For all current production system RabbitMQ is the defacto standard. 

Then let's change the docs. I would change the Docker setup as well.

>> Arguing that we already have a firewall does not help. Should you somehow get write access to the network (in case the firewall fails) you can execute every possible command our workers (remote code execution).

> I see this as an argument against using Redis as the broker. RabbitMQ requires authentication (which should naturally happen over an encrypted connection).

See point before.

## Secret reusage

>> We do not even have a good rejection mechanism which tells us which effects a change of the secret will have.

> Agree. Will you ticketize so that we can start working on it? Somekind of deprecation mechanism that would allow us to easily change our secret key on a regular basis would be great. 

I think it would require multiple secrets (or keys) to realize this, see next point.

>>Using the same key for different purposes is a bad idea.

> For my understanding, do you have some examples why it would be bad in this case?

Imagine the following: We use the secret to produce HMACs for:
 - oauth
 - cookies
 - celery task signing

We usually do not use the secret directly. For technical reasons we usually use the SHA{1,256.512} for this. So in fact our true secret is `z=sha1(SECRET_KEY)`. Now imagine one of our contributors (including me btw) makes a mistake and uses a stupid signature algorithm, which leaks `z` for...say the signing of outgoing notification emails (just a dump example). While the primary impact (unusable signatures for these mails) is not that heavy, the secondary impact (leakage of `z`) affects all the components mentioned above. I rate this mistake as quite likely, because honestly cryptography is a very complicated thing (see the TLS/SSL problems of the past 12 months).

## Addition stuff
>  In the end, I'm concerned that adding something which is hard to manage in a production system won't be used, and thus have no effect, whereas if we enable message signing by default using a shared secret we already have, then it adds security by default. 

I share this concern and would volunteer to automate this process.

>> A last word on firewalls: it might protect one specific way to enter your systems (like a wall), but it won't protect you from other (more creative) ways, e.g. when other systems behind it have vulnerabilities.

> Naturally not. Gaining access to any of the nodes where the private key is stored, would also allow you to do remote code execution.

Imagine the following scenario: You have multiple nodes for different purposes in you network (complete Reds-driven Invenio [Redis, Web, Worker, MySQL] + mail server + analysis/machine learning). They are of course protected by a central firewall. A hacker is able to exploit one of the non-Invenio nodes, say machine learning, because it made a not-very-good (i.e., unverified) update PyPi or whatever. The impact is quite low because in your setup it (=machine learning node) does not contain any critical data. But now he has access to your network. Your Redis machine has his own firewall/iptables to only allow access from Invenio:Web and Invenio:Worker by checking either the MAC-Address or IP-Adress of incoming connections (or even both). For the first filter the hacker changes the MAC-Address of the network card (it is programmed by software, the hardware data is only a suggestion). The second one can be bypassed by [ARP-Spoofing](https://en.wikipedia.org/wiki/ARP_spoofing), at least in the most not-very-good networks (which I guess are 80% out there). Because your admin (sorry) choose a to short [AUTH-secret for redis](http://redis.io/commands/AUTH), the hacker can bruteforce it and gets access to your celery queue. This enables remote code execution. Attack complete. This is why deep security can help and is never to much. In other words: Better wear a helmet.

crepererum:
Blocked until https://github.com/mher/flower/issues/408 gets resolved.

crepererum:
Closed, since I'm not a part of the Invenio dev. team anymore.

PR #3210, opened 2015-06-03
docker: boot2docker documentation

### LINKED ISSUE #3194: docker: support for boot2docker
In general the new docker+docker compose works very nicely. I've tested it on OS X with boot2docker and found some few issues.

**Problem 1** 
Boot2docker creates a minimal VirtualBox VM with docker installed on it and pipes all docker commands to that VM. 
Docker-compose bridges the containers ports to the localhost network interface on the docker host, which in case of boot2docker is localhost on the VM. This means that none of the ports are open to the VM host system (OS X in this case). 

The fix is quite easy as provided by @crepererum:

```
ports:
    - "127.0.0.1:5000:5000"
```
->
```
ports:
    - "5000:5000"
```

It makes sense when the docker host is also the developer machine that the docker containers ports are not opened to your entire LAN. So not sure what the best fix is here. Perhaps just add documentation for boot2docker or provide two different docker-compose files. 

**Problem 2**
Given that problem 1 is fixed, then there's still a problem that ``CFG_SITE_URL`` is set to ``http://localhost:5000`` in ``docker_boot.sh``. Perhaps the fix here is just to ingest ``CFG_SITE_URL`` 
via env variables similar to e.g. ``CFG_DATABASE_HOST``.

crepererum:
* Adds documentation for boot2docker. (closes #3194)

### DIFF
--- a/docs/developers/docker.rst
+++ b/docs/developers/docker.rst
@@ -38,6 +38,22 @@ This builds and runs the docker containers. You can now connect to
 `localhost:28080` to see your Invenio installation. The `admin` user does not
 have any password.
 
+.. note::
+    If you are using `boot2docker`_ you need to set up port forwarding by
+    running the following command in a new terminal:
+
+    .. code-block:: shell
+
+        boot2docker ssh -vnNT \
+            -Llocalhost:28080:localhost:28080 \
+            -Llocalhost:26379:localhost:26379 \
+            -Llocalhost:23306:localhost:23306
+
+    You have to run this after Invenio booted up. Do **not** stop it while you
+    are working with Invenio. Otherwise the port forwarding gets stopped. You
+    have to restart the forwarding when your restart the Docker containers. The
+    process can be stopped using `CTRL-C`.
+
 Should you require a fresh installation and therefore wipe all your instance
 data, run:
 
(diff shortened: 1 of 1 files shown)


docs/developers/docker.rst
```diff
@@ -38,6 +38,22 @@ This builds and runs the docker containers. You can now connect to
...
 have any password.
 
+.. note::
+    If you are using `boot2docker`_ you need to set up port forwarding by
+    running the following command in a new terminal:
+
+    .. code-block:: shell
+
+        boot2docker ssh -vnNT \
+            -Llocalhost:28080:localhost:28080 \
```
jirikuncar: is it correct or shall it better be `-Llocalhost:28080:localhost:8080 \`?

crepererum: Nope, the `CFG_SITE_URL` is configured to port 28080. The reason for this is that the URL sometimes contains the port and therefor it should be the same in `CFG_SITE_URL` and in the fronted (here: the port forwarding). You can force flask to listen on another port, but because our URL handling is sometimes kinda confusing I decided to use the same internal and external port. See `docker-compose-dev.yml` which uses the same setup.

jirikuncar: ok

crepererum:
Ready

PR #3225, opened 2015-06-04
WebAccess: FireRole documentation fix

### LINKED ISSUE #3107: WebAccess: minor ommission in a doc example (regarding groups and fireroles) may produce an exception
It seems that if one follows the example in the WebAccess docs, and tries to enter:
```deny group badguys```
in a firerole, an exception is thrown.

By reading the code, I found out that the solution is VERY simple: I just have to put single or double quotes in the groupname to make it work! So, 
```deny group 'badguys'```

is OK and produces no exceptions.

The requirement to have expressions included in quotes is very reasonable, so my proposal is to simply change the example(s) in the docs, according to this requirement.

kaplun:
* FIX Improves the WebAccess FireRole documentation by providing
  corrected example on how to use groups in FireRole definitions.
  (closes #3107)

Reported-by: Theodoros Theodoropoulos <theod@physics.auth.gr>

### DIFF
--- a/modules/webaccess/doc/admin/webaccess-admin-guide.webdoc
+++ b/modules/webaccess/doc/admin/webaccess-admin-guide.webdoc
@@ -1119,7 +1119,7 @@ II. Example - connecting role and user</a>
 
     Example of role definition:
         allow not email /.*@gmail.com/,/.*@hotmail.com/
-        deny group badguys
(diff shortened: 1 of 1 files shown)


tiborsimko:
Thanks, merged as e7ba096.

PR #3228, opened 2015-06-06
template: adds test 'list'

### LINKED ISSUE #3227: template: jinja2 is missing a simple test for lists
As long as this is lacking from the official jinja2 tests, we should add one to our own filters and tests.

blixhavn:
* Jinja2 lacks a test to check if a variable is a list. `iterable`
  is not sufficient, as it will return true for strings.
  (closes #3227)

### DIFF
--- a/invenio/ext/template/__init__.py
+++ b/invenio/ext/template/__init__.py
@@ -231,4 +231,8 @@ def setup_app(app):
         )
         return rv
 
+    @app.template_test('list')
+    def _is_list(value):
+        return isinstance(value, list)
+
     return app


dset0x:
Sorry for jumping in without understanding the actual problem, but this solution seems to fit your actual problem instead of checking against `list`:

```python
isinstance(foo, collections.Iterable) and not isinstance(foo, six.string_types)
```

invenio/ext/template/__init__.py
```diff
@@ -231,4 +231,8 @@ def _url_args(d, append=u'?', filter=[]):
         )
         return rv
 
+    @app.template_test('list')
```
jirikuncar: Should we call it `is_list`?

blixhavn: Then the jinja test would be `{% if variable is is_list %}`.

jirikuncar: what about `{% if variable is sequence and variable is not string %}`?

jirikuncar:
One should use build-in tests:
- `interable` (http://jinja.pocoo.org/docs/dev/templates/#iterable)
- `string`

PR #3229, opened 2015-06-07
Invenio v2.1.0

tiborsimko:
- [x] deprecation warnings
- [x] release note amendments

### DIFF
--- a/invenio/legacy/bibcheck/__init__.py
+++ b/invenio/legacy/bibcheck/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2013 CERN.
+# Copyright (C) 2013, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -22,6 +22,13 @@ BibCheck API
 
 This API lets other modules interact with bibcheck.
 """
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio22Warning
+
+warnings.warn("Legacy BibCheck will be removed in 2.2.",
+              RemovedInInvenio22Warning)
 
 from invenio.legacy.bibcheck import task as bibcheck_task
 
@@ -53,4 +60,3 @@ def check_record(record, enabled_rules=None):
             plugin["check_record"](record, **rule["checker_params"])
 
     return record
-
--- a/invenio/legacy/bibclassify/__init__.py
+++ b/invenio/legacy/bibclassify/__init__.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibClassify."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio22Warning
+
+warnings.warn("Legacy BibClassify will be removed in 2.2. Please check "
+              "'invenio.modules.classifier' module.",
+              RemovedInInvenio22Warning)
--- a/invenio/legacy/bibedit/__init__.py
+++ b/invenio/legacy/bibedit/__init__.py
@@ -17,10 +17,12 @@
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
+"""Legacy BibEdit."""
+
 import warnings
 
-from invenio.utils.deprecation import RemovedInInvenio22Warning
+from invenio.utils.deprecation import RemovedInInvenio23Warning
 
-warnings.warn("BibEdit will be removed in 2.2. Please check "
+warnings.warn("Legacy BibEdit will be removed in 2.3. Please check "
               "'invenio.modules.editor' module.",
-              RemovedInInvenio22Warning)
+              RemovedInInvenio23Warning)
--- a/invenio/legacy/bibexport/__init__.py
+++ b/invenio/legacy/bibexport/__init__.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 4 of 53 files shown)


invenio/legacy/bibcheck/__init__.py
```diff
@@ -23,6 +23,13 @@
 This API lets other modules interact with bibcheck.
 """
 
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio22Warning
+
+warnings.warn("Legacy BibCheck will be removed in 2.2.",
+              RemovedInInvenio22Warning)
+
```
kaplun: Note, there is an on going native checker implementation in: https://github.com/dset0x/invenio/tree/checker-port

invenio/legacy/bibclassify/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+"""Legacy BibClassify."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio22Warning
+
+warnings.warn("Legacy BibClassify will be removed in 2.2. Please check "
+              "'invenio.modules.classifier' module.",
+              RemovedInInvenio22Warning)
```
kaplun: Note: `classifier` depends on legacy `BibClassify`

invenio/legacy/bibedit/__init__.py
```diff
@@ -17,10 +17,12 @@
...
 import warnings
 
-from invenio.utils.deprecation import RemovedInInvenio22Warning
+from invenio.utils.deprecation import RemovedInInvenio23Warning
 
-warnings.warn("BibEdit will be removed in 2.2. Please check "
+warnings.warn("Legacy BibEdit will be removed in 2.3. Please check "
               "'invenio.modules.editor' module.",
-              RemovedInInvenio22Warning)
+              RemovedInInvenio23Warning)
```
kaplun: Note: `editor` depends on legacy `BibEdit`

invenio/legacy/bibexport/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibExport."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio23Warning
+
+warnings.warn("Legacy BibExport will be removed in 2.3. Please check "
+              "'invenio.modules.exporter' module.",
```
kaplun: Note the `exporter` module is currently just a stub. On the other hand for sitemap support we should rather look into `flask-sitemap`.

invenio/legacy/bibedit/__init__.py
jirikuncar: It can be solved during module removal.

invenio/legacy/bibmatch/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibMatch."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio23Warning
+
+warnings.warn("Legacy BibMatch will be removed in 2.3. Please check "
+              "'invenio.modules.matcher' module.",
```
kaplun: Note that the `matcher` module is currently a stub.

invenio/legacy/bibsched/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibSched."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio23Warning
+
+warnings.warn("Legacy BibSched will be removed in 2.3.",
```
kaplun: Shall we point out to look into celery and flower?

invenio/legacy/bibupload/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibUpload."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio23Warning
+
+warnings.warn("Legacy BibUpload will be removed in 2.3.",
```
kaplun: And this is being replaced by `uploader`.

invenio/legacy/bibclassify/__init__.py
tiborsimko: Yes, and that's true for most other legacy modules.  The deprecation warning there basically says: if someone's looking in `invenio.legacy.foo`, they had better look into `invenio.modules.bar` instead.  It is true that the later one may stop working once the former one is removed... so `bar` has to be either seriously amended or even taken away and released independently, about which this PR does not say anything.

invenio/legacy/bibcheck/__init__.py
tiborsimko: What about starting `invenio-checker` repository via cookiecutter? (perhaps with another name?)

invenio/legacy/bibsched/__init__.py
tiborsimko: We can advertise these when they will be ready.  (Ditto for new uploading ways, etc.)

invenio/legacy/bibupload/__init__.py
jirikuncar: I would not advertise `uploader` as it might not be needed with new `record` API.

invenio/legacy/bibformat/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Legacy BibFormat."""
+
+import warnings
+
+from invenio.utils.deprecation import RemovedInInvenio23Warning
+
+warnings.warn("Legacy BibFormat will be removed in 2.3. Please check "
```
jirikuncar: we can say it will be removed in 2.2 as most of the modules depends on new `formatter` module.

invenio/legacy/bibedit/__init__.py
kaplun: If you talk about only `run_sql()` calls, maybe yes. Otherwise is quite a larger task. Additionally it relies on BibRecord, and MARCXML in general.

invenio/legacy/bibformat/__init__.py
kaplun: ...but not many of the legacy modules being deprecated in this very PR.

invenio/legacy/bibsched/__init__.py
kaplun: What exactly do you mean by ready? Isn't celery ready by simply installing it? It doesn't need to be invenio-ized.

invenio/legacy/bibupload/__init__.py
kaplun: OK, what you say is quite a game changer as we mused at the forum. Would be great if you can document the proposed API in a github issue.

invenio/legacy/bibformat/__init__.py
jirikuncar: We have agreed on removing `bfe` and `bft` support in 2.2, hence there is no need for keeping legacy `bibformat` module.

invenio/legacy/bibcheck/__init__.py
kaplun: @dset0x, @jalavik: @tiborsimko is right! We should factor out the new checker into an independent module.

invenio/legacy/bibupload/__init__.py
jirikuncar: > you say is quite a game changer

It's nothing new. It has been proposed few months ago during the relationship discussions, but there was no wide adoption of the idea.

invenio/legacy/bibsched/__init__.py
tiborsimko: Yes, Celery is ready and we even refer to it *lightly* in the docs already... what I had in mind was mostly indexing/ranking that still needs bibsched.  When these will go away, and when we have new upload mechanism in place, we can advertise new "task scheduling" more prominently, e.g. via before-insert and after-insert hooks...

jirikuncar:
:shipit:

PR #3231, opened 2015-06-08
global: removal of deprecated features

jirikuncar:
- removes deprecated `formatter` features (legacy support for `bft` and `bfe`)
- removes `jinja2hacks`


kaplun:
@jirikuncar, thanks for this mega-branch!

Just amending the milestone, because, as mentioned today at the forum (and represented into Tibor's mindmap), bibformat is used indirectly by many modules that have yet to be deprecated (as detailed in #3233).

So after these have been removed we can proceed to also remove the inner and closer to the core modules such as BibFormat and WebSearch.

jirikuncar:
@kaplun it's going to 2.2 as most of the features has been already deprecated for 2.2.

kaplun:
Then @tiborsimko should amend the mindmap accordingly and kill also all the heavily dependent module WebJournal, and amend OAIRepository and BibSword which directly hook into the old BibFormat API

PR #3245, opened 2015-06-10
oauth2server: fix PostgreSQL compatibility

### LINKED ISSUE #3145: oauth2server: oauth2CLIENT not cleaned if user is deleted
I found that there is not any cascade action after a user is deleted.
So, e.g. if you run testsuite on tags module, at the end we have 14 rows in `oauth2CLIENT` table.

hachreak:
* Sets user_id nullable on Client and Token tables.

* Fixes testsuite to completely clean database records after tests are
  finished. (closes #3145)

### DIFF
--- a/invenio/modules/oauth2server/models.py
+++ b/invenio/modules/oauth2server/models.py
@@ -29,6 +29,7 @@ from invenio.base.i18n import _
 from invenio.config import SECRET_KEY as secret_key
 from invenio.ext.login.legacy_user import UserInfo
 from invenio.ext.sqlalchemy import db
+from invenio.modules.accounts.models import User
 
 import six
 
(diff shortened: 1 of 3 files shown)


invenio/modules/oauth2server/models.py
```diff
@@ -167,7 +168,7 @@ class Client(db.Model):
         default=u'',
     )
 
-    user_id = db.Column(db.ForeignKey('user.id'))
+    user_id = db.Column(db.ForeignKey('user.id'), nullable=True)
```
jirikuncar: upgrade recipe needed

hachreak: That's what I wanted to do.. but then, I checked the db before and after... and, surprisingly, it's always `DEFAULT NULL`.

jirikuncar: :+1:

hachreak:
ping @jirikuncar

PR #3251, opened 2015-06-10
legacy: WebGroup and WebMessage removal

jirikuncar:
* NOTE Removes legacy WebGroup module before.


invenio/legacy/websession/webgroup.py
```diff
@@ -748,13 +748,13 @@ def perform_request_reject_member(uid,
...
-                msg_subject=msg_subjet,
-                msg_body=msg_body,
-                ln=ln)
+            # (body, dummy, dummy) = perform_request_send(
+            #     uid,
+            #     msg_to_user=user,
+            #     msg_to_group="",
+            #     msg_subject=msg_subjet,
+            #     msg_body=msg_body,
+            #     ln=ln)
```
kaplun: Isn't it simply better to kill it for the good? As anyway it will stay in git history?

jirikuncar: I might remove this whole file first.

Issue #3233, opened 2015-06-08
roadmap: legacy removal


lnielsen:
These modules are going to disappear/be reimplemented by v2.2.0 (see also: http://simko.home.cern.ch/simko/tmp/invenio-3-roadmap-2015-04-01/invenio-3-legacy-renaissance.png)

This would accelerate achievement of parallel roadmap #3079.
An orthogonal roadmap related to simplifying the code base is #3126.

- [x] WebAlert (@egabancho)
- [x] WebBasket (@egabancho)
- [x] WebComment (@hachreak)
- [x] WebHelp (@egabancho)
- [x] WebJournal (@egabancho)
- [x] WebLinkBack (@jirikuncar)
- [x] WebMessage (@jirikuncar)
- [x] BibSword
- [x] BibCirculation: (see also #2976) (@egabancho)
- [ ] ckeditor
- [x] errorlib (@lnielsen)
- [ ] goto
- [x] httptest (@jirikuncar)
- [ ] dbdump/dbexec
- [x] inveniogc: (see also #3209)
- [ ] solrutils/xapianutils
- [x] ping: shoud be replaced by https://github.com/Runscope/healthcheck (@lnielsen)
- [x] WebSession (@hachreak)
- [ ] WebStyle
- [x] WebSubmit (@jirikuncar)
- [x] BibIngest (@egabancho)
- [x] ELMSubmit (@hachreak)
- [x] OAIHarvest (@jalavik) Moving to `invenio-oaiharvester`
- [ ] BibConvert
- [x] BibClassify (@jalavik) Ready to move to `invenio-classifier`
- [ ] DocExtract  (@jalavik) Moving to external module (extractutils).
- [ ] RefExtract  (@jalavik) Moving to external module (extractutils). (Support for Grobid, bibrank part to be implemented separately)
- [ ] authorextract   (@jalavik) Moving to external module (extractutils).
- [x] BibAuthorList ~~(@kaplun)~~  (@hachreak)
- [x] BibEditMulti (@jmartinm)
- [x] BatchUploader (@lnielsen)
- [ ] BibCatalog 
- [x] BibCheck (@egabancho)
- [x] knowledge (@hachreak)
- [x] BibMerge (@jmartinm)
- [x] BibEdit (@jmartinm)
- [x] PDFChecker (@kaplun)
- [x] hepdata (@kaplun)
- [ ] plotextractor (@jalavik)
[...]

PR #3253, opened 2015-06-10
pdfchecker: module removal

kaplun:
* INCOMPATIBLE Removes legacy pdfchecker module together with ported
  'pdfchecker' module.  (addresses #3233)


kaplun:
I guess I should leave intact the upgrade recipe that at some point was creating the tables needed by pdfchecker.

jirikuncar:
> I should leave intact the upgrade recipe

Yes. You can also remove the model so for new installations the tables are not created. When new module is ready it can do migration if the old tables still exists in db (`op.has_table(...)`).

kaplun:
OK. Then I think it is got for integration.

jirikuncar:
@kaplun please rebase ... thanks

kaplun:
...aaaaaand done.

PR #3254, opened 2015-06-10
utils: gnd identifier support addtion

bouzlibop:
* Adds support for gnd identifier. (addresses zenodo/zenodo#217)

### DIFF
--- a/invenio/utils/persistentid.py
+++ b/invenio/utils/persistentid.py
@@ -1,5 +1,5 @@
 # This file is part of Invenio.
-# Copyright (C) 2013, 2014 CERN.
+# Copyright (C) 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -18,6 +18,7 @@
 """Persistent identifier validation and scheme detection."""
 
 import re
+
 from six.moves.urllib.parse import urlparse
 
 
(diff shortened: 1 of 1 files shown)


bouzlibop:
cc @lnielsen

jirikuncar:
@lnielsen can we apply it to all `maint-2.x` branches?

invenio/utils/persistentid.py
```diff
@@ -297,6 +303,18 @@ def is_pmcid(val):
     return pmcid_regexp.match(val)
 
 
+def is_gnd(val):
+    """Test if argument is a GND Identifier."""
+    if val.startswith('http://d-nb.info/gnd/'):
```
lnielsen: Can you assign this string to a module level variable (similar to the regexs) and use that variable instead (i.e. to ensure any change will be propagated correctly).

invenio/utils/persistentid.py
```diff
@@ -316,6 +334,7 @@ def is_pmcid(val):
     ('ean8', is_ean8),
     ('istc', is_istc),
     ('pmid', is_pmid),
+    ('gnd', is_gnd),
```
lnielsen: Hmm, probably we should take out ``pmid`` as it matches integers.

invenio/utils/persistentid.py
```diff
@@ -371,6 +390,14 @@ def normalize_orcid(val):
     return "-".join([val[0:4], val[4:8], val[8:12], val[12:16]])
 
 
+def normalize_gnd(val):
+    """Normalize a GND identifier."""
+    if val.startswith("http://d-nb.info/gnd/"):
```
lnielsen: Here you can assume you have a valid GND identifier. Current a valid GND can only have one form, so I'm about to think we don't need the normalise function

lnielsen:
@jirikuncar We can, but it's a new feature so should probably just go to master?

invenio/utils/persistentid.py
bouzlibop: Isn't that already done in https://github.com/inveniosoftware/invenio/pull/3254/files#diff-f3e79904dc66a9f6d0ebca1960b25154L336?

jirikuncar:
@blixhavn do you need it in `maint-2.0`?

lnielsen:
Closing request as it was integrated in inveniosoftware/idutils

Issue #3211, opened 2015-06-03
PostgreSQL: create equivalent function get_table_update_time() and get_table_status_info() 


tiborsimko:
- stems from #3079 roadmap issue

tiborsimko:
Superseded by Invenio 3.

PR #3268, opened 2015-06-12
WIP legacy: create PostgreSQL equivalent of functions

jacquerie:
Addresses #3211; for now I'm only touching MySQL stuff. Implementing these queries in PostgreSQL is harder than I thought, but their counterparts needed fixing as well.


jacquerie:
[This answer](http://dba.stackexchange.com/a/58246) contains the state of the art for creating the function `get_table_update_time` in PostgreSQL. We can either:

1. Have an inaccurate but simple solution using tables in the `pg_catalog`.
2. Have an accurate table of update times created by a trigger, which has the (terrible!) effect of serializing every transaction.
3. Use an external process that listens to modifications and writes them lazily to a table.
4. Extend PostgreSQL to add this functionality.

Given the alternatives, and the [relatively trivial use cases](https://github.com/inveniosoftware/invenio/search?utf8=%E2%9C%93&q=get_table_update_time) we have for this function, I think alternative 1 is preferable.

jirikuncar:
We should try to completely remove usage of `get_table_update_time` instead of porting it.

### Proposal

1. Identify tables that has to be watched for inserts/updates/deletes.
2. Use `invenio.ext.cache:cache` to keep last modification time for these tables.
3. Hook to SQLAlchemy events `after_insert`, `after_update`, and `after_delete` (and make sure that these tables are not modified via `run_sql`).
4. Amend data cachers ...

kaplun:
At the same time we can really investigate if the functionality exploiting `get_table_update_time()` is still making sense nowadays and otherwise remove it.

jacquerie:
In fact, the other function `get_table_status_info` can probably be removed. [Its usage](https://github.com/inveniosoftware/invenio/search?utf8=%E2%9C%93&q=get_table_status_info&type=Code) is limited, and it clearly [contains a bug](https://github.com/inveniosoftware/invenio/blob/0127513fd85bf5e9eeb8281957e10e2bb343ba10/invenio/legacy/dbquery_mysql.py#L482-L512): if several tables matched the `LIKE` clause, only the information of the last is returned.

jacquerie:
Actually, I decided to port `get_table_status_info` by splitting it into two functions, the other called `table_exists`. In fact, sometimes that function was just used to check if a table creation succeded or not, and its data was discarded.

The function was only tested locally, a test will be committed later.

invenio/legacy/dbquery_mysql.py
```diff
@@ -545,3 +396,15 @@ def real_escape_string(unescaped_string, run_on_slave=False):
     connection_object = _db_login(dbhost)
     escaped_string = connection_object.escape_string(unescaped_string)
     return escaped_string
+
+
+def table_exists(table_name, run_on_slave=False):
```
jirikuncar: have a look at `op.has_table`

invenio/legacy/dbquery_mysql.py
```diff
@@ -447,69 +323,44 @@ def log_sql_query(dbhost, sql, param=None):
...
+        FROM information_schema.tables
+        WHERE table_name LIKE %s and table_schema=%s''',
+        (table_name, cfg['CFG_DATABASE_NAME']), run_on_slave=run_on_slave))
+
+    return str(result[0])
 
 
-def get_table_status_info(tablename, run_on_slave=False):
-    """Return table status information on TABLENAME.
+def get_table_status_info(table_name, run_on_slave=False):
```
jirikuncar: IMHO this function is not needed anymore

kaplun: It's still used by BibIndex which is not yet deprecated:
```grep
 <master> (<detached:remotes/origin/pr/3268=be77d docs/ invenio/) skaplun@PCSK4 ~/.virtualenvs/master/src/invenio➔ git grep get_table_status_info
invenio/legacy/bibindex/adminlib.py:from invenio.legacy.dbquery import (run_sql, get_table_status_info,
invenio/legacy/bibindex/adminlib.py:        forward_table_status_info = get_table_status_info(
invenio/legacy/bibindex/adminlib.py:        reverse_table_status_info = get_table_status_info(
invenio/legacy/bibindex/scripts/bibstat.py:    from invenio.legacy.dbquery import get_table_status_info
invenio/legacy/bibindex/scripts/bibstat.py:    table_info = get_table_status_info(tablename)
invenio/legacy/dbquery_mysql.py:def get_table_status_info(table_name, run_on_slave=False):
invenio/legacy/dbquery_postgresql.py:def get_table_status_info(table_name, run_on_slave=False):
```

So what do you suggest @jacquerie should do?

jirikuncar: see https://github.com/inveniosoftware/invenio/pull/3268/files#diff-d00eda3b8584aa0f27adab861939a087R2724

It's used only to guess if table exits ..

invenio/legacy/dbquery_mysql.py
jacquerie: Not relevant anymore since 9d7151.

invenio/legacy/dbquery_mysql.py
jacquerie: Nope, it's also used for some statistics printing. But it's stuff that's going to get killed with BibIndex anyway, so this function can be removed.

jacquerie:
Tables that are watched for modifications:
* `accARGUMENT`
* `accROLE_accACTION_accARGUMENT`
* `collection`
* `collection_collection`
* `collection_name`
* `fieldname`
* `idxINDEX`
* `knwKBRVAL`

(I'm jotting this down for tomorrow)

jacquerie:
Point 3 of [Jiri's plan](https://github.com/inveniosoftware/invenio/pull/3268#issuecomment-111970745) looks hard: with the exception of `collection_name`, `idxINDEX` and `knwKBRVAL`, all other tables are modified through `run_sql` statements.

jirikuncar:
@hachreak is working on `acc*` tables. We can also create simple *Flask-Admin* interface for manipulating `collection*` tables.

jacquerie:
As far as I can tell, the `acc*` tables are still in flight: https://github.com/inveniosoftware/invenio/pull/3328. I couldn't find where (or if) the `collection` tables are being ported to `Flask-Admin`.

On the other hand, I see that none of the tables I mentioned before are now modified by a `run_sql` statement. So, as soon as models for these tables are implemented I can continue working on this.

tiborsimko:
Superseded by Invenio 3.

PR #3270, opened 2015-06-15
global: instance path and static folder config

crepererum:
Adds env variable to config the Invenio instance path and static folder. Uses this to improve the docker volume layout.

### DIFF
--- a/docs/developers/docker.rst
+++ b/docs/developers/docker.rst
@@ -66,6 +66,29 @@ data, run:
 
     docker-compose -f docker-compose-dev.yml rm -f
 
+Interactive sessions
+--------------------
+
+For interactive sessions inside the container simply attach a shell to the
+running instance:
+
+.. code-block:: shell
+
+    docker exec -it invenio_web_1 bash
+
+.. note::
+    Some tools (mostly ncurses-based ones) require the ability to detect the
+    used terminal. To enable this, set the `TERM` environment variable to the
+    correct value as a first command inside the container, e.g. by running:
+
+    .. code-block:: shell
+
+        export TERM=xterm
+
+.. note::
+    Do not forget to run `bibsched` to schedule the processing of uploaded
+    records. You can put it into automatic mode if you like.
+
 Debugging
 ---------
 
@@ -75,6 +98,13 @@ available via webinterface.
 
 .. image:: /_static/docker-debug.png
 
+You can also insert a tracepoint into the code to start an interactive debugger
+session:
+
+.. code-block:: python
+
+    import ipdb; ipdb.set_tracepoint()
+
 Furthermore you can debug MySQL at `localhost:3306`
 and Redis at `localhost:6379`. You might want to use flower_ for celery
 debugging and analysis as well. Just run the following command to open the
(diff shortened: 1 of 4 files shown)


docs/developers/docker.rst
```diff
@@ -61,6 +61,29 @@ data, run:
...
+    docker exec -it invenio_web_1 bash
+
+.. note::
+    Some tools (mostly ncurses-based ones) require the ability to detect the
+    used terminal. To enable this, set the `TERM` environment variable to the
+    correct value as a first command inside the container, e.g. by running:
+
+    .. code-block:: shell
+
+        export TERM=xterm
```
dset0x: I wonder what prompted you to make this addition. Having to redefine TERM implies issues with the way the environment is set up the majority of times.

invenio/base/factory.py
```diff
@@ -155,22 +156,44 @@ def configure_warnings():
...
+    will be used which is the uppercase version of the app name, excluding
+    any non-alphabetic ('[^A-Z]') characters.
+
+    If `instance_path` is `None`, the `<PREFIX>_INSTANCE_PATH` environment
+    variable will be used. If that one does not exist, a path inside
+    `sys.prefix` will be used.
+
+    If `static_folder` is `None`, the `<PREFIX>_STATIC_FOLDER` environment
+    variable will be used. If that one does not exist, a path inside the
+    detected `instance_path` will be used.
```
jirikuncar:
```
.. versionadded:: 2.1
   `static_folder` ...

```

docs/developers/docker.rst
crepererum: By default TERM is not defined inside the interactive session (bash inside the container). So e.g. `bibsched` won't even start because curses is not able to detect the terminal. The variable is only for the interactive session and is only populates to child progresses of bash, not to other processes inside the container or (especially) to professes outside the container.

invenio/base/factory.py
```diff
@@ -155,22 +156,44 @@ def configure_warnings():
...
-    # Force instance folder to always be located in under system prefix
-    instance_path = instance_path or os.path.join(
-        sys.prefix, 'var', app_name + '-instance'
-    )
+    # Prefix for env variables
+    env_prefix = re.sub('[^A-Z]', '', app_name.upper())
+
+    # Detect instance path
+    instance_path = instance_path or \
+        os.getenv(env_prefix + '_INSTANCE_PATH') or \
```
jirikuncar: We should discuss potential security issues when using OS environment variables.

--
cc @lnielsen @tiborsimko @jalavik

crepererum: ENV vars get only propagated to child processes. Possible ways are:

1. Attacker own the parent process.
   - Discussion ends here.
2. Attacker sets variables via (u)WSGI-script:
   - You can only set certain variables, if the web server (e.g. Apache) is not buggy. See https://www.python.org/dev/peps/pep-0333/#environ-variables
   - If the server is buggy, you could also set `LD_PRELOAD` or `PYTHONHOME`, discussion ends.

BTW: environment variables are used in different places all over the system. For an example what you could do when you can control these variables, just see `man python2`>`ENVIRONMENT VARIABLES`.

jirikuncar:
:shipit:

crepererum:
:+1:

crepererum:
PING

Issue #3218, opened 2015-06-03
PostgreSQL: foreign key problems when delete records


hachreak:
In MySQL the problem does not occur because the engine MyISAM doesn't support foreign keys.

jirikuncar:
Do you have a traceback?

tiborsimko:
Superseded by Invenio 3.

PR #3275, opened 2015-06-16
deposit: fix deleting function

hachreak:
* Fixes the delete function to support cascading deletion on PostgreSQL.
  (addresses #3218)

* PEP8/257 code style improvements.

### DIFF
--- a/invenio/modules/deposit/testsuite/test_deposit_models.py
+++ b/invenio/modules/deposit/testsuite/test_deposit_models.py
@@ -20,11 +20,16 @@
 """Unit tests for the Deposit models."""
 
 from flask_registry import RegistryError
-from invenio.testsuite import make_test_suite, run_test_suite, InvenioTestCase
+
+from invenio.testsuite import InvenioTestCase, make_test_suite, run_test_suite
 
 
 class DepositionTest(InvenioTestCase):
+
+    """Test."""
+
     def setUp(self):
+        """Test."""
         from invenio.modules.deposit.models import DepositionType
         from invenio.modules.deposit.registry import deposit_types, \
             deposit_default_type
@@ -50,6 +55,7 @@ class DepositionTest(InvenioTestCase):
         deposit_default_type.register(DefaultType)
 
     def test_create(self):
+        """Test."""
         from invenio.ext.login.legacy_user import UserInfo
         from invenio.modules.deposit.models import Deposition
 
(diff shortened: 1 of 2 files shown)


invenio/modules/deposit/testsuite/test_deposit_models.py
```diff
@@ -57,9 +63,13 @@ def test_create(self):
...
         assert Deposition.get(d.id).type == self.DefaultType
-        d = Deposition.create(user, type=self.AnotherType)
-        assert d.type == self.AnotherType
-        assert Deposition.get(d.id).type == self.AnotherType
+        d2 = Deposition.create(user, type=self.AnotherType)
+        assert d2.type == self.AnotherType
+        assert Deposition.get(d2.id).type == self.AnotherType
+
+        # remove the records
+        Deposition.delete(d)
```
lnielsen: Could you please add a test case that checks if both all the underlying BibWorkflowObjects (multiple) and the associated Workflow object has been removed and nothing more than that.

hachreak: On which objects I have to test the existence/not existence?

lnielsen: Deposition is just a wrapper a around a ``BibWorkflowObject`` (workflows module) so you should test something like:

```python
bwl_d2_count = BibWorkflowObject.query.filter_by(uuid=d2.workflow_object.id_workflow).count()
d.delete()
assert Workflow.query.filter_by(uuid=d.workflow_object.id_workflow).count() == 0
assert  BibWorkflowObject.query.filter_by(id_workflow=d.workflow_object.id_workflow).count() == 0
```
and that no other objects where deleted, e.g.:

```python
assert Workflow.query.filter_by(uuid=d2.workflow_object.id_workflow).count() == 1
assert BibWorkflowObject.query.filter_by(uuid=d2.workflow_object.id_workflow).count() == bwl_d2_count
#...
d2.delete()
```

hachreak: something like this? 

https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/workflows/testsuite/test_workflows_model.py#L28

lnielsen: Similar except that above tests Deposit/Workflow integration, and in the example above you only have one BibWorkflowObject, whereas you can normally have many.

It's from experience that this is a soft spot between the two modules.

Issue #3214, opened 2015-06-03
global: remove MySQL specific legacy code


tiborsimko:
- legacy: remove mysql specific codes (see `git grep MySQLdb`, `git grep "SHOW TABLES"`, `git grep "CHECK TABLE"`, `git grep "RLIKE"`, `git grep "PRIORITY"`, `git grep DATE_FORMAT`, `regexp`)

- stems from #3079 roadmap issue

jacquerie:
There's also `SHOW CREATE TABLE`, which was used in the past in `invenio/modules/upgrader` to build the upgrade recipe. Should these be migrated as well?

jirikuncar:
@jacquerie IMHO you can ignore old upgrade recipes.

jacquerie:
Ok! :+1:

hachreak:
We need also find a equivalent for postgresql of `git grep "ON DUPLICATE KEY UPDATE"`

hachreak:
The list:
- [ ] `git grep "ON DUPLICATE KEY UPDATE"`
- [ ] `git grep "SELECT LAST_INSERT_ID()"`
- [ ] `git grep UNIX_TIMESTAMP`
- [ ] `git grep "SLAVE STATUS"`
- [ ] `git grep "INSERT IGNORE"`
- [ ] `git grep inet_aton`
- [ ] `git grep MySQLdb`: waiting #3326
- [ ] `git grep "SHOW TABLES"`
- [ ] `git grep "CHECK TABLE"`
- [ ] `git grep "RLIKE"`
- [ ] `git grep "PRIORITY"`
- [ ] `git grep DATE_FORMAT`
- [ ] `regexp`
- [ ] `git grep SHOW CREATE TABLE`

jirikuncar:
Invenio 3 is using only SQLAlchemy database independent models.

PR #3278, opened 2015-06-16
global: fix PostgreSQL compatibility

hachreak:
* Fixes run_sql query to obtain PostgreSQL compatibility.
  (addresses #3089) (addresses #3214)

### DIFF
--- a/invenio/base/scripts/demosite.py
+++ b/invenio/base/scripts/demosite.py
@@ -22,21 +22,23 @@
 from __future__ import print_function
 
 import os
+
 import sys
+
 import warnings
-from itertools import count
 
 import pkg_resources
 
+from itertools import count
+
 from invenio.base.utils import run_py_func
 from invenio.ext.script import Manager
+from invenio.modules.scheduler.models import SchTASK
 
 
 warnings.warn("Use of `inveniomanage demosite populate` is being deprecated. "
               "Please use `uploader` module to insert demo records.",
               PendingDeprecationWarning)
-
-
 
 
 manager = Manager(usage=__doc__)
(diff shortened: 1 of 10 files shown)


hachreak:
ping @jirikuncar :shipit:

invenio/base/scripts/demosite.py
```diff
@@ -94,10 +96,9 @@ def populate(packages=[], default_data=True, files=None,
     from werkzeug.utils import import_string
     map(import_string, packages)
 
-    from invenio.ext.sqlalchemy import db
     print(">>> Going to load demo records...")
-    db.session.execute("TRUNCATE schTASK")
```
jirikuncar: `TRUNCATE` was resetting `autoincrement` column. Can we do the same or use the low-level task submission?

hachreak: the only way I know is to use the [truncate_table](https://github.com/inveniosoftware/invenio/blob/9d71513860f25751d8378c5e51707609db4c0600/invenio/legacy/dbquery_postgresql.py#L432) function.

jirikuncar: Or try the low-level submission for bibsched tasks (maybe @egabancho can help a bit here).

invenio/modules/comments/api.py
```diff
@@ -411,9 +412,9 @@ def check_user_can_vote(cmt_id, client_ip_address, uid=-1):
     cmt_id = wash_url_argument(cmt_id, 'int')
     client_ip_address = wash_url_argument(client_ip_address, 'str')
     uid = wash_url_argument(uid, 'int')
-    query = """SELECT id_cmtRECORDCOMMENT
-               FROM cmtACTIONHISTORY
-               WHERE id_cmtRECORDCOMMENT=%s"""
+    query = """SELECT "id_cmtRECORDCOMMENT"
```
PXke: Shouldn't all this part be rewrite with SQLAlchemy ?

hachreak: Yes, I know. :)
But, we are so close to have a "usable" version with PostgreSQL.
So, I proposed to parallel working on:

 1. Remove raw sql
 2. Make invenio compatible as possible with PostgreSQL

Because to wait of completely remove all raw sql, require to much time and we can instead start playing soon with PostgreSQL.

PXke: Ok :) I makes sense !

PR #3280, opened 2015-06-16
jasmine: app config in bundles

jochenklein:


jirikuncar:
@jochenklein I would try to reformulate the commit message in following way:
```
jasmine: support for app config in bundles

* BETTER Allows using variables from application config for
  building asset bundles.

...
```

PR #3289, opened 2015-06-18
workflows: partial undo of e8f6551

### LINKED ISSUE #3288: workflows: UUID upgrade problem
The change to using UUID type and the following upgrade script introduced in https://github.com/inveniosoftware/invenio/commit/e8f6551d9787ceec99248b2b0ad04b8f22a0d8b4 causes a problem with truncated values for the existing values - essentially corrupting the data (observed locally):
```
>>> Applying workflows_2015_06_05_resize_uuid_columns (Resize uuid columns on workflows tables.)
>>> workflows_2015_06_05_resize_uuid_columns: WARNING: Data truncated for column 'uuid' at row 1
>>> workflows_2015_06_05_resize_uuid_columns: WARNING: Data truncated for column 'uuid' at row 2
...
```
Before the uuid column was `db.String(36)` and a value included hyphens `00011d78-fd76-11e4-b9a6-ecf4bb24`. With the UUID type on MySQL, the hyphens are not considered and the length of the CHAR type is set to 32. During the upgrade the data is still stored with hyphens causing the truncation. 

I noticed this in the output of the upgrade, the corrupt IDs then cause a range of [troubles](https://gist.github.com/anonymous/53c521015b0c34685173) post-upgrade.

I suggest we either (a) change the MySQL type to `CHAR(36)`, or (b) remove hyphens from the data during upgrade.

EDIT: Considering changing a primary key is never a good idea, I suggest (a).

@hachreak @jirikuncar

hachreak:
* Uses of strings to represent workflow uuid. (closes #3288)

* Adds casting to string when someone set a uuid.

### DIFF
--- a/invenio/modules/workflows/upgrades/workflows_2015_06_05_resize_uuid_columns.py
+++ b/invenio/modules/workflows/upgrades/workflows_2015_06_05_resize_uuid_columns.py
@@ -27,25 +27,15 @@ depends_on = [u'workflows_2014_08_12_initial']
 
 def info():
     """Info message."""
-    return "Resize uuid columns on workflows tables."
+    return "Resize id_object column on bwlWORKFLOWLOGGING table."
 
 
 def do_upgrade():
     """Implement your upgrades here."""
-    with op.batch_alter_table("bwlWORKFLOW") as batch_op:
-        batch_op.alter_column(
-            column_name='uuid',
-            type_=db.UUID(), nullable=False
-        )
-    with op.batch_alter_table("bwlOBJECT") as batch_op:
-        batch_op.alter_column(
-            column_name='id_workflow',
-            type_=db.UUID(), nullable=True
-        )
     with op.batch_alter_table("bwlWORKFLOWLOGGING") as batch_op:
         batch_op.alter_column(
             column_name='id_object',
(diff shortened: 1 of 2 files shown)


jalavik:
@hachreak Awesome. Thanks!

hachreak:
you are welcome! :)

jirikuncar:
@hachreak and what about the upgrade recipe?

hachreak:
I modified also the upgrade recipe to change only the length of `bwlWORKFLOWLOGGING.id_object`.
https://github.com/inveniosoftware/invenio/pull/3289/files#diff-3812489ac7b9a6a9285fddee2fe3ba0fR33
It's ok? @jirikuncar

invenio/modules/workflows/upgrades/workflows_2015_06_05_resize_uuid_columns.py
```diff
@@ -27,25 +27,15 @@
 
 def info():
     """Info message."""
-    return "Resize uuid columns on workflows tables."
+    return "Resize id_object columns on bwlWORKFLOWLOGGING table."
```
jirikuncar: `s/columns/column/`

PR #3291, opened 2015-06-19
docker: fix tests

crepererum:
Fixes tests when running under docker dev setup.  Also fixes a problem with the way we are parsing config variables from ENV.


jirikuncar:
@crepererum thanks. It looks good to me. :shipit:

PR #3292, opened 2015-06-19
accounts: panel_start macro extra class parameter

ioannistsanaktsidis:
Signed-off-by: Ioannis Tsanaktsidis <ioannis.tsanaktsidis@cern.ch>
Reviewed-by: Javier Martin Montull <javier.martin.montull@cern.ch>

### DIFF
--- a/invenio/modules/accounts/templates/accounts/settings/helpers_base.html
+++ b/invenio/modules/accounts/templates/accounts/settings/helpers_base.html
@@ -17,11 +17,10 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 #}
 
-
-{%- macro panel_start(title, icon="", btn_text='', btn='', btn_icon='', btn_href='', btn_class='btn-default', btn_name='', id="", with_body=True) %}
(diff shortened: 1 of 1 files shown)


invenio/modules/accounts/templates/accounts/settings/helpers_base.html
```diff
@@ -18,10 +18,10 @@
 #}
 
 
-{%- macro panel_start(title, icon="", btn_text='', btn='', btn_icon='', btn_href='', btn_class='btn-default', btn_name='', id="", with_body=True) %}
+{%- macro panel_start(title, icon="", panel_extra_class="", btn_text='', btn='', btn_icon='', btn_href='', btn_class='btn-default', btn_name='', id="", with_body=True) %}
```
jmartinm: Sorry what I meant was to add ```panel_extra_class="panel-default"``` here in the macro parameters as default value.

Like this everything will behave in the same way for people using this macro but still it allows to set something different than ```panel-default``` (hardcoded in this version)

jirikuncar: can you add the `panel_extra_class` argument at the end?

giannistsan: Yes, done!

2015-06-19 11:27 GMT+02:00 Jiri Kuncar <notifications@github.com>:

PR #3294, opened 2015-06-22
WIP celery: add support for uuid serialization

dset0x:
* Introduces a UUID serializer and deserializer and registers it to
  kombu for use, as `msgpack_append`. Celery is then configured to use
  this instead of `msgpack`.

Note: This resulted while discussing solutions regarding #3288


PXke:
:+1:

PR #3302, opened 2015-06-24
global: `run_py_func` list argv

dset0x:
* Have `run_py_func` always set `sys.argv` to a list, as some
  applications may want to manipulate it.

### DIFF
--- a/invenio/base/utils.py
+++ b/invenio/base/utils.py
@@ -300,7 +300,7 @@ def run_py_func(manager_run, command_line, passthrough=False):
         else:
             sys.argv = shlex.split(command_line)
     else:
-        sys.argv = command_line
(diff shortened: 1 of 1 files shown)


jirikuncar:
@dset0x do you have any use-case/example where this is needed?

dset0x:
@jirikuncar I don't believe there is a case in invenio where this has been causing a problem so far, but argv can be assumed to be of type `list`, so this is a bug in setting up the environment if someone passes a tuple or an iterator for example.

jirikuncar:
@dset0x ok. :shipit:

jirikuncar:
@dset0x shall we backport it to any `maint-2.x`?

jirikuncar:
Please move it to `invenio-utils` repo if needed.

PR #3304, opened 2015-06-24
installation: allow only valid names for database

### LINKED ISSUE #3303: docs: db setup breaks for non-master
`INSTALL.rst` states

    (invenio)$ inveniomanage config set CFG_DATABASE_NAME $BRANCH
    (invenio)$ inveniomanage config set CFG_DATABASE_USER $BRANCH

which will fail if `$BRANCH` is e.g. `maint-2.1` due to the - in the branch name.

aw-bib:
INSTALL.rst used $BRANCH for the database name which might end up to be invalid e.g. for maint-2.1 like branches. Remove all non characters and numbers from the name so it will read maint21 in this example.

Closes #3303

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -460,8 +460,8 @@ the following commands.
 
     (invenio)$ inveniomanage config set CFG_EMAIL_BACKEND flask_email.backends.console.Mail
     (invenio)$ inveniomanage config set CFG_BIBSCHED_PROCESS_USER $USER
-    (invenio)$ inveniomanage config set CFG_DATABASE_NAME $BRANCH
-    (invenio)$ inveniomanage config set CFG_DATABASE_USER $BRANCH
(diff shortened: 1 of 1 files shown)


INSTALL.rst
```diff
@@ -460,8 +460,8 @@ the following commands.
 
     (invenio)$ inveniomanage config set CFG_EMAIL_BACKEND flask_email.backends.console.Mail
     (invenio)$ inveniomanage config set CFG_BIBSCHED_PROCESS_USER $USER
-    (invenio)$ inveniomanage config set CFG_DATABASE_NAME $BRANCH
-    (invenio)$ inveniomanage config set CFG_DATABASE_USER $BRANCH
+    (invenio)$ inveniomanage config set CFG_DATABASE_NAME `echo $BRANCH | sed -e 's/[^A-Za-z0-9]//g'`
```
jirikuncar: can we create better some cleaned name after https://github.com/inveniosoftware/invenio/pull/3304/files#diff-26dee18463b5da190f343abcc74cf708R343?

--
PS: Can you rebase this patch for `maint-2.0`?

PR #3305, opened 2015-06-24
WIP: Z39

PXke:
Proposition for legacy integration of Z39


PXke:
Hello @tiborsimko all feedbacks are welcome. It is our first time we develop a new module for legacy.

modules/bibz39/Makefile
```diff
@@ -0,0 +1,646 @@
+# Makefile.in generated by automake 1.14.1 from Makefile.am.
```
kaplun: @PXke you should not commit the `Makefile(.in)` files, but only `Makefile.am`. The former are generated from the latter by `automake` stack.

requirements.txt
```diff
@@ -29,3 +29,5 @@ redis==2.9.0
 nydus==0.10.6
 Cerberus==0.5
 matplotlib
+git+https://github.com/asl2/PyZ3950.git
```
kaplun: PyZ3950 is on pypi and latest version corresponds to the github master. So you can simply put `PyZ3950==2.04`

modules/bibz39/Makefile
PXke: Thank you !

requirements.txt
PXke: Thank you !

kaplun:
Hi! Is this still a WIP?

PXke:
Sorry @kaplun but yes, QA failed on our side so there is still work done on this module. It should be finished today or tomorrow though as we are reaching the end of the process.

kaplun:
No problem! Thanks for feedback :smile_cat:

PR #3308, opened 2015-06-25
legacy: replaces DatabaseError with IntegrityError in bibtasks

blixhavn:
* BibIndex and BibRank uses `try` to test `insert`
  when a conflict may arise. After 2.1, this fails
  as the exception thrown is now an IntegrityError

### DIFF
--- a/invenio/legacy/bibindex/engine.py
+++ b/invenio/legacy/bibindex/engine.py
@@ -98,6 +98,7 @@ from invenio.utils.serializers import serialize_via_marshal, \
     deserialize_via_marshal
 
 from sqlalchemy.exc import DatabaseError
+from MySQLdb import IntegrityError
 
 from .engine_utils import get_index_id_from_index_name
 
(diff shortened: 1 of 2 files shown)


jirikuncar:
Please amend commit message as follows:
```
legacy: fix database exception in bibtasks
    
* FIX BibIndex and BibRank uses `try` to test `insert` when a conflict
  may arise, however this fails as the exception thrown is now an
  `IntegrityError`.
    
```

invenio/legacy/bibindex/engine.py
```diff
@@ -98,6 +98,7 @@
     deserialize_via_marshal
 
 from sqlalchemy.exc import DatabaseError
+from MySQLdb import IntegrityError
```
jirikuncar: can you check if you can use `sqlalchemy.exc.IntegrityError` instead?

invenio/legacy/bibindex/engine.py
```diff
@@ -982,7 +983,8 @@ def update_index(self, index_id, recID_low, recID_high):
                        (recID, serialize_via_marshal(to_serialize))) # kwalitee: disable=sql
             try:
                 run_sql("INSERT INTO %s (id_bibrec,termlist,type) VALUES (%%s,%%s,'CURRENT')" % wash_table_column_name(virtual_tab_name), (recID, serialize_via_marshal([]))) # kwalitee: disable=sql
-            except DatabaseError:
+            except IntegrityError:
```
jirikuncar: shall we keep `(DatabaseError, IntegrityError)`?

blixhavn: I guess it wouldn't hurt

invenio/legacy/bibindex/engine.py
blixhavn: I've checked, it doesn't work. Probably because as you can see below, the exception comes from the legacy function `run_sql`, and not SQLalchemy.

PR #3314, opened 2015-06-26
accounts: userEXT model upgrade

### LINKED ISSUE #3290: accounts: userEXT id as VARBINARY
In the userEXT table, the ```id``` is created as ```db.VARBINARY```

https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/accounts/models.py#L426

That seems to be the only occurrence of ```db.VARBINARY``` in Invenio and does not make much sense as that column holds ```Strings```. I just realised about it because I am having problems interacting with that column with an external MySQL editor like MySQL Workbench.

Would it make sense to change it to ```db.String``` ?

jmartinm:
* NOTE Changes id column from VARBINARY to String. (closes #3290)

### DIFF
--- a/invenio/modules/accounts/upgrades/accounts_2015_06_26_userext.py
+++ b/invenio/modules/accounts/upgrades/accounts_2015_06_26_userext.py
@@ -1,0 +1,54 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Invenio.
+# Copyright (C) 2015 CERN.
+#
+# Invenio is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Change id column from VARBINARY to String."""
+
+from invenio.ext.sqlalchemy import db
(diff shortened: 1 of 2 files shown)


invenio/modules/accounts/upgrades/accounts_2015_06_26_userext.py
```diff
@@ -0,0 +1,52 @@
...
+
+from invenio.ext.sqlalchemy import db
+from invenio.modules.upgrader.api import op
+
+depends_on = [u'accounts_2015_03_06_passlib']
+
+
+def info():
+    """Info message."""
+    return "Change id column from VARBINARY to String."
```
jirikuncar: usually it's good idea move this line as a file docstring and refer to it as `__doc__`.

jmartinm:
Changed to ```return __doc__```

jirikuncar:
What about?
```
* BETTER Improves compatibility with PostreSQL by changing 'id' column of 'userEXT' table from VARBINARY to String.  (closes #3290)
```

hachreak:
@jmartinm @jirikuncar it's ready the PR? :)

invenio/modules/accounts/upgrades/accounts_2015_06_26_userext.py
```diff
@@ -0,0 +1,54 @@
...
+# Invenio is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
+"""Change id column from VARBINARY to String."""
```
tiborsimko: A general comment for the future: when a user sees an upgrade recipe message saying:

- Change id column from VARBINARY to String.

that it may not be very informative.  It's better to be more specific and at least repeat the table name.  Here are some examples of the messages we used in the past in these situations:

- New xtrJOB.last_recid column
- Larger values allowed for seqSTORE.seq_value

PR #3317, opened 2015-06-26
deposit: datetimepicker version bump

jmartinm:
* Updates eonasdan-bootstrap-datetimepicker to 4.14.30

### DIFF
--- a/invenio/modules/deposit/bundles.py
+++ b/invenio/modules/deposit/bundles.py
@@ -35,7 +35,7 @@ js = Bundle(
         "plupload": "latest",
         "ckeditor": "latest",
         "flight": "latest",
-        "eonasdan-bootstrap-datetimepicker": "3.1.3",
+        "eonasdan-bootstrap-datetimepicker": "4.14.30",
     }
 )
 
--- a/invenio/modules/deposit/static/js/deposit/form.js
+++ b/invenio/modules/deposit/static/js/deposit/form.js
@@ -57,7 +57,6 @@ define(function(require, exports, module) {
         autocomplete_url: "",
         datepicker_options: {
           format: "YYYY-MM-DD",
-          pickTime: false,
         },
 
         // Selectors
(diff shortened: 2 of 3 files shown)


jmartinm:
We were having the following problem https://github.com/inspirehep/inspire-next/issues/247

Updating datetimepicker fixes it.

jmartinm:
I have removed `pickTime` option as it is no longer supported https://github.com/Eonasdan/bootstrap-datetimepicker/issues/864#issuecomment-76004406

jirikuncar:
@jmartinm please check milestone where this patch should be applied.

- IMHO we can't really merge it to `maint-2.0` as it's breaking the UI. Shall we instead pin version of some other library?
- It should contain INCOMPATIBLE label in commit message.
- There is a missing dot at the end of the bullet point.

jmartinm:
> IMHO we can't really merge it to maint-2.0 as it's breaking the UI. Shall we instead pin version of some other library?

Does it break the UI? Did not try it but with the change in the code that this commit introduces (removal of `picktime`) there should be no problem in principle.

jmartinm:
> There is a missing dot at the end of the bullet point.

I'll fix that. Shouldn't the kwalitee script capture that?

jirikuncar:
> Shouldn't the kwalitee script capture that?

You can submit a patch :wink:

jirikuncar:
@jmartinm ping https://github.com/inveniosoftware/invenio/pull/3317#issuecomment-117215206

jmartinm:
Updated PR. It is rebased against master.

jirikuncar:
@jmartinm are we going to merge it to `maint-2.0` or just `master`?

jmartinm:
For us it does not make a difference and `master` would be simpler. I updated it because the old version had a glitch on the display.

Is there any use in merging into `maint-2.0` ?

jirikuncar:
IMHO the old version is pinned and works just fine except the special case mentioned above.

PR #3319, opened 2015-06-26
access: fix cookie creation with PostgreSQL

### LINKED ISSUE #3318: access: cookie generation fail with PostgreSQL
I discovered that, when I read a Integer from the DB (e.g. an ID), with
 - MySQL: I have a python long integer
 - PostgreSQL: I have a python integer

This [hex()](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/models.py#L144) calling returns a empty string with PostgreSQL and invalidate the cookie creation.

hachreak:
* Fixes cookie creation with PostgreSQL. (closes #3318)


hachreak:
ping @jirikuncar

Issue #3299, opened 2015-06-23
RFC plans for refactoring access module


hachreak:
- [x] Remove [admin_lib.py](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/access/admin_lib.py)
- [ ] Remove [webaccessadmin.py](https://github.com/inveniosoftware/invenio/blob/master/invenio%2Fmodules%2Faccess%2Fscripts%2Fwebaccessadmin.py) and move functionality in `manage.py` (see #3381)
- [ ] Remove not used functions from [control.py](https://github.com/inveniosoftware/invenio/blob/master/invenio%2Fmodules%2Faccess%2Fcontrol.py).
This is a list of functions candidates:
  - acc_add_argument
  - acc_delete_action
  - acc_delete_argument
  - acc_delete_argument_names
  - acc_delete_possible_actions
  - acc_delete_role_action
  - acc_delete_role_action_arguments
  - acc_delete_role_action_arguments_names
  - acc_delete_role_action_arguments_group
  - acc_find_delegated_roles
  - acc_find_possible_actions_all
  - acc_find_possible_actions_user
  - acc_find_possible_actions_user_from_user_info
  - acc_find_possible_activities (used only by `webuser.py`)
  - acc_find_user_role_actions
  - acc_get_action_details
  - acc_get_action_name
  - acc_get_action_roles (only used in legacy `webuser` and legacy `admin_lib.py`)
  - acc_get_all_roles
  - acc_get_role_actions
  - acc_get_role_definition
  - acc_get_role_details
  - acc_get_role_users
  - acc_merge_argument_groups
  - acc_split_argument_group
  - acc_verify_action
  - acc_is_role (used only in `invenio/legacy/bibdocfile/webinterface.py`)
- [ ] Move model validation from `control.py` to the `models.py`.
- [ ] Convert raw sql to SQLAlchemy code.
- [x] access: fix foreign key declaration. #3099
- [ ] Create a official API
- [ ] Create a complete testsuite

Any comments and suggestions?

cc @jirikuncar

kaplun:
Are you going to tackle the usage of `-1` to mean no arguments for an action? (which breaks foreign key usage)?

hachreak:
I already resolved the problem removing foreign key and create a ad-hoc SQLALchemy relationship.
I need more time to finish to prepare PR to fix it.
I'll add the Issue/PR to the list.

kaplun:
:+1:

Issue #3098, opened 2015-05-06
global: remove all run_sql from modules


hachreak:
Actually we have this modules that contains one or more `run_sql` (and consequently raw sql queries).

- [ ] access
- [ ] bulletin
- [ ] comments
- [ ] formatter
- [ ] indexer
- [ ] messages
- [ ] records
- [ ] search
- [ ] sequencegenerator

One of the next goal should be convert them completely.

jirikuncar:
Invenio 3 is using only SQLAlchemy models.

PR #3328, opened 2015-07-01
access: general module refactoring

hachreak:
* INCOMPATIBLE Removes legacy admin interface. (addresses #3233)

* Improves access models with delete, update, count, exists, factory
  functions, smart getter and setter and cascade actions.
  (addresses #3299)

* Deprecates most of the raw sql. (addresses #3098)

* Adds a testsuite for the models.


invenio/modules/access/api.py
```diff
@@ -0,0 +1,238 @@
...
+            models.AccAuthorization.argumentlistid > 0,
+            models.AccAuthorization.id_accACTION == db.bindparam('id_action')
+        )).join(models.AccAuthorization.argument)
+
+    id_action = db.session.query(models.AccACTION.id).filter(
+        models.AccACTION.name == name_action).scalar()
+    roles = intbitset(db.engine.execute(query_roles_without_args.params(
+        id_action=id_action)).fetchall())
+
+    if always_add_superadmin:
```
jalavik: Depending on the use cases for this function `find_possible_roles`, this way of checking or adding superadmin role should be changed as it is not reliable.

I know this is how it was in legacy code, but I have noticed some serious problems with it. First of all the database row ID for the admin role should not always be assumed to be 1, or anything for that matter, as it can change on any instance. This check should therefore not base itself on the role id valie, but instead use the name of the role `"superadmin"` if this is even needed anymore. Because I guess if the `current_user` is superadmin, there is no point in checking anymore. E.g. `current_user.is_super_admin` or `current_user.is_admin` can tell this information directly.

So I suggest to drop the `always_add_superadmin` argument overall as I guess it is mostly pointless to check the user anymore if he is indeed an admin.

hachreak: The function in API I simply moved without rewrite them.
I'm sure they need it.

jalavik: Yep! No worries, @hachreak. I just wanted to point this thing out for when it happens :dancer:

invenio/modules/access/models.py
```diff
@@ -159,13 +322,110 @@ class AccROLE(db.Model):
...
+    _firerole_def_ser = db.Column(db.iBinary, nullable=True,
+                                  name="firerole_def_ser")
+    _firerole_def_src = db.Column(db.Text, nullable=True,
+                                  name="firerole_def_src")
 
     def __repr__(self):
         """Repr."""
         return "{0.name} - {0.description}".format(self)
 
+    @db.hybrid_property
```
jirikuncar: readonly?

invenio/modules/access/models.py
```diff
@@ -222,8 +647,114 @@ class UserAccROLE(db.Model):
...
+    @session_manager
+    def update(cls, criteria, updates):
+        """Update values.
+
+        :param criteria: filter list
+        :param updates: list of update
+        """
+        objs = cls.query.filter(*criteria).all()
+        for obj in objs:
+            for (k, v) in updates.iteritems():
```
jirikuncar: six.iteritems

jirikuncar:
Follow [Invenio-Access](https://github.com/inveniosoftware/invenio-access) repo.

PR #3330, opened 2015-07-03
legacy: recreates JSON cache in bibupload

blixhavn:
* BETTER Adds `True` flag to the get_record call in bibupload, to force
  recreating JSON cache.

### DIFF
--- a/invenio/legacy/bibupload/engine.py
+++ b/invenio/legacy/bibupload/engine.py
@@ -567,7 +567,7 @@ def bibupload(record, opt_mode=None, opt_notimechange=0, oai_rec_id="", pretend=
             # does not exists (i.e. all process will try to generate and save
             # the recjson)
             from invenio.modules.records.api import get_record as _get_record
-            _get_record(rec_id)
+            _get_record(rec_id, True)
 
             # Fire record signals.
             from invenio.base import signals


jirikuncar:
@blixhavn is there any related issue?

egabancho:
Maybe we should review this more carefully, see https://github.com/inveniosoftware/invenio/commit/efe48d167cebf8ba46a0b4921d5d2f2723d1988e

cc @lnielsen

blixhavn:
I guess @egabancho is right. This was kind of a quickfix on my part – I don't know about any implications it might have. For TIND this is no longer critical, as we're staying on 2.0 for a while. It might be better to close this PR and make an issue instead, where we can discuss how this best can be solved.

audub:
Actually, this is also an issue on 2.0. At the moment, we have customers that don't see changes after using BibEdit, which means they have no good way of changing a record. It is kind of critical. Any suggestions on how to best fix this? @lnielsen maybe?

egabancho:
In 2.0 `bibreformat` https://github.com/inveniosoftware/invenio/blob/maint-2.0/invenio/legacy/bibformat/bibreformat.py takes of updating the json once it gets outdated
i.e. `bibreformat -u admin -o HB,recjson -s 5m`

audub:
@egabancho Thanks, at least it will solve our problem. Still, the user has to wait for some time for the record to be updated, which is not ideal. But we can live with it for a little while :)

jirikuncar:
One can use `bibreformat` (see https://github.com/inveniosoftware/invenio/pull/3330#issuecomment-121965034).

tiborsimko:
(See also `CFG_BIBUPLOAD_DELETE_FORMATS` that should delete `recjson` format from `bibfmt` cache as soon as a record is updated.  Hence for any first consumer asking for `recjson` after record update, a new `recjson` should be generated.  Hence there should not be any outdated recjson lying around in the first place?)

egabancho:
@tiborsimko AFAIK in Invenio 2.x `recjson` is no longer stored inside `bibfmt`. 
On `legacy` branch instead this does the trick.

tiborsimko:
@egabancho Indeed, it is not stored there anymore, so `bibreformat` would be the way to go with `maint-2.0`. But perhaps we could amend `bibupload`'s behaviour in order to delete the `recjson` format from the `record_json` table in this case in order to keep 1.2 <-> 2.0 behavioural correspondence?

PR #3332, opened 2015-07-05
docker: more complete overlay documentation

tiborsimko:
* BETTER Improves Docker documentation notably related to how to work
   with Invenio site overlays.
    
Reviewed-by: Marco Neumann <marco@crepererum.net>

### DIFF
--- a/docs/developers/docker.rst
+++ b/docs/developers/docker.rst
@@ -31,8 +31,8 @@ Install Docker_ and `Docker Compose`_. Now run:
 
 .. code-block:: shell
 
-    docker-compose -f docker-compose-dev.yml build docker-compose -f \
-        docker-compose-dev.yml up
+   docker-compose -f docker-compose-dev.yml build
+   docker-compose -f docker-compose-dev.yml up
 
 This builds and runs the docker containers. You can now connect to
 `localhost:28080` to see your Invenio installation. The `admin` user does not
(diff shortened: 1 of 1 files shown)


tiborsimko:
See also https://github.com/inveniosoftware/invenio-demosite/pull/69.

crepererum:
This was rejected the first time I've tried it. You have to set this specific config using an ENV variable.

crepererum:
Not required as of https://github.com/inveniosoftware/invenio/blob/master/docker-compose-dev.yml#L46

tiborsimko:
Indeed, the `CFG_DATABASE_HOST` bit is not needed, after rerunning. I had a case during weekend where the app couldn't connect to DB, and this fixed it; but it was perhaps caused by that `../invenio/invenio` mismatch that  mentioned at `https://github.com/inveniosoftware/invenio-demosite/pull/69#issuecomment-118839838` and that I've been tracking at the same time.

Anyway, is there an advantage in keeping `CFG_REDIS_HOSTS` in but `CFG_REDIS_HOSTS` out?  It would be nice to treat them similarly, for consistency...

crepererum:
I'm currently doing some Docker work anyway, so I will try to isolate `CFG_REDIS_HOSTS` as well.

tiborsimko:
> CFG_REDIS_HOSTS

OK, I'll then remove the `CFG_DATABASE_HOST` part and keep only the documentation amendments.

crepererum:
:+1:

PR #3337, opened 2015-07-07
pages: default page template

jalavik:
* Sets default template and avoids crashing when no template
  is specified in admin interface.

Reviewed-by: Jan Aage Lavik <jan.age.lavik@cern.ch>

### DIFF
--- a/invenio/modules/pages/admin.py
+++ b/invenio/modules/pages/admin.py
@@ -20,7 +20,11 @@
 
 from __future__ import unicode_literals
 
+import os
+
 from flask import current_app
+
+from invenio.base.globals import cfg
 
 from invenio.ext.admin.views import ModelView
 from invenio.ext.sqlalchemy import db
@@ -28,12 +32,17 @@ from invenio.modules.pages.models import Page
 
 from jinja2 import TemplateNotFound
 
-from wtforms.validators import ValidationError
+from werkzeug import secure_filename
+from werkzeug.local import LocalProxy
+
+from wtforms.validators import DataRequired, ValidationError
 
 
 def template_exists(form, field):
     """Form validation: check that selected template exists."""
-    template_name = "pages/" + field.data
+    if not field.data:
+        return
+    template_name = "pages/" + secure_filename(field.data)
     try:
         current_app.jinja_env.get_template(template_name)
     except TemplateNotFound:
@@ -60,7 +69,8 @@ class PagesAdmin(ModelView):
 
     form_args = dict(
         template_name=dict(
-            validators=[template_exists]
+            default=LocalProxy(lambda: os.path.basename(cfg["PAGES_DEFAULT_TEMPLATE"])),
+            validators=[DataRequired(), template_exists]
         ))
 
     # FIXME if we want to prevent users from modifying the dates


invenio/modules/pages/admin.py
```diff
@@ -60,6 +62,7 @@ class PagesAdmin(ModelView):
 
     form_args = dict(
         template_name=dict(
+            default="default.html",
```
bouzlibop: Shouldn't we go here with sth like `PAGES_DEFAULT_TEMPLATE` from `pages/config.py`

invenio/modules/pages/admin.py
```diff
@@ -33,6 +33,8 @@
 
 def template_exists(form, field):
     """Form validation: check that selected template exists."""
+    if not field.data:
```
jirikuncar: what about adding 'required' validator to https://github.com/inveniosoftware/invenio/pull/3337/files#diff-b83bf3bae0fb93452518c475004ffe97R66 instead?

invenio/modules/pages/admin.py
jmartinm: I was looking at adding:

```python
default=os.path.basename(cfg["PAGES_DEFAULT_TEMPLATE"])
```

here but then I get:

```
RuntimeError: working outside of application context
```

How can the app config be accessed from the class definition?

invenio/modules/pages/admin.py
jmartinm: Makes sense. Will add that.

invenio/modules/pages/admin.py
```diff
@@ -33,6 +33,8 @@
...
+        return
     template_name = "pages/" + field.data
```
jirikuncar: consider using `werkzeug. secure_filename`

invenio/modules/pages/admin.py
bouzlibop: Hmm, I'm not sure how to do it. @jirikuncar, do you know how could this be solved?

jirikuncar: It used to work with lambda function or alternatively you can wrap it to a LocalProxy.

jmartinm:
I will create a PR with these changes for `Invenio-pages`

invenio/modules/pages/admin.py
```diff
@@ -60,7 +69,8 @@ class PagesAdmin(ModelView):
 
     form_args = dict(
         template_name=dict(
-            validators=[template_exists]
+            default=LocalProxy(lambda: os.path.basename(cfg["PAGES_DEFAULT_TEMPLATE"])),
```
jirikuncar: btw have you checked if it works only with lambda function?

jmartinm:
superseded by https://github.com/inveniosoftware/invenio-pages/pull/3

invenio/modules/pages/admin.py
jmartinm: Yes, it seems to work with:

```python
default=lambda: os.path.basename(cfg["PAGES_DEFAULT_TEMPLATE"]),
```

jmartinm: I updated inveniosoftware/invenio-pages#3

PR #3340, opened 2015-07-07
search: no dynamic numeric fields in elasticsearch

### LINKED ISSUE #3339: search: elasticsearch schema fields wrongly mapped to numeric
The elasticsearch mapping for Record contains `numeric_detection": True`. This  enables the numeric fields autodetection.
With this option enabled, if a field which is not explicitly defined in the schema is present in a document and if it looks like a number, then it will mapped as a number. If next documents don't contain an integer for this field, the error `MapperParsingException` will be returned.

*Scenario:*
```
# create demo records
$ inveniomanage records create -t marcxml < invenio/testsuite/data/demo_record_marc_data.xml
# index the data
$ inveniomanage shell
  $ from invenio_records.models import Record
  $ from invenio.ext.es import index_record
  $ [index_record(r.id) for r in Record.query.all()]
...
RequestError: TransportError(400, u'MapperParsingException[failed to parse [universal_decimal_classification_number.universal_decimal_classification_number]]; nested: NumberFormatException[multiple points]; ')
```

nharraud:
* FIX Fixes elasticsearch indexing by disabling automatic detection of
  numeric fields. (closes #3339)

Signed-off-by: Nicolas Harraudeau <nicolas.harraudeau@cern.ch>

### DIFF
--- a/invenio/ext/es.py
+++ b/invenio/ext/es.py
@@ -70,7 +70,7 @@ SEARCH_RECORD_MAPPING = {
         "record": {
             "_all": {"enabled": False},
             "date_detection": False,
-            "numeric_detection": True,
+            "numeric_detection": False,
             "dynamic_templates": [
                 {"default": {
                     "match_mapping_type": "string",
--- a/invenio/ext/es.py
+++ b/invenio/ext/es.py
@@ -189,6 +189,7 @@ SEARCH_RECORD_MAPPING = {
 
 @celery.task
 def index_record(recid):
+    """Index a record in elasticsearch."""
     from invenio_records.models import RecordMetadata
     record = RecordMetadata.query.get(recid)
     es.index(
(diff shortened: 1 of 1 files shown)


egabancho:
Tested-by: me :shipit:

jirikuncar:
@nharraud we don't need to use `FIX` label for bugs that were not in any released version.

tiborsimko:
> @nharraud we don't need to use FIX label for bugs that were not in any released version.

Indeed.  The developers using master branch are supposed to read all commit logs anyway, and users upgrading from the previous stable release would be wondering if this item went into the release notes, because there was no Elasticsearch then.  

We could amend the release notes afterwards, but it's good to get it right from the get go.  I hope to contribute the release notes gathering script to the kwalitee package [soon](https://github.com/inveniosoftware/kwalitee/issues/71) so that release notes previewing could be an integral part of the regular feature branch kwalitee checking process...

PR #3342, opened 2015-07-08
legacy: mysql specific code removal

hachreak:
* Removes import from Mysqldb. (addresses #3214)


hachreak:
Files already removed from invenio.

PR #3346, opened 2015-07-10
search: response recids fix

### LINKED ISSUE #3335: search: Results class attribute not found
In https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/search/api.py#L98

```python
self.query
```

gives an error because the `Results` class does not have such attribute.

I guess the correct code might be:

```python
self.body["query"]
```

jalavik:
* Fixes an exception when calling `Results.recids`. (closes #3335)


jalavik:
@jirikuncar @jmartinm

PR #3352, opened 2015-07-13
global: fix fixtures loading on PostgreSQL

### LINKED ISSUE #3320: accounts: user creation with id = 1 problem on PostgreSQL
PostgreSQL doesn't update the sequence after insertion of [users fixtures](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/accounts/fixtures.py#L32) (user admin).

http://stackoverflow.com/questions/9108833/postgres-autoincrement-not-updated-on-explicit-id-inserts

Here you can read that:

> That's how it's supposed to work - next_val('test_id_seq') is only called when the system needs a value for this column and you have not provided one. If you provide value no such call is performed and consequently the sequence is not "updated".

So, after inserting the user admin in the table, the sequence associated with the id is not updated.
It means that next user that will be inserted will generate the error:

```python
IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "user_pkey"
DETAIL:  Key (id)=(1) already exists.
```

A possible solution is to not specify the id in `fixtures.py`.
Any ideas? :)

hachreak:
* Removes the id on fixtures to generate automatically. (closes #3320)

### DIFF
--- a/invenio/modules/accounts/fixtures.py
+++ b/invenio/modules/accounts/fixtures.py
@@ -29,7 +29,6 @@ class UserData(DataSet):
     """User data."""
 
     class admin:
-        id = 1
         email = current_app.config.get('CFG_SITE_ADMIN_EMAIL')
         password = ''
         note = '1'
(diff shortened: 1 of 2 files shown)


tiborsimko:
Superseded by Invenio 3.

PR #3356, opened 2015-07-14
search: reseting jrec on facet click

### LINKED ISSUE #3353: search: reset jrec on facet click
Navigating in the second page of the search results has a problem with facets.

if the number of results after faceting is less than `jrec` no results are shown:

See https://www.dropbox.com/s/cqiyoktfr0sawlb/Screenshot%202015-07-13%2018.33.19.png?dl=0

The value of `jrec` should be reseted.

jirikuncar:
* Resets pagination on facet click.  (closes #3353)

### DIFF
--- a/invenio/modules/search/templates/search/search_base.html
+++ b/invenio/modules/search/templates/search/search_base.html
@@ -160,18 +160,27 @@ require(['jquery', 'js/search/search'], function($) {
     }).join(' AND ')
 
     data['p'] += ' ' + limit;
-
     $('form[name="search"] input[name=p]').val(data['p'])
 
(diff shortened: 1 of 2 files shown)


jmartinm:
It works for me although the URL still shows e.g. `jrec=11` even though after clicking on the facets there is only one page of results (with 2 hits for example)

A refresh of the page also makes the facets go away.

We could use as inspiration http://doc.rero.ch/ where the `jrec` is correctly set to `1` and a refresh of the page still keeps the facets.

jirikuncar:
> A refresh of the page also makes the facets go away.

I'm not going to re-implement javascript facets here. It was just proof of concept and some serious try using FlightJS should be made.

PR #3360, opened 2015-07-14
encoder: correction of function call

### LINKED ISSUE #3354: encoder: process_batch_job gives wrong arguments to compose_file
In `process_batch_job`, `compose file` is called with directory, docname and extension as unnamed arguments [1], while the definition [2] takes only `dirname` and `extension` as unnamed arguments. This leads to `compose_file` returning `<directory>/<docname>.<docname>;<extension>` as path, instead of `<directory>/<docname>.<extension>`

[1] https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/encoder/batch_engine.py#L542
[2] https://github.com/inveniosoftware/invenio/blob/master/invenio/legacy/bibdocfile/api.py#L520

blixhavn:
* FIX Corrects the `compose_file` function call in `process_batch_job`
  to produce `<directory>/content.<extension>` instead of
  `<directory>/content.content;<extension>` (closes #3354).

### DIFF
--- a/invenio/modules/encoder/batch_engine.py
+++ b/invenio/modules/encoder/batch_engine.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Invenio.
-# Copyright (C) 2011 CERN.
+# Copyright (C) 2011, 2015 CERN.
 #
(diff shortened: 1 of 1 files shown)


ludmilamarian:
:shipit:

PR #3361, opened 2015-07-14
installation: usage of sanitised database name

### LINKED ISSUE #3303: docs: db setup breaks for non-master

jirikuncar:
* BETTER Uses sanitised variable `$BRANCH_ALNUM` (from `$BRACH` name)
  in installation guide for the database name and user. (closes #3303)
  (closes #3304)

Signed-off-by: Alexander Wagner <alexander.wagner@desy.de>

--
WDYT @aw-bib ?

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -351,7 +351,9 @@ locally and it will make your life easier. ``(invenio)$`` tells your that the
 
 .. code-block:: console
 
-    $ mkvirtualenv invenio
+    $ # choose an unique name for your virtual environment
+    $ export VENAME=invenio
+    $ mkvirtualenv $VENAME
     (invenio)$ # we are in the invenio environment now and
     (invenio)$ # can leave it using the deactivate command.
     (invenio)$ deactivate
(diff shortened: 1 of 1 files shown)


dset0x:
There's no error messages in the issues that this PR closes. What's the actual problem?

```
mysql> create database `Βάση!@#$%^&*()[];'\,./{}:"|<>?-=_+`;
Query OK, 1 row affected (0.00 sec)


```

jirikuncar:
@dset0x please see the description in #3303 and follow the discussion there. Thank you

tiborsimko:
Option 1: I guess we could escape DB name everywhere, then there should be no problem with any character, as @dset0x says.

Option 2: This PR may be sufficient and quicker though,  e.g. in `invenio-devscripts`'s [invenio-initialise-virtualenv](https://github.com/tiborsimko/invenio-devscripts/blob/master/invenio-initialise-virtualenv#L58) I'm using this sanitisation:

```
DBNAME=$(echo $VENAME | sed "s/\-//g" | sed "s/\.//g")
```

ending up with database names like `inveniomaint11`, `inveniomaint12`, `inveniolegacy`, `inveniomaint20`, `inveniomaster`...

aw-bib:
It sounds logical to go for a common sanitisation. Looks like @tiborsimko approach would be the way to go as it's already in the `invenio-devscripts`.

dset0x:
Even though Option 2 is quicker to implement:
* `maint-2.1` and `maint--2,1` are now exactly the same
* we are avoiding the actual problem, which is lack of correct quoting

INSTALL.rst
```diff
@@ -458,10 +460,12 @@ the following commands.
 
 .. code-block:: console
 
+    (invenio)$ # sanitaze for usage as database name and user
+    (invenio)$ export SAFE_NAME=`echo $VENAME$BRANCH | sed -e 's/[^A-Za-z0-9]//g'`
```
jirikuncar: WDYT @tiborsimko @aw-bib ?

aw-bib: sounds good.

Issue #3283, opened 2015-06-17
roadmap: simplified records module


jirikuncar:
- [x] **storage** 
  - [x] removal of `invenio.modules.records`
  - SQLAlchemy model - PostgreSQL JSONB field if available (see "[invenio-records](https://github.com/jirikuncar/invenio-records)")
  - [x] release *Invenio-Records*
  - [x] ~~create an upgrade recipe to migrate to new *Invenio-Records* storage (added column with version, re-submit data)~~ recommended way is to re-upload data to new instance
- [x] **indexing**
  - [x] removal of BibIndex, BibRank, BibSort
  - [x] write signal handlers to launch ~~bibindex~~ Elastic search tasks
  - [x] create configurable `es` instance ~~(`invenio.ext.es`)~~ in *Invenio-Search*.
  - [x] prepare mapping configuration for Elasticsearch (https://github.com/inveniosoftware/es-jsonschema/pull/1 by @nharraud)
- [x] **quering** 
  - [x] native IR (later Elasticsearch) for indexed records 
  - [x] SQLAlchemy + JSONB for exact database searches if needed (#2379)
  - [x] create new search API: `invenio.modules.search.api:Query('title:higgs')[10:20]` based on https://github.com/inveniosoftware/invenio/issues/2478#issuecomment-71647267 from January 27th
- [x] **cli** 
  - [x] `inveniomanage records create` Flask-Script/Click
  - [x] `inveniomanage records patch/update`
  - [x] removal of *uploader* (#3365)
  - [x] removal of *BibUpload*
  - [x] create registry of format convertors (https://github.com/inveniosoftware/invenio-records/pull/3)
- [x] **data conversion** 
  - [x] "[DoJSON](https://github.com/jirikuncar/dojson)", XSLT, JSONPatch depending on data format
  - [x] release *DoJSON* (inveniosoftware/dojson#18 @tiborsimko)
  - [x] proof read `dojson.contrib.marc21` (@egabancho)
  - [x] discuss possibilities for extending/editing DoJSON rules
- [x] **formatting**
  - [x] ~~port `bfe_*` formatting elements to Jinja (#3244)~~ **invenio-records-ui** by @lnielsen 
  - [x] ~~collection count for restricted collections?~~ (updated: new UI without counts)
  - [x] ~~update formatter templates (#2662)~~ new **invenio-search-js** and **invenio-search-ui** by @drjova 

--
If you would like to test current implementation check #3307.

```
[...]

PR #3365, opened 2015-07-15
uploader: module removal

jirikuncar:
* INCOMPATIBLE Removes uploader module that it is not needed since
  ACID database backend and new Invenio-Records module is in use.
  (addresses #3283)


tiborsimko:
> that it is not needed 

INCOMPATIBLE Removes uploader module that is not needed anymore as per the new ACID database backend and the new Invenio-Records module.

PR #3367, opened 2015-07-16
workflow: protection against DB session problems caused by the payload

crepererum:
I know that this should not be required, but right now it is because (for some bad reason) some parts of our code are messing up the db session using a rollback. Without this, you do not get any usable debug information but instead a `InvalidRequestError` from the workflows code.

In general we should think about the following 2 things:
 1. Not using a global db session object (i.e. kill `invenio.ext.sqlalchemy.db`).
 2. Better isolation and decoupling of workflow payloads. When doing this we might also move the execution to celery instead of doing it in the request thread.

### DIFF
--- a/invenio/modules/workflows/client.py
+++ b/invenio/modules/workflows/client.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # This file is part of Invenio.
-# Copyright (C) 2012, 2013, 2014 CERN.
+# Copyright (C) 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
@@ -19,9 +19,14 @@
 """Low-level functions to run workflows."""
 
 import traceback
-from .errors import WorkflowHalt, WorkflowError
-from .models import ObjectVersion
+
+from invenio.ext.sqlalchemy import db
+
+from sqlalchemy.exc import InvalidRequestError
+
 from .engine import WorkflowStatus
+from .errors import WorkflowError, WorkflowHalt
+from .models import ObjectVersion
 
 
 def run_workflow(wfe, data, stop_on_halt=False,
@@ -39,15 +44,25 @@ def run_workflow(wfe, data, stop_on_halt=False,
     :param wfe: workflow engine in charge of workflow execution
     """
     while True:
+        # Make a backup of the db session and replace the global one with a
+        # new clean version. This is required because the workflow payload
+        # might do horrible things with it and we need to for querying+storing
+        # the workflow state/objects.
+        session_backup = db.session
+        db.session = db.create_scoped_session()
+
         try:
             if initial_run:
                 initial_run = False
-                wfe.process(data)
+                wfe.process(data, session_backup)
+                db.session = session_backup
                 break
             else:
-                wfe.restart('next', 'first')
+                wfe.restart('next', 'first', session_backup)
+                db.session = session_backup
                 break
         except WorkflowHalt as workflowhalt_triggered:
+            db.session = session_backup
             current_obj = wfe.get_current_object()
             if current_obj:
                 if workflowhalt_triggered.action:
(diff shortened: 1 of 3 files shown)


invenio/modules/workflows/client.py
```diff
@@ -39,15 +44,25 @@ def run_workflow(wfe, data, stop_on_halt=False,
     :param wfe: workflow engine in charge of workflow execution
     """
     while True:
+        # Make a backup of the db session and replace the global one with a
+        # new clean version. This is required because the workflow payload
+        # might do horrible things with it and we need to for querying+storing
+        # the workflow state/objects.
+        session_backup = db.session
+        db.session = db.create_scoped_session()
```
kaplun: Does this physically creates a new connection to the DB? Or are connections taken from a fixed-size pool? Otherwise we risk to exhaust the available DB connections (e.g. in case of many parallel workflows).

invenio/modules/workflows/client.py
```diff
@@ -39,15 +44,25 @@ def run_workflow(wfe, data, stop_on_halt=False,
...
+
         try:
             if initial_run:
                 initial_run = False
-                wfe.process(data)
+                wfe.process(data, session_backup)
```
kaplun: Why are you working on the `session_backup` rather than the new clean session?

invenio/modules/workflows/client.py
```diff
@@ -39,15 +44,25 @@ def run_workflow(wfe, data, stop_on_halt=False,
...
+                db.session = session_backup
                 break
             else:
-                wfe.restart('next', 'first')
+                wfe.restart('next', 'first', session_backup)
+                db.session = session_backup
                 break
         except WorkflowHalt as workflowhalt_triggered:
+            db.session = session_backup
```
kaplun: Shouldn't this be rather part of a `try..finally`?

invenio/modules/workflows/client.py
crepererum: because `wfe` is bound to the old session

invenio/modules/workflows/client.py
crepererum: You need the session back in the `except` branch, not afterwards.

jalavik:
Do you have the steps on how to reproduce this problem? I have not seen it while running workflows and depositions on INSPIRE labs, but perhaps the payload's have not been naughty enough so far :-)

invenio/modules/workflows/client.py
crepererum: You need another connection from the pool for the payload, during the payload execution. See comment on the celery solution for a cleaner future.

crepererum:
Use the demosite and make a deposit WITH a file (whatever you want).

jirikuncar:
This is problem has become more visible since we switched to ACID database backend. Hence we need to be more careful about DB session usage.

--
cc @lnielsen

jalavik:
As you point out the problem is that you are unable to find (and debug) the original exceptions that are thrown inside a workflow. @dset0x and I had a look at testing this against the upcoming new workflows code (WIP in #2628) which does a better job of spitting out this kind of debug information than current master. It allowed us to get one step closer to the real issue but we were stopped by the `import_string` "wall" of current werkzeug and flask-registry. By applying @dset0x PRs to these packages (https://github.com/inveniosoftware/flask-registry/pull/34) we were better able to pinpoint the problems. Unfortunately, a long way to go just for a stacktrace..

To summarize, while this patch provides the original stacktrace representation I hope in the future we will find a better way of providing this information. Perhaps as a combination of better session handling and the refactored code from @dset0x .

crepererum:
Can we merge it to finally fix that problem in `master`? I'm asking because master is broken right now and it is nearly impossible to debug it without these changes.

dset0x:
@crepererum It's a go from our side since it's causing debugging issues.

jirikuncar:
Workflow engine has been notably changed in Invenio-Workflows and [Workflow 2.0](https://github.com/inveniosoftware/workflow/releases/tag/v2.0.0).

PR #3371, opened 2015-07-17
travis: speed up

crepererum:
* Redesigns .travis.yml to use the new container-based infrastructure.
  This lowers the waiting time for the queue, gives us the ability to
  use caches for apt and pip and improves testing time down to ~15min.
  We are not allowed too use `sudo` any more but we should need it,
  because Apache is not tested anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ services:
   - mongodb
   - elasticsearch
 
+sudo: false
+
 language: python
 
 cache:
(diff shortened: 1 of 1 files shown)


jirikuncar:
@crepererum please prepare separate PR for maint-2.0. Thanks

lnielsen:
Nice!

tiborsimko:
:+1: 

Invenio being more and more of a Pythonic framework, testing for real Apache is not necessary anymore (perhaps just checking generated vhost configuration) here.  For deployment-like testing scenarios, we can rely on stronger regression/integration test suite living in the demo site overlays (that is somewhat thin in v2.x still, while it was much richer in v1.x).

PR #3377, opened 2015-07-17
global: installation slimdown

crepererum:
Slimdown installation. See individual commits for explanation.

Rebased against: #3371


setup.py
```diff
@@ -160,8 +159,6 @@ def run(self):
     "dropbox": [
         "dropbox>=2.1.0"
     ],
-    "elasticsearch": [
```
jirikuncar: you should keep it for empty for few versions not to break overlays depending on Invenio.

setup.py
```diff
@@ -175,9 +172,6 @@ def run(self):
         "qrcode>=5.1",
         "Pillow>=2.7.0"
     ],
-    "mongo": [
-        "pymongo>=3.0"
```
jirikuncar: just make it empty

crepererum:
done

jirikuncar:
`s/remove/removal/`

crepererum:
done

jirikuncar:
:shipit:

PR #3383, opened 2015-07-21
comments: module separation

jirikuncar:
* INCOMPATIBLE Separates comments module to independent
  Invenio-Comments package.


egabancho:
:shipit:

invenio/base/config.py
```diff
@@ -70,6 +71,7 @@
 PACKAGES_EXCLUDE = [
     'invenio.modules.annotations',
     'invenio.modules.archiver',
+    'invenio.modules.comments',
```
kaplun: Hi @jirikuncar , just for me to understand: you are adding here the legacy module in the `PACKAGE_EXCLUDE`, to be sure it's not referenced by mistake in overlays? In that case: aren't overlays overriding localy this variable anyway?

PR #2841, opened 2015-03-04
BibCheck: improvements

kaplun:

PR #2824, opened 2015-02-26
BibFormat: outdated_caches() fix

kaplun:
- Fixes outdated_caches() query that, by filtering out record not
  modified since the last bibreformat run, was implicitly skipping
  those records that were, by whatever reason, not reformatted in
  the past.

Signed-off-by: Samuele Kaplun samuele.kaplun@cern.ch

PR #2758, opened 2015-02-10
BibIndex: infinite loop in beautify_range_list

kaplun:
- Fixes infinite loop in beautify_range_list() function that
  was triggered in particular conditions.

Signed-off-by: Samuele Kaplun samuele.kaplun@cern.ch

PR #3384, opened 2015-07-21
WIP Prod for legacy

kaplun:
This is just a mega branch with every patch done on INSPIRE that applied without conflict on top of Invenio legacy branch. Maybe that it could be used as reference for pushing upstream some quick fixes.
(cc @egabancho)


kaplun:
@egabancho to fix conflicts is quite a painful process. What do you suggest is the best way to proceed? Maybe we should one day simply sit together sifting all the commits that we have never integrated upstream and see which one you are interested into?

egabancho:
I would say, at least, make a PR module by module.
Ideally I would go commit by commit, if they are not closely related i.e one can't live without the other,  and make PRs for them, like you did already for https://github.com/kaplun/invenio/commit/64333fe3ae88de3faeb22b419d02d034c4ef2882 (I can imagine it is doable for most of them)

Thinking with the CDS hat the smaller the granularity of the PR the easier to deploy and test. 
I know you can't change the past, but at least for the future I suggest you to make a PR to your repo and to inveniosoftware with the same branch, it just takes a few seconds and it will save you a few headaches. 

> which one you are interested into?

[Un]fortunately it's not only us who might be interested in Inpire contributions but anyone using Invenio 1.x ... potentially all the enhancements/fixes should be welcome, but of course I can give you our opinion about which ones are particularly interesting for CDS

kaplun:
I did several PRs in the past, but since they were not integrated, they became stale and eventually incompatible. Check e.g. #2365, #2824, #2787, #2254, #2021, #2758, #2841, #2361, (and almost #2526). 
Since at INSPIRE we are basically no longer actively working on legacy (and we were based on maint-1.1), it is quite painful for me to go back and resurrect all patches and amend them against latest legacy, and finding out no-one needs them or that they are simply not integrated... Additionally, unfortunately, many patches are not nicely self-contained, and are cross-modules.

> I know you can't change the past, but at least for the future I suggest you to make a PR to your repo and to inveniosoftware with the same branch, it just takes a few seconds and it will save you a few headaches.

Ah that is already the case, and is made easy by the new course of split-up modules.

PR #3385, opened 2015-07-21
pages: module separation

jmartinm:
* INCOMPATIBLE Separates pages module to independent
  Invenio-Pages package.


jmartinm:
Together with https://github.com/jmartinm/invenio-pages/tree/pages

Separates pages module into an independent package.

jirikuncar:
@jmartinm I have pushed your branch jmartin/invenio-pages@pages as inveniosoftware/invenio-pages@master.

invenio/base/config.py
```diff
@@ -74,7 +75,6 @@
     'invenio.modules.comments',
     'invenio.modules.communities',  # remove with invenio/modules/communities
     'invenio.modules.multimedia',
-    'invenio.modules.pages',
```
jirikuncar: IMHO we can still keep it here.

invenio/base/config.py
```diff
@@ -63,6 +63,7 @@
 PACKAGES = [
     'invenio_records',
     'invenio_comments',
+    'invenio_pages',
```
jirikuncar: We should not force people to use this package as it was disabled by default. We can maybe add it to `extras_require = {..., 'pages': ['invenio-pages>=0.1.0'], ...}`.

invenio/base/config.py
jmartinm: I can keep it, what is the use of it when the module is completely removed from Invenio?

PR #3387, opened 2015-07-22
global: maximum oauthlib requirement

hachreak:
* Fixes a maximum version requirement for oauthlib. (closes #3379)

### DIFF
--- a/invenio/modules/knowledge/testsuite/test_knowledge_restful.py
+++ b/invenio/modules/knowledge/testsuite/test_knowledge_restful.py
@@ -172,13 +172,7 @@ class TestKnowledgeRestfulAPI(APITestCase):
             user_id=1,
         )
 
-        answer = get_answer.json
-
-        expected_result = dict(
-            status=404,
-        )
-
-        assert answer['status'] == expected_result['status']
+        assert get_answer.status_code == 404
 
     def test_get_knwkb_mappings(self):
         """Test the return of list of mappings."""
--- a/setup.py
+++ b/setup.py
@@ -116,6 +116,9 @@ install_requires = [
     "MySQL-python>=1.2.5",
     "numpy>=1.7",
     "nydus>=0.10.8",
+    # FIXME new oauthlib release after 0.7.2 has some compatible problems with
+    # the used Flask-Oauthlib version.
+    "oauthlib==0.7.2",
     "passlib>=1.6.2",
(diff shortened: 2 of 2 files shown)


hachreak:
ping @jirikuncar

setup.py
```diff
@@ -116,6 +116,7 @@ def run(self):
     "MySQL-python>=1.2.5",
     "numpy>=1.7",
     "nydus>=0.10.8",
+    "oauthlib==0.7.2",
```
lnielsen: ``"oauthlib<=0.7.2"``?

invenio/modules/knowledge/testsuite/test_knowledge_restful.py
```diff
@@ -172,13 +172,7 @@ def test_get_knwkb_not_exist(self):
...
         )
 
-        answer = get_answer.json
-
-        expected_result = dict(
-            status=404,
-        )
-
-        assert answer['status'] == expected_result['status']
+        assert get_answer.status_code == 404
```
lnielsen: Is there a reason why this is change is in the same commit?

hachreak: because, for some reasons, devel and release requirements suffers also about of another problem: https://s3.amazonaws.com/archive.travis-ci.org/jobs/71936871/log.txt
```python
ERROR: test_get_knwkb_not_exist (invenio.modules.knowledge.testsuite.test_knowledge_restful.TestKnowledgeRestfulAPI)
  File "/home/travis/build/inveniosoftware/invenio/invenio/modules/knowledge/testsuite/test_knowledge_restful.py", line 181, in test_get_knwkb_not_exist
    assert answer['status'] == expected_result['status']
KeyError: 'status'
```

setup.py
hachreak: it's not permitted by `requirements.py`.
Maybe if you set major and minor version like `oauthlib>=0.7.2<1.0`. 
But, what's the difference? This is an "hack" to require a specific (old) version and in future we need to fix dependencies anyway.

jirikuncar: Can you add `FIXME` or more verbose commit message with a reason why we pin the package version and what should be done to unpin it again. Thanks

hachreak:
ping @jirikuncar :shipit:

PR #3389, opened 2015-07-22
global: on pip installed via brew no sudo required

errikos:
* Removes `sudo` from `pip` command because brew directories belong to the user, so no `sudo` is required to execute any command that writes to them (such as `pip` for example).

* Removes the extra space from the `-- universal` argument in line 172.

Also please consider removing the `--framework` argument in line 172, as it seems that is no longer included in the python formula options (so it's just ignored by brew).

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -169,9 +169,9 @@ Now, it is time to start installing the prerequisites.
 
 .. code-block:: console
 
-    $ brew install python --framework -- universal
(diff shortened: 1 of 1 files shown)


jirikuncar:
Maybe we should backport it for `maint-2.0`. @errikos can you please keep break the lines in commit message before 72 characters? (tip: you can install `kwalitee` tool - `pip install --pre kwalitee`, `kwalitee check message`).

PR #3391, opened 2015-07-22
oaiharvester: module separation

jalavik:
* INCOMPATIBLE Separates oaiharvester module to independent
  Invenio-OAIharvester package.


jirikuncar:
@jalavik `s/Invenio-OAIharvester/Invenio-OAIHarvester/`?

jalavik:
@jirikuncar Updated. I left the upgrade scripts for now - tests were failing otherwise. The upgrade scripts will be part of the external module and can be removed when we release the Invenio-OAIHarvester and include it upon Invenio installation.

PS: At some point we need to do something with upgrade dependencies across modules. E.g. `invenio_2015_07_14_innodb` depends on `oaiharvester_2015_07_14_innodb`, but this won't work if we start to package Invenio with a selective approach.

jirikuncar:
:shipit:

Issue #1769, opened 2014-06-08
RFC: JavaScript Code Style


invenio-developers:
<a href="https://github.com/jirikuncar"><img src="https://avatars.githubusercontent.com/u/189798?" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jirikuncar](https://github.com/jirikuncar)**
_Thursday May 22, 2014 at 11:54 GMT_
_Originally opened as https://github.com/jirikuncar/invenio/issues/262_

---

| Revision | Date | Author |
| --- | --- | --- |
| Initial | 2014-05-22 | @jirikuncar |
# Idea

It'd be nice to use the standard way for JavaScript.
# Proposals
- @lnielsen-cern: See previous discussion as well http://invenio-software.org/wiki/Talk/JSFrameworks
- @giokokos https://github.com/airbnb/javascript
### Testing
- @greut The first point to enforce regarding JS is **no inline JavaScript, no `onclick`, no `href=javascript:`**. The second would be **tests**, [QUnit](http://qunitjs.com/) is super simple and nice. 
  Extra pain points: [jslint](http://jslint.com/), [jsfmt](http://rdio.github.io/jsfmt/)
### Related
- Require.js #1794, #1830, #1833

jacquerie:
It appears that this RFC didn't reach consensus, so no decision was made. Let me try to rekindle the discussion with some organization and some proposals.

The discussion started from code style, then touched code quality, the use of frameworks, testing, and documentation. I think these are very different problems, which require different discussions.

1. JavaScript code style: IMHO, not as important as the others, but still nice to have. Let me mention [JSCS](https://github.com/jscs-dev/node-jscs), a configurable tool to enforce code style conventions. Once we decide on a style guide (or perhaps we can lean on the defaults of some project we like), it should make the job much easier.
2. JavaScript code quality: I think there are two parts to this. First we have the style of JavaScript files: a tool that was already mentioned is [JSHint](https://github.com/jshint/jshint), which will require a discussion on the settings. Then there's the problem of the obtrusiveness of the JavaScript in A LOT of the templates. Judging from the output of `ag onclick` this will require A LOT of work.
3. Use of JavaScript frameworks: consensus in #2110 is that this shouldn't be pursued.
4. Testing JavaScript libraries: RFC #2221 is still open, but, from what I've seen, Jasmine is the de-facto standard.
5. Documenting JavaScript: RFC #2845 is still open, and this is where it should be discussed.

[...]

PR #3400, opened 2015-07-28
search: UI improvements by using FlightJS components

nharraud:


invenio/modules/search/templates/search/form/index_base.html
```diff
@@ -33,6 +33,22 @@
     </div>
   </div>
 </div>
+<div id="search-query-filters"
```
jirikuncar: please add a block `search_query_filters` around this div

invenio/modules/search/walkers/facets.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @visitor(NotOp)
+    def visit(self, node, op):
+        return self._invert_facets(op)
+
+    @visitor(KeywordOp)
+    def visit(self, node, left, right):
+        return {
+            node.left.value: {
+                'inc': Set([node.right.value]),
```
jirikuncar: can you use standard python set - `set` or `{}`?

jmartinm:
Hi @nharraud do you mind adding a screenshot of how the facets look in your demo site after these changes?

I was trying to install it but the demo site is currently quite broken so could not load the demo records.

Thanks!

invenio/modules/search/static/js/search/facets_menu.js
```diff
@@ -0,0 +1,95 @@
...
+        value: data.el.value,
+        state: data.el.checked ? 'include' : 'default',
+      });
+    }
+
+    /**
+     * Initialize the component
+     */
+    this.init = function() {
+      var cpn = this;
```
jalavik: Just a minor thing. In recent JS files added to Invenio the pattern for this has usually been `var that = this` (ala Crockford/jQuery). I do not mind either way, but it would be good if this is standard across all new JS we add. See also discussion in #1769

nharraud: Good point.

jmartinm:
@nharraud I gave it a try on the demo site and I can see one bug:

* Selecting 2 options from the same facet box and then click edit on the `Query filter` adds to the search box an extra 'AND' in front.

Regarding how the facets work, IMO it would be nice to have a user interface similar to http://doc.rero.ch/search?c=NAVDOCTYPE.ARTICLE&ln=en where every time you click on one facet option the facets get recomputed and the count close to each option reflects the number of results you will get when you click on the facet. What do you think?

What are the plans of development? Are you going to keep working on facets after this PR?

invenio/modules/search/static/js/search/search.js
```diff
@@ -18,293 +18,240 @@
...
-                try {
-                    filter = $.parseJSON(hash_filter)
-                } catch (exc) {
-                    console.exception(exc)
-                }
+define([
+        'jquery',
+        'vendors/URIjs/src/URI',
+        'flight/lib/component',
+        './facets_filter',
```
jmartinm: maybe better to include abs path like https://github.com/inveniosoftware/invenio/pull/3400/files#diff-adcf435606f6def2c24216a197668bf3R25 ?

jirikuncar:
> Regarding how the facets work, IMO it would be nice to have a user interface similar to http://doc.rero.ch/search?c=NAVDOCTYPE.ARTICLE&ln=en where every time you click on one facet option the facets get recomputed and the count close to each option reflects the number of results you will get when you click on the facet. What do you think?

In the future it will be configurable if you want facet filter to be applied as "filter" (with facet recomputing) or "post-filter" (without facet recomputing). We had this discussion before and the first solution doesn't play nicely with "OR" or "AND NOT" selections.

jmartinm:
I see also that the `post_filter=` parameter gets saved in between searches so when I search for something completely different in the search bar the same filters will get applied.

To me it would be more natural if doing a new search in the search bar would clear the filters and start from scratch.

invenio/modules/search/templates/search/form/index_base.html
jmartinm: +1

invenio/modules/search/static/js/search/search.js
nharraud: agreed

nharraud:
There is now an RFC issue for discussing the faceting design. Don't hesitate to comment on it if you have any other idea/need and/or if the current notes are unclear. #3415

invenio/modules/search/templates/search/form/index_base.html
```diff
@@ -33,6 +33,25 @@
     </div>
   </div>
 </div>
+
+{%- block search_query_filters -%}
+<div id="search-query-filters"
+     class="row {{ 'empty' if request.args.get('post_filter', '') == '' }}">
```
jirikuncar: `     class="row {{ 'empty' if request.args.get('post_filter') }}">`

```python
assert bool('') == False
assert bool(None) == False
```

invenio/modules/search/views/search.py
```diff
@@ -248,12 +248,21 @@ def search(collection, p, of, ot, so, sf, sp, rm, rg, jrec):
...
+    filtered_facets = ''
     from invenio.modules.search.walkers.elasticsearch import ElasticSearchDSL
     if 'post_filter' in request.values:
         post_filter = Query(request.values.get('post_filter')).query.accept(
             ElasticSearchDSL()
         )
         response.body['post_filter'] = post_filter
+        # extracting the facet filtering
+        from invenio.modules.search.walkers.facets import FacetsVisitor
+        filtered_facets = Query(request.values.get('post_filter')).query.accept(
```
jirikuncar: you can reuse parsed query from https://github.com/inveniosoftware/invenio/pull/3400/files#diff-66b51f83bbfcce07b1d663fe0cf19b92R254

nharraud:
This implementation still has one big limitation: the facets are not part of the Ajax responses. Thus we need to reload the whole page when the user query changes.

nharraud:
fixed one last bug which happened when history was browsed too rapidly for FlightJS components to load.

PR #3405, opened 2015-07-28
formatter: changes in id.tpl

Lilykos:
* Fixes the id.tpl file, since Jinja does not accept
  list comprehensions.

Signed-off-by: Ilias Koutsakis <ilias.koutsakis@cern.ch>

### DIFF
--- a/invenio/modules/formatter/templates/format/records/id.tpl
+++ b/invenio/modules/formatter/templates/format/records/id.tpl
@@ -17,4 +17,10 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 -#}
 {#- Prints list of records' IDs. -#}
-{{- [r['recid'] for r in records or []]|list|tojson -}}
+{% block list_records_ids %}
+    {% set record_list = [] %}
+    {% for r in records %}
+        {{ record_list.append(r['recid']) }}
+    {% endfor %}
+    {{ record_list|tojson }}
+{% endblock %}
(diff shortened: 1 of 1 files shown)


invenio/modules/formatter/templates/format/records/id.tpl
```diff
@@ -17,4 +17,10 @@
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 -#}
 {#- Prints list of records' IDs. -#}
-{{- [r['recid'] for r in records or []]|list|tojson -}}
```
jirikuncar: hups, good catch. we should change it however  to something like:
```jinja2
{{- records or [] | map(attribute="recid") | list | tojson -}}
```

jirikuncar:
```
formatter: fix list comprehension in id.tpl

...
```

jirikuncar:
@Lilykos thanks. Can you please rebase and re-push your branch?

invenio/modules/formatter/templates/format/records/id.tpl
```diff
@@ -17,4 +17,4 @@
...
+{{- records or [] | map(attribute="recid") | list | tojson -}}
```
jirikuncar: I've been testing and found an issue with `or`. The solution is using `default`:
```
{{- records | default([]) | map(attribute='recid') | list | tojson -}} 
```

jirikuncar: maybe the `default` is not needed after all

PR #3408, opened 2015-07-29
docker: fix macosx ports forwarding

jma:
Signed-off-by: Johnny Mariéthoz <Johnny.Mariethoz@rero.ch>

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -21,7 +21,7 @@ web:
   build: .
   command: inveniomanage runserver -h 0.0.0.0 -p 28080 -d -r
   ports:
-    - "127.0.0.1:28080:28080"
+    - "28080:28080"
   environment:
     - CACHE_REDIS_HOST=cache
     - CFG_DATABASE_HOST=db
@@ -55,11 +55,11 @@ worker:
 redis:
   image: redis
   ports:
-    - "127.0.0.1:26379:6379"
+    - "26379:6379"
 mysql:
   image: mysql
   ports:
-    - "127.0.0.1:23306:3306"
+    - "23306:3306"
   environment:
     - MYSQL_DATABASE=invenio
     - MYSQL_USER=invenio


jirikuncar:
@jma thank you. Can you please rebase on top of `maint-2.0` and remove `macosx` for commit message as it's more general problem if you are not running docker on localhost.

crepererum:
It is written down in the documentation how to forward ports on a boot2docker setup. See https://crepererum.github.io/invenio/developers/docker.html for an up-to-date version (the current read-the-docs version is outdated because of some issues).

We do NOT expose the dev server to the outer world in general (this is what binding to all interfaces = your PR is doing). The reason for this are simple: The dev server is not secure (e.g. no PW for admin, standard PWs for all other servers). If you now expose them in a public WiFi, your company network or any other shared network, you risk:
 1. your own security
 2. leakage/stealing of your maybe secretly developed new module/patch/overlay
 3. others security (because someone might hijack the container or use credentials that are stored there)
 4. ... (you can extend that list).

Therefor I reject this PR. Should you have any problems with the port forwarding described in the docs, feel free to ask for help.

jma:
Ok,
I understand.
If I well understood I have to use ssh tunnelling which can be slow. I cannot use a port forwarding at the NAT level, right?
Thanks.

crepererum:
@jma NAT level forwarding does NOT work for software bound to the loopback interface (= `lo` = `127.0.0.1` or `[::1]`). I doubt that you will recognize the slowness of the ssh port forwarding. For me more time is spend in the python code itself. Also you do not do any real time stuff (like X server forwarding or VNC), so the small latency should not be a problem.

In general: Should you find any better way on how to bind the boot2docker container ports to the loopback interface of the host system (and ONLYt that one), please let me know (or file a PR). I kinda wish that docker machine would be a little bit more helpful here, instead of relaying on the SSH workaround.

jma:
I have a suggestion: what about setting the HOST IP to 127.0.0.1 to limit the port forwarding to the host machine. In summary: apply the proposed PR run in the command line:
VBoxManage controlvm "boot2docker-vm" natpf1 "demosite_web,tcp,127.0.0.1,28080,,28080";
when the boot2manage is running (or modifyvm for stoped vm).

I tested this approach with two machines on the same wifi and the second machine cannot access the port 28080 (Connection refused).

The advantage of this approach is that the config should be done only once per boot2docker-vm.

VirtualBox port forwarding for boot2docker reference can found here:
https://github.com/boot2docker/boot2docker/blob/master/doc/WORKAROUNDS.md

crepererum:
No, because your PR still messes up security for non-boot2docker setups. Some people are using directly on their machines :wink:

jma:
Ok thus, I will keep this configuration for my machine. Thanks.

Issue #3406, opened 2015-07-29
global: inveniomanage database create fails


jmartinm:
How to reproduce
==============

With Invenio `master` and Invenio-demosite `master`

```bash
$ inveniomanage database init --user=root --yes-i-know
```

```bash
$ inveniomanage database create
```

```python
>>> There are 5 tables to be loaded.
Traceback (most recent call last):
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/bin/inveniomanage", line 9, in <module>
    load_entry_point('invenio==2.2.0.dev20150608', 'console_scripts', 'inveniomanage')()
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/src/invenio/invenio/base/manage.py", line 103, in main
    manager.run()
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/lib/python2.7/site-packages/flask_script/__init__.py", line 412, in run
    result = self.handle(sys.argv[0], sys.argv[1:])
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/lib/python2.7/site-packages/flask_script/__init__.py", line 383, in handle
    res = handle(*args, **config)
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/src/invenio/invenio/ext/script/__init__.py", line 179, in __call__
    post_command.send(sender, args=args, **kwargs)
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/lib/python2.7/site-packages/blinker/base.py", line 267, in send
    for receiver in self.receivers_for(sender)]
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/src/invenio/invenio/modules/upgrader/__init__.py", line 82, in populate_existing_upgrades
    map(iu.register_success, iu.get_upgrades())
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/src/invenio/invenio/modules/upgrader/engine.py", line 410, in get_upgrades
    self.order_upgrades(plugins, self.history))
  File "/Users/jmartinm/.virtualenvs/invenio-demosite/src/invenio/invenio/modules/upgrader/engine.py", line 497, in order_upgrades
    " upgrade %s" % (node_id, d))
RuntimeError: Upgrade invenio_2015_07_14_innodb depends on an unknown upgrade oaiharvester_2015_07_14_innodb
```

jalavik:
[...]

PR #3409, opened 2015-07-29
oaiharvester: fix missing __init__.py file

jma:
Signed-off-by: Johnny Mariéthoz <Johnny.Mariethoz@rero.ch>


jalavik:
Thanks. This is related to #3406 and to fully fix it I think you need to remove `invenio.modules.oaiharvester` from the `PACKAGES_EXCLUDE` as well (unless you override this on your overlay already). 

However, as mentioned in #3406 we were leaning towards fixing this with a release of the externalised OAI harvest module.

jma:
Thanks a lot.

Sorry for the troubles.

PR #3414, opened 2015-07-30
WebSession: unauthorized error message correction

### LINKED ISSUE #1414: Wrong error msg in "Submit" tab when NO submit role is connected to a user
Originally on 2013-03-19

When a logged-in user has NO sumbit permissions (to any doctype) and tries to click the submit button (on top), he gets: "Account 'xxx@yyy.zzz' is not yet activated. Try to login with another account"

Which is misleading... He should get an error that says "You are not authorized to perform this action."
Probably, an extra simple check should be incorporated in websubmit_webinterface for this case...

I've spotted this behavior a few months ago (and notified the list), but I think(?) the issue still stands even for the current version, so I thought it would be better to submit a ticket. Apologies if it was fixed in the meantime :)

FYI, the call is initiated in websubmit_webinterface.py

```
[...]
                if not at_least_one_submission_authorized and submission_exists:

                    if isGuestUser(uid):
                        return redirect_to_url(req, "%s/youraccount/login%s" % (
                            CFG_SITE_SECURE_URL,
                            make_canonical_urlargd({'referer' : CFG_SITE_SECURE_URL + req.unparsed_uri, 'ln' : args['ln']}, {}))
                                            , norobot=True)
                    else:

                        return page_not_authorized(req, "../submit",              # <-- this is executed
                                                   uid=uid,
                                                   navmenuid='submit')
                return home(req,catalogues_text, c,ln)
```

and then page_not_authorized (from webuser) is called:

```
[...]
            if res and res[0][0]:
                if text:
                    body = text
                else:
                    body = "%s %s" % (CFG_WEBACCESS_WARNING_MSGS[9] % cgi.escape(res[0][0]),                # <-- this is executed
                                      ("%s %s" % (CFG_WEBACCESS_MSGS[0] % urllib.quote(referer), CFG_WEBACCESS_MSGS[1])))
[...]
```

then from access_control_config: CFG_WEBACCESS_WARNING_MSGS[9] = Account '%s' is not yet activated.

egabancho:
- BETTER Changes error message depending on
  `CFG_ACCESS_CONTROL_LEVEL_ACCOUNTS` value as the meaning of the column
  `note` also changes with it. (closes #1414)

Signed-off-by: Esteban J. G. Gabancho esteban.gabancho@gmail.com

### DIFF
--- a/modules/bibsched/lib/tasklets/bst_create_icons.py
+++ b/modules/bibsched/lib/tasklets/bst_create_icons.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 ##
 ## This file is part of Invenio.
-## Copyright (C) 2013 CERN.
+## Copyright (C) 2013, 2016 CERN.
 ##
 ## Invenio is free software; you can redistribute it and/or
 ## modify it under the terms of the GNU General Public License as
@@ -18,7 +18,7 @@
 ## 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
 """Invenio Bibliographic Tasklet for generating subformats.
-   Usange:
+   Usage:
    $bibtasklet -N createicons -T bst_create_icons -a recid=123 -a icon_sizes=180,640,1440
    $bibtasklet -N createicons -T bst_create_icons -a "collection=ABC Photos" -a icon_sizes=180,640,1440"""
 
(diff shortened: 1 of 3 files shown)


tiborsimko:
The branch is made against `legacy`, the milestone is `v1.2`, and the patch applies also to `v1.1`... I guess it could go to the latter. Which oldest version have you check this against?

tiborsimko:

PR #3423, opened 2015-08-04
jsonref: Addition of invenio_jsonref in config

Theer108:
Adds invenio_jsonref to config to autodetect jsonloaders

Signed-off-by: Łukasz Chrząszcz <l.chrzaszcz@gmail.com>

### DIFF
--- a/invenio/base/config.py
+++ b/invenio/base/config.py
@@ -69,6 +69,7 @@ PACKAGES = [
     'invenio_oauthclient',
     'invenio_groups',
     'invenio_accounts',
+    'invenio_jsonref',
     'invenio.base',
 ]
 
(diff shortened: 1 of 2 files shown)


invenio/base/config.py
```diff
@@ -69,6 +69,7 @@
     'invenio_oauthclient',
     'invenio_groups',
     'invenio_accounts',
+    'invenio_jsonref',
```
Theer108: Should it be here? I want to allow autodetection of jsonloaders via registry.

jirikuncar:
@Theer108 IMHO we don't need it in Invenio itself. Invenio-Records, Invenio-JSONSchemas or Invenio-Documents should depend on the package.

jirikuncar:
Solved in Invenio-Records for Invenio 3.0.

PR #3443, opened 2015-08-14
Docextract/Refextract/Authorextract removal

jalavik:
Removes docextract/refextract/authorextract.

Depends on bibedit and bibcatalog removal due to imports from those modules. Also some useful functions where moved into classifier module which uses them.

@jirikuncar @jmartinm


jalavik:
There is also a dependency on refextract inside search unit rawref.py. But I guess this will be taken away with the search engine externalization? @jirikuncar

jirikuncar:
@jalavik `rawref.py` is not used with ElasticSearch. You can remove `refextract`.

jalavik:
@jirikuncar Ready to go :+1:

PR #3444, opened 2015-08-14
legacy: removal of bibconvert module

jirikuncar:
* INCOMPATIBLE Removes bibconvert module. (addresses #3233)


jirikuncar:
@jalavik @tiborsimko are we good to go with removin BibConvert?

jalavik:
@jirikuncar I guess, but we should somehow retain the XSLT and/or KB work - as separate thing perhaps.

tiborsimko:
The two biggest users of BibConvert were WebSubmit and OAIHarvest.  The removal is relatively safe WRT WebSubmit that is gone, but we may need to be careful WRT new OAI harvester.  (usually connected to XSLT converting mode that @jalavik mentioned)

jalavik:
So there are some sample workflows in `invenio-oaiharvester` from `invenio-workflows` which use `convert()` from BibConvert XSLT engine (but it is not a hard requirement). We can rewrite those to use the xslt directly via some utility function, so should be fine. 

My main worry is the actual XSLT files which we should keep somewhere.

kaplun:
You mean the files under https://github.com/inveniosoftware/invenio/tree/master/invenio/modules/converter/converterext/templates ? Since they are all specific to MARCXML, shoudn't they be moved somhere into invenio-demosite?

jalavik:
@kaplun Yes, indeed that could be an option. Those who'd want to use it would have to copy them to their own overlay, but it could suffice for now.

PR #3447, opened 2015-08-17
docs: update node installation instructions

jacquerie:
Chris Lea's PPA has been deprecated since March 2015: https://nodesource.com/blog/nodejs-v012-iojs-and-the-nodesource-linux-repositories#the-nodesource-linux-repositories.

Also includes a defensive change for the Dockerfile, so that if in the future the version downloaded from `https://deb.nodesource.com/setup` changes we won't have problems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,7 +29,7 @@ RUN apt-get update && \
     apt-get -qy install --fix-missing --no-install-recommends \
         curl \
         && \
-    curl -sL https://deb.nodesource.com/setup | bash -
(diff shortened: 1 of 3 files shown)


jirikuncar:
We should update released versions too. Please rebase on top of `maint-2.0` and re-push to the same PR. Thanks

jacquerie:
> Please rebase on top of maint-2.0 and re-push to the same PR.

Done, I think, in the sense that I `git rebase -i maint-2.0` my patch and deleted all commits except the last two. As far as I know there's no way to change Github's target branch; is this enough?

PR #3450, opened 2015-08-18
WIP search: allow splitting records in ES indices

jmartinm:

### DIFF
--- a/invenio/ext/es.py
+++ b/invenio/ext/es.py
@@ -21,194 +21,29 @@
 
 from __future__ import absolute_import
 
+import json
+import six
+
 from elasticsearch import Elasticsearch
 
 from elasticsearch.connection import RequestsHttpConnection
 
+from invenio.base.globals import cfg
 from invenio.celery import celery
+from invenio.modules.search.api import Query
+from invenio.modules.search.registry import mappings
 
 es = None
 
-SEARCH_RECORD_MAPPING = {
-    "settings": {
-        "analysis": {
-            "filter": {
-                "asciifold_with_orig": {
-                    "type": "asciifolding",
-                    "preserve_original": True
-                },
 
-                "synonyms_kbr": {
-                    "type": "synonym",
-                    "synonyms": [
-                        "production => creation"
-                    ]
-                }
-            },
-            "analyzer": {
-                "natural_text": {
-                    "type": "custom",
-                    "tokenizer":  "standard",
-                    "filter": [
-                        "asciifold_with_orig",
-                        "lowercase",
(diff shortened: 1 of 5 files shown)


jmartinm:
Some comment on this PR:

* I did not see in the [current search API](https://github.com/inveniosoftware/invenio/blob/master/invenio/modules/search/api.py#L57) if several collections can be selected to search on.
    * This means that we would not be able with this PR to do a federated search in all our ES indices.

kaplun:
> I did not see in the current search API if several collections can be selected to search on.
> This means that we would not be able with this PR to do a federated search in all our ES indices.

@tiborsimko, indeed multiple selected collections were possible in Invenio legacy (and CDS rely on it, and we were thinking of supporting federated search in INSPIRE too). Shall we create dedicated issue?

tiborsimko:
WRT multiple selection of collections prior doing search, two comments: 

1. The general thinking was that this isn't needed to do before doing search, thanks to intelligent collection faceting the users can explore collections (and more) after search. 

2. A custom front-end UI, such as INSPIRE's "Easy search", can create custom entry page using e.g. checkboxes and generate query with ` ... AND (collection:"Articles & Preprints" OR collection:"Multimedia & Videos")` based on user's choices, which is now easily possible including access control checking etc. IOW, there should be no need to specify extra collection argument, since "virtually everything" can be passed in the search pattern argument.

WRT federated search of distributed digital library instances, this would be better addressed "on top", not "inside" the module.

kaplun:
> WRT federated search of distributed digital library instances, this would be better addressed "on top", not "inside" the module.

Sorry, I guess we used the wrong word here. We didn't mean `federated` as in across multiple digital libraries, rather, across multiple document types (@jmartinm PR is in fact proposing to implement recommended ES best practices of indexing each document type, say authors, institutions, papers, into dedicated ES index) from which stems the question on how the search APIs allows to search across multiple ES indexes. (while Invenio collections can indeed be simply considered as facets as you propose).

jirikuncar:
@jmartinm please re-open in `invenio-search` repo.

jmartinm:
@tiborsimko as @kaplun mentions, the aim of this PR is to use best practices of ElasticSearch ( https://www.elastic.co/blog/great-mapping-refactoring) in order to cleanly separate different document types into different ES indices.

This allows to specify a mapping per index and avoid conflicts between fields that are named in the same way in different document types.

jmartinm:
To be continued in

https://github.com/inveniosoftware/invenio/pull/3458

and

https://github.com/inveniosoftware/invenio-search/pull/2

PR #3465, opened 2015-08-24
BibFormat: add current year to copyright

ludmilamarian:
* BETTER Adds current year to the copyright statement.

### DIFF
--- a/modules/bibformat/lib/elements/bfe_copyright.py
+++ b/modules/bibformat/lib/elements/bfe_copyright.py
@@ -3,7 +3,8 @@
 # $Id$
 #
 # This file is part of Invenio.
-# Copyright (C) 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2012, 2013, 2014 CERN.
+# Copyright (C) 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011,
+# 2012, 2013, 2014, 2015 CERN.
 #
 # Invenio is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License as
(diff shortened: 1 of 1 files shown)


tiborsimko:
(1) Is this safe to do it *automatically* for every site and every record, even when there may not have been any updates to the concerned records?  If a site contains say archived videos from 2008, that were never touched afterwards, do we want to show the current year in addition as well? 

Motivation: for the source code, we only amend copyright years when the given file is visited.  See also [these musings about US copyright law](http://www.contentious.com/2007/01/07/copyright-notice-is-the-year-really-necessary/).  Haven't look more closely though.

(2) If we merge "as is", then it may be good to clarify in the commit message that this concerns one particular formatting element, say "BETTER The formatting element `bfe_copyright` now automatically displays current year in the copyright statement".

(3) Shall we merge to `maint-1.2` (as per milestone v1.2.2) or to `legacy`?

ludmilamarian:
(1) In CERN's case, Tim has been in contact with the legal service, and the requirement/advice is that we need to assert the copyright in every year since creation (e.g: ©2012-2015 CERN), and as far as I understood, the modification date of the record does not count.

(2) Correct! Reading it again I see that it's really unclear what is affected by the change. I will amend the message.

(3) It should be `legacy`, I got a bit confused by the milestones. I will fix this.

TimSmithCH:
Concerning (1), I admit that the advice we are getting, from different lawyers for different use cases (software, data, web pages, multimedia) is not entirely consistent! However this seems to be the nature of the beast... I will non the less bring this up with our legal service to see if we wish to apply a local rule more consistently...

tiborsimko:
OK, let's merge now, and if some non-CERN installation of Invenio uses `bfe_copyright` with different legal conventions, let's parametrise its behaviour then.

PR #3479, opened 2015-09-01
BibIndex: fix computation of records to be indexed

ludmilamarian:
- FIX The function find_affected_records_for_index() should
  calculate the affected records based on the `recIDs` argument,
  and not add additional records to the affected ones.


ludmilamarian:
@tiborsimko Can you please take a look at this PR? As far as I've seen you have added the sql statement that I'm modifying.

tiborsimko:
I had a quick look, it seems erring on the safer sode to check potentially all 005-uploaded recordsz moreover the usual records should be OK due to modification check... To be seen in detail for v1.2.3.

PR #3485, opened 2015-09-03
babel: fallback to NullTranslations

jirikuncar:
* BETTER Improves translations lookup fallback when no compiled catalog
  is found.

--
cc @hachreak


hachreak:
tested with https://github.com/jirikuncar/invenio-accounts/pull/1 and https://github.com/inveniosoftware/invenio/pull/3486

PR #3486, opened 2015-09-03
global: optional groups field for users logged in

hachreak:
* Sets group list on logged in user as not mandatory field.


invenio/ext/login/legacy_user.py
```diff
@@ -156,8 +156,9 @@ def _login(self, uid, force=False):
             data['family_name'] = user.family_name or ''
             data['email'] = user.email or ''
             data['note'] = user.note or ''
-            data['group'] = map(lambda x: x.group.name,
-                                user.groups or [])
+            if(hasattr(user, 'groups')):
+                data['group'] = map(lambda x: x.group.name,
+                                    user.groups or [])
```
jirikuncar: `getattr(user, 'groups', [])`

PR #3497, opened 2015-09-09
i18n: initial message catalogues

tiborsimko:
* Creates initial message catalogues after separation of `invenio-base`
  module.  This fixes installation troubles related to `python setup.py
  compile_catalog`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 *~
 *.swm
 *.swn
+*.mo
 .coverage
 .eggs
 .kdev4/
--- a/.kwalitee.yml
+++ b/.kwalitee.yml
@@ -86,6 +86,7 @@ components:
 - submit
 - sword
 - tags
+- tests
 - template
 - textminer
 - tickets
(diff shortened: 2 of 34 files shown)


jirikuncar:
@tiborsimko it doesn't solve problem with translations but only installation. The translations are not going to be loaded, because `invenio` is not listed in `PACKAGES`. Please revert these changes as `invenio.{ext,utils}` will be separated in upcoming days. Thanks

The correct temporary solution would be removing following lines:
- https://github.com/inveniosoftware/invenio/blob/master/setup.cfg#L7-L8
- https://github.com/inveniosoftware/invenio/blob/master/setup.py#L48-L62
- https://github.com/inveniosoftware/invenio/blob/master/setup.py#L334-L337

tiborsimko:
FWIW  I preferred to keep the catalogues generated until `invenio-ext` and `invenio-utils` are separated out: just a matter of dependency, i.e. first remove I18N phrases then remove their catalogues.  But I fully agree that we won't need them anymore after packages are separated... so better speed these up.

PR #3504, opened 2015-09-15
utils: removal of plotextractor

### LINKED ISSUE #3482: RFC plotextractor utilities
The module `invenio-oaiharvester` is the only one that use `invenio.utils.plotextractor`.
In the [roadmap](https://github.com/inveniosoftware/invenio/issues/3233) for legacy removal  these utilities are in the list.
What's the plan?
 1) Do we move the utilities inside `invenio-oaiharvester`?
 2) Do we remove from the #3233 list and leave inside invenio?
 3) Do we move into a separete package?
 4) .....

hachreak:
* INCOMPATIBLE Removes plotextractor utility. (addresses #3233)
  (closes #3482)


setup.py
```diff
@@ -293,7 +293,7 @@ def run(self):
     platforms='any',
     entry_points={
         'console_scripts': [
-            'plotextractor = invenio.utils.scripts.plotextractor:main',
+            'inveniomanage = invenio.base.manage:main',
```
jirikuncar: please remove this line

PR #3509, opened 2015-09-17
elasticsearch: virtualenv support

kaplun:
* BETTER When developing and using honcho to start elasticsearch, executes
  elasticsearch by storing indexes and logs within the local virtual
  environment.


jirikuncar:
> When developing and using honcho ...

@kaplun because it's only for developers this is a feature should not be reported in RELEASE-NOTES, hence we don't need `BETTER` in commit message.

kaplun:
OK. Wasn't sure about it (since most of what we announce is indeed geared towards developers rather than admins). I have amended the commit log.

jirikuncar:
Please rebase and we can :shipit:

kaplun:
Done!

PR #3511, opened 2015-09-17
tests: demo_record_marc_data.xml into demosite

kaplun:
* Moves demo_record_marc_data.xml from invenio repository
  to invenio-demosite where it really should belong to.

* Amends install guide to describe the new "inveniomanage records"
  based command to load demo records.


kaplun:
See symmetric inveniosoftware/invenio-demosite#81

jirikuncar:
Demo data has been moved to `invenio-records` as they are required by tests there. Please amend commit message to reflect this change. Thanks

kaplun:
Since the demo records are now moved into `invenio-records`, and this PR was also about updating documentation, @jirikuncar just tell me what should go into the `INSTALL.rst` for a user to load demo records nowadays. Thanks.

INSTALL.rst
```diff
@@ -564,7 +564,8 @@ When all the servers are running, it is possible to upload the demo records.
 
     $ # in a new terminal
     $ workon invenio
-    (invenio)$ inveniomanage demosite populate --packages=invenio_demosite.base
+    (invenio)$ cdvirtualenv src/invenio-demosite
+    (invenio)$ inveniomanage records create -t marcxml < invenio_demosite/testsuite/data/demo_record_marc_data.xml
```
jirikuncar: For now the file is in both places ...

kaplun: So the commit message is still consistent.

jirikuncar: IMHO we can remove this part from here completely as it should be part of `invenio-records` documentation and here we should just point there.

jirikuncar:
Solved in Invenio 3 installation guide.

PR #3513, opened 2015-09-18
legacy: removal of hstRECORD table

jirikuncar:
* INCOMPATIBLE Removes legacy hstRECORD table definition.


egabancho:
Naive question, how are we going to keep the different versions of records?

jirikuncar:
@egabancho on `before_update` we create a copy or we use versionable behavior from SQLAlchemy.

kaplun:
Other naive question: if you delete the model, how will we be able to migrate the history to the new support that will come?

egabancho:
The model is deleted, but the table will remain there :wink:

kaplun:
Aren't models needed for querying the table? Or is done directly with `run_sql`?

egabancho:
@kaplun I guess `run_sql` should be good enough

PR #3380, opened 2015-07-20
docker: multiple optimizations

crepererum:
See commits for details.

Issue #3286, opened 2015-06-17
docker: common/dev/debug/prod configurations


tiborsimko:
Currently, we advertise how to develop Invenio 2.x via Docker in our documentation, using DEV configuration. However, when somebody goes that way, he/she would find that running test suite reports many failures due to HTTP redirection associated with debugging mode.  This may be a rather surprising default...

I'd rather propose to strive at the defaults whether the following would hold true:

```
$ git checkout maint-2.0
$ docker-compose build
$ docker-compose up
$ docker exec -i -t invenio_web_1 python setup.py test

[...]

Ran 977 tests in 701.359s

OK (skipped=6)
```

(This is the case for Invenio 1.x Docker configurations, for our Flask-* and other projects...)

We could achieve this by reconsidering DEV -> DEBUG -> PROD modes, choosing a conservative default, and [extending our configurations progressively](https://blog.docker.com/2015/04/easily-configure-apps-for-multiple-environments-with-compose-1-2-and-much-more/) for various use case scenarios. For example:

- COMMON profile: create common Docker Compose configuration as `docker-compose-common.yml` that will be used as a base for everything;
- DEV profile [*default*]: create `docker-compose.yml` that will extend COMMON to make Invenio runnable in a DEV friendly mode, without extra debugging and stuff, but mirroring what Travis/Jenkins do; useful for regular CI/kwalitee purposes;
- DEBUG profile: if wanted, create another `docker-compose-debug.yml` that would extend COMMON/DEV with human-related DEBUG capabilities, with Flask-DebugToolbar/Werkzeug and friends on; either create new profile if needed or just pass appropriate environment variables to the DEV profile to achieve the same;
- PROD profile: create yet another `docker-compose-production.yml` that will extend COMMON in a way that will be useful for composing production Invenio environments with overlays. (TBD)

WDYT? CC @crepererum @jirikuncar

lnielsen:
Could be useful to have a full prod setup (haproxy, 2 web nodes, cache, mq, db master, db slave, task node).

jirikuncar:
Can we just improve our testing suite that checks during `setUp` that `current_app.debug` is `False` and if it's not it warns user and set `current_app.debug = False`.

tiborsimko:
@jirikuncar This would indeed take care of separate DEV/DEBUG profiles, so that only one configuration is sufficient.  And if we switch defaults to DEV mode, leaving PROD mode aside, then this should do, I think.

[...]

PR #3514, opened 2015-09-18
WIP docker: splits the invenio dockerfile

ddaze:
* BETTER Splits the invenio dockerfiles in one generic without code
  and a dockerfile for a basic Invenio installation.

* Moves the invenio files from `/home/invenio` to `/opt/invenio`
  and the static files (bower) to `/opt/invenio_static`.

* Moves all code and also packages that get installed in editable mode
  from `/code` to `/src`

* Installs vim-tiny and adds a symlink it to vim.

* Outputs more information at creation time and bower install.

* Adds `libjpeg-dev` for the pillow jpeg support.

Signed-off-by: David Zerulla <ddaze@outlook.de>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,80 +19,25 @@
 ## 1. Base (stable)                                                          ##
 ###############################################################################
 
-FROM python:2.7-slim
-MAINTAINER CERN <info@invenio-software.org>
+FROM ddaze/invenio-base
 
-# expose the right port
-EXPOSE 28080
-
-# add invenio user
-RUN useradd --home-dir /home/invenio --create-home --shell /bin/bash --uid 1000 invenio
-
-# set work dir
-WORKDIR /code
-
-# iojs, detects distribution and adds the right repo
-RUN apt-get update && \
-    apt-get -qy install --fix-missing --no-install-recommends \
-        curl \
-        && \
-    curl -sL https://deb.nodesource.com/setup_iojs_2.x | bash -
-
-#  - package cache updates
-#  - install requirements from repos
-#  - clean up
-#  - install python requirements
-#  - install iojs requirements
-RUN apt-get update && \
-    apt-get -qy upgrade --fix-missing --no-install-recommends && \
-    apt-get -qy install --fix-missing --no-install-recommends \
-        gcc \
-        git \
-        iojs \
-        libffi-dev \
-        liblzma-dev \
-        libmysqlclient-dev \
-        libssl-dev \
-        libxslt-dev \
-        mysql-client \
-        poppler-utils \
-        sudo \
-        && \
-    apt-get clean autoclean && \
-    apt-get autoremove -y && \
-    rm -rf /var/lib/{apt,dpkg}/ && \
-    (find /usr/share/doc -depth -type f ! -name copyright -delete || true) && \
-    (find /usr/share/doc -empty -delete || true) && \
-    rm -rf /usr/share/man/* /usr/share/groff/* /usr/share/info/* && \
-    pip install --upgrade pip && \
-    pip install ipdb \
-        ipython \
-        mock \
-        unittest2 \
-        watchdog \
-        && \
-    npm update && \
-    npm install --silent -g bower less clean-css uglify-js requirejs
-
+WORKDIR /src/invenio
 
 ###############################################################################
 ## 2. Requirements (semi-stable)                                             ##
 ###############################################################################
 
 # global environment variables:
-#  - docker specific paths for Invenio
 #  - proper requirements level
 #  - Python extras
-ENV INVENIOBASE_INSTANCE_PATH="/home/invenio/instance" \
-    INVENIOBASE_STATIC_FOLDER="/home/invenio/static" \
-    REQUIREMENTS="devel" \
+ENV REQUIREMENTS="devel" \
     REXTRAS="development,docs"
 #ENV REQUIREMENTS lowest
 #ENV REQUIREMENTS release
 
 # add requirement files
-COPY ./requirements-devel.txt ./requirements.py ./setup.cfg ./setup.py /code/
-COPY ./invenio/version.py /code/invenio/version.py
+COPY ./requirements-devel.txt ./requirements.py ./setup.cfg ./setup.py /src/invenio/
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -57,6 +54,7 @@ RUN apt-get update && \
         mysql-client \
         poppler-utils \
         sudo \
+        vim \
```
jirikuncar: why `vim` and not `emacs`? (cc @tiborsimko)

tiborsimko: Indeed, personal preferences such as editors or "ll" shell aliases (see below) should not be forced upon people...

Dockerfile
```diff
@@ -74,6 +72,11 @@ RUN apt-get update && \
     npm update && \
     npm install --silent -g bower less clean-css uglify-js requirejs
 
+# add invenio user
+RUN useradd --home-dir /home/invenio --create-home --shell /bin/bash --uid 1000 invenio && \
+    echo "alias ll='ls -alF'" >>  /home/invenio/.bashrc && \
+    echo "invenio ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
```
tiborsimko: Giving `invenio` superuser rights is dangerous.  It may be good for development, but it's bad for production.  Note that many services build on top of Invenio docker image, see `FROM invenio:2.0` and friends in our overlays. (e.g.  `invenio-demosite`, `opendata.cern.ch`, `analysis-preservation.cern.ch`)

scripts/setup_devmode.sh
```diff
@@ -44,9 +44,10 @@ CFG_TMPDIR = u'/tmp'
 COLLECT_STORAGE = u'flask_collect.storage.link'
 EOF
 
+echo "Building Assets, bower install"
```
tiborsimko: Maybe just call the message `inveniomanage bower/collect/assets` for the sake of harmony with earlier `inveniomanage database create` messages.

Dockerfile
ddaze: That`s true, I didn`t thought about that..
But there is no editor installed so which one should be installed?
vi? or nano? Like as the default on many Linux distributions.

Dockerfile
ddaze: Makes sense, I will move it to the development image.

Dockerfile
tiborsimko: I think none should be installed by default.  Either people can edit files in their host OS and see them changed in the running container thanks to volume mounting, or they can do something like `docker exec -i -t --user root myapp_web_1 /bin/bash` and, when inside, then `apt-get install vim` as usual.

tiborsimko: BTW, one more thing: if you are looking at amending the development setup after the recent Invenio package split, then please see my previous comments about possibly homogenising `/code/invenio` and `/code/src/*`:

- https://github.com/inveniosoftware/invenio/pull/3380#issuecomment-138509850

E.g. my first comment there seems to touch your file ownership vs sudo righs issue.

One possible option coud be to have a development setup where people will check out on their host systems all the independent packages:

```
src/invenio-records
src/invenio-documents
```

and we could directly mount them as docker volumes inside the container, and depend on `../invenio-foo/invenio_foo/` like package paths for easy editing.

P.S. See also some previous debug/devel/prod docker setup musings at:

- https://github.com/inveniosoftware/invenio/issues/3286

scripts/setup_devmode.sh
```diff
@@ -44,9 +44,10 @@ CFG_TMPDIR = u'/tmp'
...
 # fetch and build assets
 inveniomanage bower -i bower-base.json > bower.json
-bower install --silent
+bower install --quiet
```
crepererum: do we really need the `silent`/`quiet` mode? It makes debugging of dependency issues way harder.

jirikuncar: there was an issue with travis log length

Dockerfile
crepererum: @tiborsimko 

> or they can do something like `docker exec -i -t --user root myapp_web_1 /bin/bash` and, when inside, then `apt-get install vim` as usual.

Not possible because the images are mounted read-only by default. I vote for installing `nano` as a neutral, minimal editor. Right now I'm using `sed` hacks as a work-around and it is really difficult to do file modifications (for config changes or when inserting trace-points in 3rd-party libraries).

scripts/setup_devmode.sh
crepererum: There IS an issue. Right now, you have to download the log anyway to see the full test results.

jirikuncar: The problem was that the log was cut after 10MB because of bower, pip and friends.

crepererum: I think most of our logs are spilled by (mostly useless) test outputs which btw also makes it hard to read the test results.

jirikuncar: Since we don't have almost any test left in Invenio anymore we can enable verbose output.

Dockerfile
```diff
@@ -57,6 +57,7 @@ RUN apt-get update && \
         mysql-client \
         poppler-utils \
         sudo \
+        nano \
```
tiborsimko: Ouch! :smile: Instead of `nano` let's rather use `vim-tiny`...

Dockerfile
tiborsimko: > Not possible because the images are mounted read-only by default.

IIRC I've been modifying stuff like this: just "login" to a container as root, remount in "rw" mode, and install what is wanted.

Dockerfile
ddaze: Okay, I added vim-tiny and added a symlink to vim.

ddaze:
It might make sense to move the bower downloads outside of the `docker_boot.sh` script?

Usage of the invenio-base image can be seen in PR https://github.com/CERNDocumentServer/cds/pull/51.

jirikuncar:
@ddaze the base Docker image should probably go to *Invenio-Base* package.

ddaze:
Okay, and there than in the sub folder docker-base?
Or even as a own repository?

ddaze:
Is not longer relevant. (See other Docker improvements)

PR #2169, opened 2014-08-28
BibIndex: improvements for 1.x

fpoli:
Corrections from PR https://github.com/inveniosoftware/invenio/pull/2117
Includes commits from PR https://github.com/inveniosoftware/invenio/pull/1840

Issue #1862, opened 2014-06-23
BibAuthorID: broken disambiguation on Atlantis demo site


tiborsimko:
BibAuthorID stopped working for Atlantis demo site.

Firstly, collection treatment should be amended:

```
$ diff -u modules/bibauthorid/lib/bibauthorid_config.py /opt/invenio/lib/python/invenio/
--- modules/bibauthorid/lib/bibauthorid_config.py       2014-06-20 14:43:36.823044058 +0200
+++ /opt/invenio/lib/python/invenio/bibauthorid_config.py       2014-06-23 16:06:45.973087504 +0200
@@ -92,8 +92,10 @@
 # Special for ADS: Focus on ASTRONOMY collection
 if GLOBAL_CONFIG and CFG_ADS_SITE:
     LIMIT_TO_COLLECTIONS = ["ASTRONOMY"]
-else:
+elif CFG_INSPIRE_SITE:
     LIMIT_TO_COLLECTIONS = ['HEP']
+else:
+    LIMIT_TO_COLLECTIONS = []

 # Exclude documents that are visible in a collection mentioned here:
 EXCLUDE_COLLECTIONS = ["HEPDATA","Data", "HEPNAMES", "INST", "Deleted", "DELETED", "deleted"]
```

Secondly, location of Tortoise cache should be put under `/opt/invenio/var/cache/bibauthorid/tortoise_cache` by default.  Currently one has to do:

```
$ sudo mkdir /opt/tortoise_cache/
$ sudo chown www-data.www-data /opt/tortoise_cache/
```

Thirdly, the first process runs fine:

```
$ sudo -u www-data /opt/invenio/bin/bibauthorid -u admin --update-personid --all-records -v 9
```

and populates well tables:

```
mysql> SELECT * FROM aidPERSONIDPAPERS LIMIT 5;
+----------+--------------+--------------+--------+-------------------+------+------+---------------------+
| personid | bibref_table | bibref_value | bibrec | name              | flag | lcul | last_updated        |
+----------+--------------+--------------+--------+-------------------+------+------+---------------------+
|        8 | 100          |            1 |      1 | Photolab          |    0 |    0 | 2014-06-23 16:06:53 |
|        9 | 100          |            2 |      2 | Loïez, Patrice    |    0 |    0 | 2014-06-23 16:06:53 |
|       10 | 100          |            3 |      4 | Caron, J. L.      |    0 |    0 | 2014-06-23 16:06:54 |
[...]

PR #3520, opened 2015-09-21
Master bibindex improvements

kaplun:
Superseeds #2169


config/invenio.conf
```diff
@@ -492,6 +492,12 @@ CFG_WEBSEARCH_RSS_I18N_COLLECTIONS =
 # a feed cache is valid.
 CFG_WEBSEARCH_RSS_TTL = 360
 
+# CFG_WEBSEARCH_INSTITUTION_COLLECTIONS -- comma-separated list of
+# collections that contains records of institutions.
+# It is used by BibIndexCountryTokenizer to know in which collections the
+# institution's name should be searched.
+CFG_WEBSEARCH_INSTITUTION_COLLECTIONS = Institutes,Authority Institution,Institutions
```
tiborsimko: We only have "Institutes" as the demo collection name.  The others could be set in INSPIRE overlay.

BTW, based on the CFG description, shouldn't it be named `s/CFG_WEBSEARCH_/CFG_BIBINDEX/`?  But then again we've been moving all configurations into DB, deprecating ones in `invenio.conf`, which this addition seems to go against...

modules/bibconvert/bin/Makefile.am
```diff
@@ -15,7 +15,7 @@
 # along with Invenio; if not, write to the Free Software Foundation, Inc.,
 # 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
 
-bin_SCRIPTS = bibconvert
+bin_SCRIPTS = bibconvert kbload
```
tiborsimko: `kbload` should live in `modules/bibknowledge`, not in `modules/bibconvert`.

modules/miscutil/demo/demobibdata.xml
```diff
@@ -26288,4 +26288,56 @@ Only the mountain and I.
...
+    <datafield tag="245" ind1=" " ind2=" ">
+      <subfield code="a">Additional data from: Probing color coherence effects in pp collisions at $\sqrt{s}=7\,\text {TeV}$</subfield>
+    </datafield>
+    <datafield tag="786" ind1=" " ind2=" ">
+      <subfield code="w">10</subfield>
+    </datafield>
+    <datafield tag="980" ind1=" " ind2=" ">
+      <subfield code="a">ARTICLE</subfield>
+    </datafield>
+  </record>
```
tiborsimko: Is it OK that this data record is classified as an ARTICLE?  Shouldn't this be a data collection on its own?

modules/miscutil/lib/inveniocfg.py
```diff
@@ -879,21 +885,25 @@ def cli_cmd_load_demo_records(conf):
...
+                "%s/bin/bibrank 7" % CFG_PREFIX,
                 "%s/bin/bibsort -u admin -R" % CFG_PREFIX,
-                "%s/bin/bibsort 7" % CFG_PREFIX,
+                "%s/bin/bibsort 8" % CFG_PREFIX,
                 "%s/bin/oairepositoryupdater -u admin" % CFG_PREFIX,
-                "%s/bin/oairepositoryupdater 8" % CFG_PREFIX,
-                "%s/bin/bibupload 9" % CFG_PREFIX,]:
+                "%s/bin/oairepositoryupdater 9" % CFG_PREFIX,
+                "%s/bin/bibupload 10" % CFG_PREFIX,
+                "%s/bin/bibauthorid -u admin --update-personid" % CFG_PREFIX,
```
tiborsimko: `bibauthorid` shouldn't really be plugged in here?  See https://github.com/inveniosoftware/invenio/issues/1862

modules/miscutil/lib/inveniocfg.py
```diff
@@ -879,21 +885,25 @@ def cli_cmd_load_demo_records(conf):
                 "%s/bin/bibupload 1" % CFG_PREFIX,
                 "%s/bin/bibdocfile --textify --with-ocr --recid 97" % CFG_PREFIX,
                 "%s/bin/bibdocfile --textify --all" % CFG_PREFIX,
-                "%s/bin/bibindex -u admin" % CFG_PREFIX,
+                "%s/bin/bibindex -u admin -w collection" % CFG_PREFIX,
```
tiborsimko: If a change is required in index scheduling, then I guess we should update HOWTO Run and other documentation at the same time...

modules/miscutil/lib/upgrades/invenio_2014_05_26_new_index_country.py
```diff
@@ -0,0 +1,224 @@
...
+        SELECT id FROM tag
+        WHERE value = '700__u'
+        """
+    )
+    tag_110__u_id = run_sql(
+        """
+        INSERT INTO tag
+        SET
+            name = 'SPIRES name'
+            value = '110__u'
```
tiborsimko: All this section is installation-specific (e.g. "SPIRES name").  We usually leave it out in upgrade recipes, distinguishing between `do_upgrade()` and `do_upgrade_atlantis()`. See for example `invenio_2013_10_18_new_index_filetype.py` and other past upgrade recipes.  We should leave it out here too...

modules/miscutil/sql/tabfill.sql
```diff
@@ -571,6 +604,8 @@ INSERT INTO idxINDEX VALUES (25,'filetype','This index contains extensions of fi
 INSERT INTO idxINDEX VALUES (26,'miscellaneous','This index contains words/phrases from miscellaneous fields','0000-00-00 00:00:00', '', 'native','','No','No','No', 'BibIndexDefaultTokenizer');
 INSERT INTO idxINDEX VALUES (27,'doi','This index contains words/phrases from doi fields','0000-00-00 00:00:00', '', 'native','','No','No','No', 'BibIndexDOITokenizer');
 INSERT INTO idxINDEX VALUES (28,'filename','This index contains file names of files connected to records.','0000-00-00 00:00:00', '', 'native', '','No','No','No', 'BibIndexFilenameTokenizer');
+INSERT INTO idxINDEX VALUES (29,'country','This index contains country names of the affiliated institutes of the authors.','1970-01-01 00:00:00', '', 'native', '','No','No','No', 'BibIndexCountryTokenizer');
+INSERT INTO idxINDEX VALUES (30,'journalpage','This index contains words/phrases from the journal page field.','1970-01-01 00:00:00', '', 'native', '','No','No','No', 'BibIndexJournalPageTokenizer');
```
tiborsimko: While the "country" index sounds very useful, do you expect users to use "journalpage" index without corresponding "journaltitle" and "journalvolume" or "journalyear"?  That's why we had standardisation of everything into the unique "journal" index following standard form as defined by `CFG_JOURNAL_PUBINFO_STANDARD_FORM` conventions...

kaplun: The `journalpage` index was necessary in order to have fancy searches that were able to cope with query having page ranges (or simple pages) against records having page ranges (or simple pages). So in this sense it is not used by users, but by scripts (such as bibrank/refextract) in order to do smarter matching.

modules/miscutil/lib/inveniocfg.py
kaplun: I guess was due to rebase of rebase of rebase glitch.

modules/miscutil/lib/inveniocfg.py
kaplun: Nope. Looks like it was just an intended speed-up during installation.

PR #3525, opened 2015-09-29
WIP: docker: moves all code from `/code` to `/src/`

ddaze:
* Moves all code and also packages that get installed in editable mode
  from `/code` to `/src`

CC: @tiborsimko @crepererum


crepererum:
Had a chat with @ddaze and we addressed some issues and changed the design a little bit. New version of the PR will follow.

ddaze:
See #3514.

PR #3541, opened 2015-10-20
WebJournal: get journal categories with 0 articles

### LINKED ISSUE #3539: WebJournal: Add an option to get the journal categories even when they don't have any articles in the given issue
in `webjournal_utils.py` the `get_journal_categories` function will return the categories in the given journal and for the given issue (if provided). If, however, no articles are found in the given category, it won't be returned. An option should be added (`boolean` argument) to allow for categories that have no articles to be returned too.

kasioumis:
* Adds a boolean argument `include_categories_with_no_articles` to
  `get_journal_categories`. If `True` then even categories with no
  articles will be returned.
  (closes #3539)

Signed-off-by: Nikos Kasioumis <kasioumis@gmail.com>


tiborsimko:
Thanks, merged as 2114da746d35c51c978be5665d1acf5fa19e9c6c.

PR #3547, opened 2015-10-27
collect: remove duplicated link storage

greut:
Removes some duplicated code that has been integrated into Flask-Collect since.


jirikuncar:
@greut there is a warning message. We can't remove files from released version for compatibility reasons.

greut:
k, I see.

PR #3551, opened 2015-11-02
installation: fix entry_points for pip7

greut:
Fixes the `entry_point` as pip 7.x expects a function and not a module.


greut:
:dancers: !

PR #3552, opened 2015-11-04
WebSearch: fix the order of prev and next links

ludmilamarian:
* FIX Fixes the order of search results when computing the
  next and previous search links on the record page by not
  reversing the initial search results.


tiborsimko:
Thanks, merged as 0df75fea4bda6025d87fb0bbd2ad865c2bb95d7f.

PR #3564, opened 2015-11-12
WebSearch: consider sort order when searching

ludmilamarian:
- FIX Take into account the sort order when searching, even if
  no specific sort or rank method is used (issue introduced in
  e49a1368bb0e4159b3a43b508140db120c2e8e82).


tiborsimko:
Closed by  #3729.

PR #3570, opened 2015-11-20
installation: latest record packages

tiborsimko:
* BETTER Amends installation scripts and documentation to use the
  recently released `Invenio-Records` and `Invenio-PIDStore` PyPI
  packages. The record IDs are now internally represented by UUIDs.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -104,9 +104,8 @@ set -o errexit
 set -o nounset
 
 # sphinxdoc-install-invenio-full-begin
-# FIXME we pin older version that works:
-pip install invenio-records==v1.0.0a3
-# FIXME we need to install PostgreSQL DB connect option explicitly:
+# FIXME we need to install PostgreSQL DB connect option explicitly
+# (for Invenio versions less than v3.0.0a3)
 pip install invenio-db[postgresql]
 # now we can install full Invenio:
 pip install invenio[full] --pre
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -159,16 +158,22 @@ celery worker -A ${INVENIO_WEB_INSTANCE}.celery -l INFO &
...
 
 # sphinxdoc-populate-with-demo-records-begin
-echo '{"title":"Invenio 3 Rocks", "recid": 1}'| \
-    python manage.py records create
+echo "from invenio_db import db; \
+from invenio_records import Record; \
+Record.create({'title': 'Invenio 3 rocks'}, \
+  id_='4b0714f6-9fe4-43d3-a08f-38c2b309afba'); \
+db.session.commit()" | python manage.py shell
 # sphinxdoc-populate-with-demo-records-end
```
tiborsimko: Some notes: 

- Using low-level Python API, which works, but which may not be the prettiest for the first-seen end-user installation example. 

- Would be nice to have some high-level CLI API such as `python manage.py records create` that would also take care of minting record IDs in PIDStore... though beware about mutual package dependencies.  

- Could switch to curl'ing JSON file to the upload REST APIs later, perhaps.

- In any case, loading records via CLI command, without having any PIDs back, does not seem to be very reuse friendly for those records... let's think about some consistent CLI API vs REST API options.

lnielsen: Yes I agree that it's somewhat cumbersome to use the CLI for ingesting records now. Not sure what is the best way forward though.

tiborsimko: One option is to: (1) keep Record Python API; (2) kill Record CLI API (i.e. `manage.py records create`) in order not to attract people to use it as their main upload API; (3) promote using REST API that will take care of uploading records and minting PIDs.

PR #3577, opened 2015-12-15
BibCirculation: BibEdit redirect

kaplun:
- No longer automatically redirects to BibEdit after an item has been
  modified. Instead displays a clickable link, so that cataloguers
  choose whether to continue editing the bibliographic information
  corresponding to this item, or rather continue modifying the item.


kaplun:
@ludmilamarian, @egabancho, @switowski this change was requested by Jens, and is actually a usability improvement. I don't have a way to test it. So would be great if you can deploy it one day to cdstest before :dancer: 

@PXke I guess you are not so much affected by this. Are you?

PXke:
No :)

egabancho:
Tested and deployed on CDS

PR #3585, opened 2016-01-18
scripts: update according to arguments unification

hachreak:
* Updates script according to last change in invenio-accounts cli.
  (addresses invenio-accounts#56)

### DIFF
--- a/scripts/populate-instance.sh
+++ b/scripts/populate-instance.sh
@@ -106,7 +106,7 @@ ${INVENIO_WEB_INSTANCE} db create
 
 # sphinxdoc-create-user-account-begin
 ${INVENIO_WEB_INSTANCE} users create \
-       --email ${INVENIO_USER_EMAIL} \
+       ${INVENIO_USER_EMAIL} \
        --password ${INVENIO_USER_PASS} \
        --active
 # sphinxdoc-create-user-account-end


jirikuncar:
@hachreak `s/accoounts/accounts `

scripts/populate-instance.sh
```diff
@@ -106,7 +106,7 @@ ${INVENIO_WEB_INSTANCE} db create
 
 # sphinxdoc-create-user-account-begin
 ${INVENIO_WEB_INSTANCE} users create \
-       --email ${INVENIO_USER_EMAIL} \
+       ${INVENIO_USER_EMAIL} \
```
tiborsimko: Maybe pass first the optional arguments and second the mandatory arguments, as is the usual style?
(`invenio3 users create --foo1 bar1 --foo2 bar2 xyzzy zyxxy`)

jirikuncar: IMHO it's just a habit. I prefer arguments first and then options.

tiborsimko: It's a widespread convention in Unix, see e.g. any system command manual page `man ls` and friends, the advertised usage synopses are `ls [options] ... <file> ...`, or `kill [options] <pid>`, etc... Kinda everywhere.

PR #3586, opened 2016-01-19
docker: explicit PATH environment settings

tiborsimko:
The explicit PATH settings should fix troubles observed by @egabancho @greut @switowski @kaplun 

@egabancho can you please test on your setup?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ ENV INVENIO_WORKER_HOST=127.0.0.1
 RUN /code/scripts/create-instance.sh --devel
 
 # Make given VENV default:
-ENV PATH=/home/invenio/.virtualenvs/invenio3/bin:$PATH
+ENV PATH=/home/invenio/.virtualenvs/invenio3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 ENV VIRTUALENVWRAPPER_PYTHON=/usr/local/bin/python
 RUN echo "source /usr/local/bin/virtualenvwrapper.sh" >> ~/.bashrc
 RUN echo "workon invenio3" >> ~/.bashrc
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -26,7 +26,7 @@ web:
   build: .
   command: /bin/bash -c "invenio3 --debug run -h 0.0.0.0"
   environment:
-    - PATH=/home/invenio/.virtualenvs/invenio3/bin:$PATH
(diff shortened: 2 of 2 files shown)


kaplun:
:+1:

switowski:
Works :)

egabancho:
:rocket:

Dockerfile
```diff
@@ -57,7 +57,7 @@ ENV INVENIO_WORKER_HOST=127.0.0.1
 RUN /code/scripts/create-instance.sh --devel
 
 # Make given VENV default:
-ENV PATH=/home/invenio/.virtualenvs/invenio3/bin:$PATH
+ENV PATH=/home/invenio/.virtualenvs/invenio3/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 ENV VIRTUALENVWRAPPER_PYTHON=/usr/local/bin/python
 RUN echo "source /usr/local/bin/virtualenvwrapper.sh" >> ~/.bashrc
```
greut: @egabancho Why not adding this into the `~/.bashrc` instead of doing this dirty hack?

```sh
RUN echo "PATH+=:/home/invenio/.virtualenvs/invenio3/bin" >> ~/.bashrc
```

tiborsimko: @greut If we only enrich `~/.bashrc` then I guess people won't be able to do things like:

```consoly
$ docker run -i -t --rm invenio_web invenio3 --help
$ docker-compose run --rm web invenio3 --help
```

without knowing where `invenio3` got exactly installed.  They would have to either launch commands via bash, or specify full path.

Hence the `PATH` enrichment bit, where the static value is not so dirty considering that it matches Python docker defaults:

```console
$ docker run -i -t --rm python:2.7-slim /usr/bin/env | grep PATH
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

Anyway, to me, the bigger question is whether to use venv inside Docker *at all*.  OT1H, this is pretty superfluous, since we are already in a nicely isolated environment. OTOH, in this way we mimick exactly what non-Docker installation examples do. So there are some pros/cons. A natural alternative would be not to use venv inside Docker and simply install the invenio3 instance directly into `/usr/local/...` place;  then we wouldn't need to play with `PATH` and shell and friends.   We briefly considered pros/cons with @jirikuncar and decided to keep venv-inside-docker for the time being... but it would be probably more natural to kill it.

greut:
@tiborsimko fair enough, it makes sense from the `docker exec` perspective.

PR #3587, opened 2016-01-20
installation: node provisioning cleanup

tiborsimko:
* Uses slim Python Docker image and cleans locally cached system
  packages that are no longer necessary after provisioning the nodes.
  Helps to reduce the size of virtual machines and of Docker containers.


tiborsimko:
@greut This reduces the size of the Docker image by ~250MB.  The size is now 1.1 GB for me.  We can go even lower, obviously, by uninstalling other vagrant-specific bits and removing logs and whatnot in case of Docker installation.

PR #3588, opened 2016-01-20
docker: rabbitmq port exposure

tiborsimko:
* Amends RabbitMQ Docker container's port exposure to use the higher
   port (by adding 20000 to the usual port), so that one can run
   docker-based installation alongside OS-based installation without
   having to shutting down any locally installed system services.
    
### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,7 +49,7 @@ web:
     - elasticsearch
     - rabbitmq
   ports:
-    - "5000:5000"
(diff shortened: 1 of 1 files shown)


tiborsimko:
- depends on https://github.com/inveniosoftware/invenio/pull/3587

- solves an issue observed by  @kaplun

audub:
This seems to break docker for OS X users. At least to all of us in TIND with mac. It also seems to work fine for @PXke without the IP.

PR #3592, opened 2016-02-03
BibCirculation: reset ILL overdue_letter_number

mvesper:
- A ILL is currently extended by manually updating the _due_date_ of
  that ILL, this leads to the following issue:
  A user receives an ILL recall letter, therefore gets the ILL extended
  at the library.
  The next ILL recall letter will be of the second category, since the
  _overdue_letter_number_ never gets reseted.
- Introduces the new function _update_ill_request_letter_number_
- Changes the _overdue_letter_number_ in case the updated due_date is a
  later date than the current due_date in
  _ill_request_details_step2_.

Signed-off-by: Martin Vesper martin.vesper@cern.ch

### DIFF
--- a/modules/bibcirculation/lib/bibcirculation_dblayer.py
+++ b/modules/bibcirculation/lib/bibcirculation_dblayer.py
@@ -2700,6 +2700,11 @@ def get_purchase_request_borrower_details(ill_request_id):
     else:
         return None
 
+def update_ill_request_letter_number(ill_request_id, overdue_letter_number):
+    query = ('UPDATE crcILLREQUEST set overdue_letter_number={0} '
+             'where id="{1}"')
+    run_sql(query.format(overdue_letter_number, ill_request_id))
+
 def update_ill_request(ill_request_id, library_id, request_date,
                        expected_date, arrival_date, due_date, return_date,
                        status, cost, barcode, library_notes):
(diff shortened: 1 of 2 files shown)


modules/bibcirculation/lib/bibcirculation_dblayer.py
```diff
@@ -2700,6 +2700,11 @@ def get_purchase_request_borrower_details(ill_request_id):
     else:
         return None
 
+def update_ill_request_letter_number(ill_request_id, overdue_letter_number):
+    query = ('UPDATE crcILLREQUEST set overdue_letter_number={0} '
+             'where id="{1}"')
+    run_sql(query.format(overdue_letter_number, ill_request_id))
```
tiborsimko: This would permit a theoretical SQL injection if `overdue_letter_number` contains an unexpected string with quotes and backslashes, for a reason or another.

Can you please use `run_sql(query, (param1, param2))` technique?  In this way the parameters will be always properly escaped.

- "Supply  your run_sql() arguments in a tuple to protect against SQL injection" 
  https://github.com/inveniosoftware/invenio/blob/maint-1.2/CONTRIBUTING.rst

modules/bibcirculation/lib/bibcirculationadminlib.py
```diff
@@ -4936,6 +4936,25 @@ def ill_request_details_step2(req, delete_key, ill_request_id, new_status,
         barcode = db.get_ill_barcode(ill_request_id)
         db.update_ill_loan_status(borrower_id, barcode, return_date, 'ill')
 
+    # ill recall letter issue
+    try:
+        from invenio.dbquery import run_sql
+        _query = ('SELECT due_date from crcILLREQUEST where id = "{0}"')
+        _due = run_sql(_query.format(ill_request_id))[0][0]
```
tiborsimko: Ditto here.

PR #3593, opened 2016-02-10
scripts: update demo records loading

hachreak:


tiborsimko:
:+1: we can do this now, but note that there is one more section in `scripts/provision-web.sh` where `uuid` needs to be removed.  (in the CentOS part)

PR #3594, opened 2016-02-23
fix for redis >= 2.4.10

vit-tucek:
Found by Samuele Kaplun.

https://github.com/andymccurdy/redis-py/blob/master/CHANGES#L303

[...]
SETEX in StrictRedis is now compliant with official Redis SETEX command.
      the name, value, time implementation moved to "Redis" for backwards
      compatibility.

### DIFF
--- a/modules/websession/lib/session.py
+++ b/modules/websession/lib/session.py
@@ -620,8 +620,8 @@ class InvenioSessionRedis(InvenioSessionBase):
 
     def save_in_storage(self, sid, session_object, timeout, uid):  # pylint: disable=W0613
         return get_redis().setex(self.generate_key(sid),
-                                 session_object,
-                                 timeout)
+                                 timeout,
+                                 session_object)
 
 if CFG_WEBSESSION_STORAGE == 'mysql':
     InvenioSession = InvenioSessionMySQL


tiborsimko:
@kaplun Are you sure about this change? In `maint-1.2`, Redis is pinned to `redis==2.9.0`, and all the tests are passing. Using this patch, one cannot login, and many tests fail...

egabancho:
This change looks odd, looking at the documentation it seems that it is ok the way it is https://redis-py.readthedocs.io/en/latest/#redis.Redis.setex

kaplun:
Assuming I wasn't crazy at the time :stuck_out_tongue_closed_eyes: maybe redis has reverted its API change in subsequent versions.

kaplun:
It probably also depends on the version of Nydus, which by default now uses the strict Redis implementation https://github.com/disqus/nydus/blob/master/nydus/db/backends/redis.py#L64 which caused at a certain point the swap of the argument in the API.

Nydus switched to enforce strict redis here: https://github.com/disqus/nydus/commit/98ee7b388e0310a9b2c2f0a9674a09e6a0e8e925

Instead by using the non-strict version, the API arguments have always stayed the same.

tiborsimko:
We pin Nydus to `nydus==0.10.6` in `requirements.txt`. 

Anyway, I'm closing this PR, because the current version works perfectly well.

Issue #3600, opened 2016-02-27
websubmit : Send_APP_Mail error with record_owners


chokribr:
I've got this error by using Send_APP_Mail
Unhandled TypeError caught when calling [Send_APP_Mail] WebSubmit function: [an integer is required]
a Warning is displayed for user 
A serious function-error has been encountered. Adminstrators have been alerted.
Please not that this might be due to wrong characters inserted into the form (e.g. by copy and pasting some text from a PDF file). 
https://github.com/inveniosoftware/invenio/blob/maint-1.2/modules/websubmit/lib/functions/Send_APP_Mail.py#L244

chokribr:
#3605

PR #3605, opened 2016-03-02
websubmit: Avoid error with Send_APP_Mail

chokribr:


tiborsimko:
See #3600

tiborsimko:
As discussed in the chat room betwenn @chokribr and @ludmilamarian:

a quick fix can be: 

record_owners = print_record(int(sysno), 'tm', \
[CFG_WEBSUBMIT_RECORD_OWNER_EMAIL]).strip()
so int(sysno) instead of sysno
the fix would be in the Send_APP_Mail.py

actually this Send_APP_Mail function doesn't really need to call the print_record (this is some very old code). It can just as simply run get_fieldvalues(sysno, CFG_WEBSUBMIT_RECORD_OWNER_EMAIL) to retrieve the same info

tiborsimko:
@chokribr Would you have an updated PR?

modules/websearch/lib/search_engine.py
```diff
@@ -5017,7 +5017,7 @@ def print_record(recID, format='hb', ot='', ln=CFG_SITE_LANG, decompress=zlib.de
         display_claim_this_paper = False
 
     can_edit_record = False
-    if check_user_can_edit_record(user_info, recID):
+    if  not (format.lower().startswith('t')) and check_user_can_edit_record(user_info, recID):
```
chokribr: if we are seeking for the content of only one field from metadata, no need to check whether user can or not edit record

chokribr:
ok done #3758, I'll sent another PR to explain  search_engine modifications

PR #3608, opened 2016-03-09
Font Awesome support

jochenklein:
Enables the support for using Font Awesome, the iconic font and CSS toolkit.


jirikuncar:
@ludmilamarian @egabancho @switowski it looks more like a feature. Shall we close it since Invenio 1.x is in feature freeze?

egabancho:
It is related with the new author profile pages, there are some icons used there, but I think there is no PR with the changes on the `bfe` elements ... so ...

PR #3613, opened 2016-04-01
websearch: advanced search and exact match

remileduc:
Actually, when we do an advanced search, it doesn't really do an exact search even if we select the _"exact match"_. In fact, if we have parentheses with several words in it, it will try to parse it as a logical statement, like `(foo OR bar)`

Thus, if we have an author with a strange name like `organization (foo, bar)` and we do an exact search on it, we won't found any records. While doing the same search surrounded by quotes to force the exact search will work.

here, I propose to add a config variable to configure the advanced search. So, if `CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH` is `True`, it will use the default search function and not try to parse the parentheses.

### DIFF
--- a/config/invenio.conf
+++ b/config/invenio.conf
@@ -412,12 +412,12 @@ CFG_WEBSTYLE_REVERSE_PROXY_IPS =
 # often during the runtime.  (Note that you may modify them
 # afterwards too, though.)
 
-# CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH -- tells if the advanced
-# search should do an exact match or use the parentheses behavior (adds
-# the possibility to do boolean operation in parentheses).
-# Note that the boolean operations can always be done thanks to the form
-# in the UI.
-CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH = False
+# CFG_WEBSEARCH_ADVANCED_SEARCH_PREFER_MATCHING_TYPE -- tells if the
+# advanced search engine should prefer the matching type selected by
+# the user rather than the Boolean query typed by the user.
+# Note that the boolean operations can always be done thanks to the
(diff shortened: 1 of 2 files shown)


modules/websearch/lib/search_engine.py
```diff
@@ -57,6 +57,7 @@
      CFG_SCOAP3_SITE, \
      CFG_OAI_ID_FIELD, \
      CFG_WEBCOMMENT_ALLOW_REVIEWS, \
+     CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH, \
```
tiborsimko: Please make a try/except when importing this new configuration variable, otherwise we might break some existing 1.2.x instances.

config/invenio.conf
```diff
@@ -412,6 +412,13 @@ CFG_WEBSTYLE_REVERSE_PROXY_IPS =
 # often during the runtime.  (Note that you may modify them
 # afterwards too, though.)
 
+# CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH -- tells if the advanced
+# search should do an exact match or use the parentheses behavior (adds
+# the possibility to do boolean operation in parentheses).
+# Note that the boolean operations can always be done thanks to the form
+# in the UI.
+CFG_WEBSEARCH_ADVANCED_SEARCH_EXACT_MATCH = False
```
tiborsimko: If the new variable works also for the other types of queries (say regexp), it may be better to name it more generally, such as `CFG_WEBSEARCH_ADVANCED_SEARCH_PREFER_MATCHING_TYPE`.

(This is basically what the variable does: it instructs the search engine to prefer the matching type selected by the user rather than the Boolean query typed by the user.)

Issue #3620, opened 2016-05-31
installation: remove googlecode hosting of JS dependencies


tiborsimko:
The installation instructions of old legacy branches use googlecode dead links:

``` console
$ git grep googlecode
Makefile.am:    wget -N http://jquery-multifile-plugin.googlecode.com/svn-history/r54/trunk/jquery.MultiFile.js && \
Makefile.am:    wget -r -np -nH --cut-dirs=4 -A "png,css" -P jquery-ui/themes http://jquery-ui.googlecode.com/svn/tags/1.8.17/themes/base/ && \
Makefile.am:    wget -r -np -nH --cut-dirs=4 -A "png,css" -P jquery-ui/themes http://jquery-ui.googlecode.com/svn/tags/1.8.17/themes/smoothness/ && \
Makefile.am:    wget -r -np -nH --cut-dirs=4 -A "png,css" -P jquery-ui/themes http://jquery-ui.googlecode.com/svn/tags/1.8.17/themes/redmond/ && \
Makefile.am:    wget -N http://jquery-ui.googlecode.com/svn/tags/1.8.17/themes/redmond/jquery-ui.css && \
Makefile.am:    wget -N http://jquery-ui.googlecode.com/svn/tags/1.8.17/demos/images/calendar.gif && \
Makefile.am:    wget -r -np -nH --cut-dirs=5 -A "png" http://jquery-ui.googlecode.com/svn/tags/1.8.17/themes/redmond/images/)
requirements.txt:https://py-editdist.googlecode.com/files/py-editdist-0.3.tar.gz
```

The prerequisites should be hosted on inveniosoftware.org.

tiborsimko:
Fully closed now.

```console
$ git grep googlecode | wc -l
0
```

PR #3621, opened 2016-05-31
installation: pinned JSON-js Medialeement versions

### LINKED ISSUE #11: use specific jQuery versions

tiborsimko:
* BETTER Pinned specific JSON-js and Mediaelement versions instead of
  using the latest master branch commits. (closes #11)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -34,10 +34,6 @@ MATHJAX = http://invenio-software.org/download/mathjax/MathJax-v$(MJV).zip
 CKV = 3.6.6
 CKEDITOR = ckeditor_$(CKV).zip
 
-# current MediaElement.js version
-MEV = master
-MEDIAELEMENT = http://github.com/johndyer/mediaelement/zipball/$(MEV)
-
 #for solrutils
 INVENIO_JAVA_PATH = org/invenio_software/solr
 solrdirname = apache-solr-3.1.0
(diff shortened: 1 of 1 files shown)


tiborsimko:
@egabancho see also #3620 that is coming...

tiborsimko:
@egabancho Yes, that's coming soon via another PR, starting from the `maint-1.0` branch...

PR #3629, opened 2016-06-03
WebSearch: fix for da today 3650 test case

tiborsimko:
* FIX Fixes a search regression test case that looks for records added
  in the past 10 years. This test started to fail because some records
  are declared with date added (see MARC tag 005) before 2006. Amends
  the test case to use safer value of last 100 years instead of last 10
  years.


egabancho:
We should add a note to remember modifying this test in 100 years 😜 

:shipit:

PR #3630, opened 2016-06-03
installation: Apache and Redis start-up

tiborsimko:
* Makes sure that Apache and Redis are started right after provisioning.
  Fixes a problem on Travis CI when not using their built-in Redis
  service.

### DIFF
--- a/scripts/provision-web.sh
+++ b/scripts/provision-web.sh
@@ -104,6 +104,12 @@ provision_web_ubuntu_precise () {
           unzip \
           vim
 
+    # make sure Apache is started:
+    sudo /usr/sbin/service apache2 restart
+
+    # make sure Redis is started:
+    sudo /usr/sbin/service redis-server restart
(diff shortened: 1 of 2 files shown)


egabancho:
:shipit:

scripts/provision-web.sh
```diff
@@ -182,12 +188,15 @@ provision_web_centos6 () {
...
     sudo /sbin/chkconfig httpd on
 
-    # start Redis:
-    sudo /etc/init.d/redis start
-
     # enable Redis upon reboot:
     sudo /sbin/chkconfig redis on
 
+    # make sure Apache is started:
+    sudo /etc/init.d/httpd restart
```
egabancho: why don't `/usr/sbin/service` instead?

tiborsimko: Yeah, coming in #3631...

PR #3632, opened 2016-06-03
installation: support for Ubuntu 14.04 LTS

tiborsimko:
* BETTER Installation scripts now support Ubuntu 14.04 LTS.
    
### DIFF
--- a/scripts/populate-instance.sh
+++ b/scripts/populate-instance.sh
@@ -62,12 +62,20 @@ apache_wsgi_restart () {
     $sudo -u "${INVENIO_WEB_USER}" touch "${INVENIO_WEB_DSTDIR}/var/www-wsgi/invenio.wsgi"
 }
 
-start_apache_ubuntu_precise () {
+start_apache_ubuntu12 () {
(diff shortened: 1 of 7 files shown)


scripts/populate-instance.sh
```diff
@@ -62,14 +62,22 @@ apache_wsgi_restart () {
...
+start_apache_ubuntu12 () {
     $sudo /usr/sbin/service apache2 start
 }
 
-stop_apache_ubuntu_precise () {
+stop_apache_ubuntu12 () {
     $sudo /usr/sbin/service apache2 stop
 }
 
+start_apache_ubuntu14 () {
```
egabancho: If both are the same, why not `[start|stop]_apache_ubuntu`, IMHO it might make readability of the script a bit easier

egabancho:
LGTM

scripts/populate-instance.sh
tiborsimko: Thought of upcoming `systemd` in Ubuntu 16.04 LTS.  Probably could be refactored to use `..._service`, `..._systemctl` names...

PR #3640, opened 2016-06-21
Translations de

aw-bib:
Fix some fuzzy translations.

Rewrite the message for searches performed (de/en). As it seems to depend on external tools how many searches are stored on the system the static text `for the last 30 days or so` was considered missleading.

Addresses join2/join2-data#1296

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -505,9 +505,8 @@ msgstr "Autor"
 
 #: modules/webhelp/web/help-central.webdoc:98
 #: modules/websearch/doc/search-guide.webdoc:20
-#, fuzzy
 msgid "Search Guide"
-msgstr "Siehe Handbuch"
+msgstr "Handbuch"
 
 #: modules/websearch/doc/search-guide.webdoc:347
 #: modules/websearch/doc/search-guide.webdoc:382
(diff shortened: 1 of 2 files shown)


po/en.po
```diff
@@ -7445,7 +7445,7 @@ msgstr ""
 msgid ""
 "You have performed %(x_nb1)s searches (%(x_nb2)s different questions) during "
 "the last 30 days or so."
-msgstr ""
+msgstr "You have performed %(x_nb1)s searches (%(x_nb2)s different questions)."
```
tiborsimko: This should be better fixed by amending the original English source code string, so that other I18N translations won't get out of sync.

po/de.po
```diff
@@ -8660,7 +8599,7 @@ msgstr "Dieser Korb ist öffentlich zugänglich unter folgender Adresse:"
 #: modules/webbasket/lib/webbasket_templates.py:2488
 #, fuzzy
 msgid "This basket does not contain any records yet."
-msgstr "Diese Sammlung enthält zur Zeit keine Dokumente."
+msgstr "Dieser Korb enthält zur Zeit keine Dokumente."
```
tiborsimko: If you review and fix message translations, then the `fuzzy` directive has to be removed, otherwise the translations wouldn't be taken into account.

Instead of amending this in this PR, we have moved to Transifex also for the Invenio legacy translations, and there have been quite a few updates since then. Can you please look there and amend the translations there directly?

- https://www.transifex.com/inveniosoftware/invenio/invenio-maint-11-messages/

Many thanks and sorry for not integrating this PR before moving to Transifex.

aw-bib: I fear I get a 403 forbidden.

Issue #3661, opened 2016-07-13
Maint-1.2: BibUpload - possibility to upload multi DOIs


rthieledesy:
Hi,

with the new version main-1.2 it is no more possible to upload a record with a known DOI. In our case, we have in some cases double DOIs and a hanging bibsched is not a good solution for us.

I would suggest a pull request with a CFG variable to switch off this default behavior.

This is related to #1298

kaplun:
@aw-bib was mentioning me this very same need few days ago 👍

aw-bib:
@kaplun this is the fix for this exact issue, right.

PR #3662, opened 2016-07-13
Bibuplaod: add CFG parameter for  DOI checking

rthieledesy:
Add new CFG_BIBUPLOAD_DOI_CHECKING in invenio.conf and bibupload.py
for avoiding the multi DOI checking while the upload. The default
state is not changed. Please configure the new parameter in the
invenio-local.conf.

Addresses inveniosoftware/invenio#3661

### DIFF
--- a/config/invenio.conf
+++ b/config/invenio.conf
@@ -1942,6 +1942,10 @@ CFG_BATCHUPLOADER_WEB_ROBOT_RIGHTS = {
 # By default all DOIs are excluded
 CFG_BIBUPLOAD_INTERNAL_DOI_PATTERN = [^\w\W]
 
+# CFG_BIBPLOAD_CHECK_DOI -- do we want to check for 
+# DOI uniqueness during upload jobs? (True/False)
+CFG_BIBUPLOAD_DOI_CHECKING = True
+
 ####################################
 # Part 18: BibCatalog parameters ##
 ####################################
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -1212,27 +1212,33 @@ def check_record_doi_is_unique(rec_id, record):
     Return (boolean, msg) where 'boolean' would be True if the DOI is
     unique.
     """
-    record_dois = record_extract_dois(record)
-    if record_dois:
-        matching_recids = set()
-        for record_doi in record_dois:
-            possible_recid = find_record_from_doi(record_doi)
-            if possible_recid:
-                matching_recids.add(possible_recid)
-        if len(matching_recids) > 1:
-            # Oops, this record refers to DOI existing in multiple records.
-            msg = "   Failed: Multiple records found in the" \
+    try:
+        from invenio.config import CFG_BIBUPLOAD_DOI_CHECKING
+    except:
+        CFG_BIBUPLOAD_DOI_CHECKING = True
+
+    if CFG_BIBUPLOAD_DOI_CHECKING:
+        record_dois = record_extract_dois(record)
+        if record_dois:
+            matching_recids = set()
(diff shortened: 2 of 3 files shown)


config/invenio.conf
```diff
@@ -1942,6 +1942,12 @@ CFG_BATCHUPLOADER_WEB_ROBOT_RIGHTS = {
 # By default all DOIs are excluded
 CFG_BIBUPLOAD_INTERNAL_DOI_PATTERN = [^\w\W]
 
+# CFG_BIBUPLOAD_DOI_CHECKING - check multi DOIs
+# Parameter for switch on / off DOI checking in bibuploads
+# CFG_BIBUPLOAD_DOI_CHECKING = 0: multi DOIs are possible
+# CFG_BIBUPLOAD_DOI_CHECKING = 1: multi DOIs are forbidden (default)
+CFG_BIBUPLOAD_DOI_CHECKING = 1
```
egabancho: What about `True/False` instead `1/0`?
Like in https://github.com/inveniosoftware/invenio/blob/legacy/config/invenio.conf#L673

modules/bibupload/lib/bibupload.py
```diff
@@ -1212,27 +1212,33 @@ def check_record_doi_is_unique(rec_id, record):
...
-        matching_recids = set()
-        for record_doi in record_dois:
-            possible_recid = find_record_from_doi(record_doi)
-            if possible_recid:
-                matching_recids.add(possible_recid)
-        if len(matching_recids) > 1:
-            # Oops, this record refers to DOI existing in multiple records.
-            msg = "   Failed: Multiple records found in the" \
+    try:
+        from invenio.config import CFG_BIBUPLOAD_DOI_CHECKING
```
egabancho: I would move the import together with the others, https://github.com/inveniosoftware/invenio/blob/legacy/modules/bibupload/lib/bibupload.py#L41, and probably you don't need the `try: ... except: ...` block as it's going to have a default value from `invenio.conf`

rthieledesy:
True / False is fully ok.

Should I make the changes or can you put this in the code?

modules/bibupload/lib/bibupload.py
tiborsimko: A try/except technique for imports is necessary in `maint-x.y` branches where we promised not to break API/DB/config.  We use techniques like this:

- https://github.com/inveniosoftware/invenio/blob/maint-1.2/modules/bibdocfile/lib/bibdocfile_config.py#L20-L31

I fully agree the import can be up in the file though.

config/invenio.conf
tiborsimko: Moreover, the explanation for values can come simply in the comment, so that the variable name is not necessarily repeated. For example:

```
# CFG_BIBPLOAD_CHECK_DOI -- do we want to check for 
# DOI uniqueness during upload jobs? (True/False)
CFG_BIBUPLOAD_CHECK_DOI = True
```

tiborsimko:
> Should I make the changes or can you put this in the code?

It would be better that you do.  Please also add yourself to the `AUTHORS` file.  You can squash your commits together, amend commit log message to [follow our practices](http://invenio.readthedocs.io/en/latest/technology/git.html#remarks-often-made-during-code-review), and re-push with `--force`.

```
BibUpload: configurable DOI checking

* BETTER Introduces new configuration variable.... 
  ... (closes #3661)
```

modules/bibupload/lib/bibupload.py
egabancho: My brain immediately thought about the `legacy` branch when I saw the code, indeed the try/except is needed ☺️

rthieledesy:
I hope, the result is ok, now.

PR #3664, opened 2016-07-15
scripts: bash fixes

david-caro:
- Fix bad -v usage
- Add some missing indents
- Use realpath instead of cd + pwd
- Use cat instead of multiple echo
- Remove unneeded which call
- Use bash builtin [[]]

Signed-off-by: David Caro dcaroest@redhat.com

### DIFF
--- a/scripts/create-instance.sh
+++ b/scripts/create-instance.sh
@@ -23,77 +23,77 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 # check environment variables:
-if [ -v ${INVENIO_WEB_HOST} ]; then
+if [[ -v INVENIO_WEB_HOST ]]; then
     echo "[ERROR] Please set environment variable INVENIO_WEB_HOST before runnning this script."
     echo "[ERROR] Example: export INVENIO_WEB_HOST=192.168.50.10"
     exit 1
 fi
-if [ -v ${INVENIO_WEB_INSTANCE} ]; then
+if [[ -v INVENIO_WEB_INSTANCE ]]; then
     echo "[ERROR] Please set environment variable INVENIO_WEB_INSTANCE before runnning this script."
     echo "[ERROR] Example: export INVENIO_WEB_INSTANCE=invenio3"
     exit 1
 fi
-if [ -v ${INVENIO_WEB_VENV} ]; then
+if [[ -v INVENIO_WEB_VENV ]]; then
     echo "[ERROR] Please set environment variable INVENIO_WEB_VENV before runnning this script."
     echo "[ERROR] Example: export INVENIO_WEB_VENV=invenio3"
     exit 1
 fi
-if [ -v ${INVENIO_USER_EMAIL} ]; then
+if [[ -v INVENIO_USER_EMAIL ]]; then
     echo "[ERROR] Please set environment variable INVENIO_USER_EMAIL before runnning this script."
     echo "[ERROR] Example: export INVENIO_USER_EMAIL=info@inveniosoftware.org"
     exit 1
 fi
-if [ -v ${INVENIO_USER_PASS} ]; then
+if [[ -v INVENIO_USER_PASS ]]; then
     echo "[ERROR] Please set environment variable INVENIO_USER_PASS before runnning this script."
     echo "[ERROR] Example: export INVENIO_USER_PASS=uspass123"
     exit 1
 fi
-if [ -v ${INVENIO_POSTGRESQL_HOST} ]; then
+if [[ -v INVENIO_POSTGRESQL_HOST ]]; then
     echo "[ERROR] Please set environment variable INVENIO_POSTGRESQL_HOST before runnning this script."
     echo "[ERROR] Example: export INVENIO_POSTGRESQL_HOST=192.168.50.11"
     exit 1
 fi
-if [ -v ${INVENIO_POSTGRESQL_DBNAME} ]; then
+if [[ -v INVENIO_POSTGRESQL_DBNAME ]]; then
     echo "[ERROR] Please set environment variable INVENIO_POSTGRESQL_DBNAME before runnning this script."
     echo "[ERROR] Example: INVENIO_POSTGRESQL_DBNAME=invenio3"
     exit 1
 fi
-if [ -v ${INVENIO_POSTGRESQL_DBUSER} ]; then
+if [[ -v INVENIO_POSTGRESQL_DBUSER ]]; then
     echo "[ERROR] Please set environment variable INVENIO_POSTGRESQL_DBUSER before runnning this script."
     echo "[ERROR] Example: INVENIO_POSTGRESQL_DBUSER=invenio3"
     exit 1
 fi
-if [ -v ${INVENIO_POSTGRESQL_DBPASS} ]; then
+if [[ -v INVENIO_POSTGRESQL_DBPASS ]]; then
     echo "[ERROR] Please set environment variable INVENIO_POSTGRESQL_DBPASS before runnning this script."
     echo "[ERROR] Example: INVENIO_POSTGRESQL_DBPASS=dbpass123"
     exit 1
 fi
-if [ -v ${INVENIO_REDIS_HOST} ]; then
+if [[ -v INVENIO_REDIS_HOST ]]; then
     echo "[ERROR] Please set environment variable INVENIO_REDIS_HOST before runnning this script."
     echo "[ERROR] Example: export INVENIO_REDIS_HOST=192.168.50.12"
     exit 1
 fi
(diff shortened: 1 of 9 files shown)


david-caro:
For some reason kwalitee complains about scripts not being a recognized component name, but I see that there are other commits with that component on it, let me know if I should use anything else

jirikuncar:
@david-caro the component name is fine. We are missing `.kwalitee.yml` in the repo.

jirikuncar:
@david-caro you are still missing dots at the end of sentences.

david-caro:
Should I open a but to kwalitee about it? (it's not complaining)

jirikuncar:
@david-caro are you still using your redhat email as signature?

david-caro:
yep, haven't find the time yet to change it

scripts/create-instance.sh
```diff
@@ -23,77 +23,77 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 # check environment variables:
-if [ -v ${INVENIO_WEB_HOST} ]; then
+if [[ -v INVENIO_WEB_HOST ]]; then
```
tiborsimko: `[[` is nice, but non-POSIX.  (If we ever have to target non-bash scenarios.) An alternative fix would be to use quotes that we use elsewhere, e.g. https://github.com/inveniosoftware/invenio/blob/maint-1.2/scripts/create-instance.sh#L29

scripts/create-instance.sh
```diff
@@ -107,15 +107,15 @@ set -o errexit
...
+    # sphinxdoc-install-invenio-full-begin
+    # FIXME the next pip commands are needed only for invenio<3.0.0a3
+    pip install invenio-db[postgresql] --pre
+    pip install invenio-access[postgresql] --pre
+    pip install invenio-search --pre
+    pip install dojson --pre
+    # now we can install full Invenio:
+    pip install invenio[full] --pre
+    # sphinxdoc-install-invenio-full-end
 else
```
tiborsimko: The indentation was intentionally left like this, so that the generated Sphinx documentation would  look good. Have you checked it out?

scripts/create-instance.sh
david-caro: in the shebang it's already enforcing bash, so non-POSIX already.

scripts/create-instance.sh
david-caro: nop, will do, though I'd prefer readable code without docs, than docs with unreadable code ;)

jirikuncar:
> yep, haven't find the time yet to change it

I think it's time so `kwalitee check authors` knows you under one email.

david-caro:
The pgp signature has multiple aliases, so it should be ok for cern or personal email too

scripts/create-instance.sh
jirikuncar: @david-caro the code was done primarily for "testable" docs 😉

scripts/create-instance.sh
tiborsimko: That's why I wrote "if we ever...". Personally I'm OK with requiring bash.

tiborsimko:
> The pgp signature has multiple aliases, so it should be ok for cern or personal email too

Nope, we just need to create a proper `.kwalitee.yml` for this repository, with wanted component names.  Would you like to do it as part of this PR? For example, look here: 

- https://github.com/inveniosoftware/kwalitee/blob/master/.kwalitee.yml

Otherwise I can do it later...

> The pgp signature has multiple aliases, so it should be ok for cern or personal email too

(BTW one can put alternative email aliases to `.kwalitee.yml` file so that they wouldn't be reported.)

> 7c98d10 scripts: bash fixes

Technically speaking, these are mostly improvements to working scripts,, so you could call the commit log headline  "scripts: bash improvements".

tiborsimko:
One more thing: can you add yourself to `AUTHORS.rst`?

scripts/create-instance.sh
```diff
@@ -23,77 +23,77 @@
...
-if [ -v ${INVENIO_WORKER_HOST} ]; then
+if [[ -v INVENIO_WORKER_HOST ]]; then
     echo "[ERROR] Please set environment variable INVENIO_WORKER_HOST before runnning this script."
     echo "[ERROR] Example: export INVENIO_WORKER_HOST=192.168.50.15"
     exit 1
 fi
 
 # load virtualenvrapper:
-source $(which virtualenvwrapper.sh)
+source virtualenvwrapper.sh
```
tiborsimko: Regarding `source $(which virtualenvwrapper.sh)` and removing the `which` part, IIRC this was necessary in the previous Invenio 1/2 kickstart scripts depending on the presence of the system package, the venv package, the possible `--system-site-packages` option, and whether the venv was activated or not before running those `source` lines (in the same shell session).  

It was safer to use `which` as part of the command for some of these combinations. Here it may be safer.  Just wanted to check whether you tested both Vagrant and Docker installation techniques to see whether it can now be safely removed? 

In case of doubt, it's safer to leave the `which` bit there.

P.S. We have some more occurrences of `which` in the other scripts.

david-caro: Will try it out, but which should not find anything that's not in the $PATH, and source already does that path crawling, but yes, better be on the safe side

scripts/create-instance.sh
david-caro: How do I run the docs tests?

david-caro: @tiborsimko it does affect the docs, but not very noticeably, it just adds the indent to the code in the docs. There's a lot of code that already does that, and imo the improvement in the code is way bigger than the nuisance in the docs (but really, for me it's not even a nuisance)

scripts/create-instance.sh
david-caro: I'm having issues trying to run the virtualbox setup on fc24 (it requires an older kernel, that breaks everything else, and the images used don't support libvirt). Will try to figure out a way to test it...

david-caro:
> Technically speaking, these are mostly improvements to working scripts,, so you could call the commit log headline "scripts: bash improvements".

It turns out that all the '[ -v $variable ]' are actually errors, as what it's doing is checking if the variable that has the name of the contents of $variable is set. At least on every system I've tried so far.

tiborsimko:
>  It turns out that all the '[ -v $variable ]' are actually errors, as what it's doing is checking if the variable that has the name of the contents of $variable is set. 

Ah, indeed!  `$` shouldn't be there...

> I'm having issues trying to run the virtualbox setup on fc24 

I'll retest this locally then...

tiborsimko:
(1) OK, after a quick test, things don't work:

```console
$ vagrant up redis
[...]
==> redis: Running provisioner: shell...
    redis: Running: inline script
==> redis: [ERROR] Please set environment variable INVENIO_REDIS_HOST before runnning this script.
==> redis: [ERROR] Example: export INVENIO_REDIS_HOST=192.168.50.12
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

because you'd need the NOT operator when testing for the environment variable non-presence in places like:

```console
$ git log master.. -p -- scripts/provision-redis.sh
[...]
-if [ -v ${INVENIO_REDIS_HOST} ]; then
+if [[ -v INVENIO_REDIS_HOST ]]; then
```

(2) Additionally, this PR makes the Docker build not to work:

```console
$ docker-compose build --no-cache
[...]
Step 3 : RUN /tmp/provision-web.sh
 ---> Running in c64e9d4d25be
[ERROR] Sorry, unsupported distribution UNDETECTED-UNDETECTED.
```

because you removed the `.dockerinit`, `.dockerenv` detection bits.

(3) Finally, after you fix the above issues, can you please rebase against the latest master and add yourself to the AUTHORS file as [mentioned above](https://github.com/inveniosoftware/invenio/pull/3664#issuecomment-233002441)?

david-caro:
About `1`, it looks to me as if the bug was already there, and my fix just showed it (as it seems like it was a double-error that countered itself). So should I change the old logic?

About `2`, I'll investigate, as docker containers should have the lsb_tools installed, or the redhat-release file in any case. I don't agree that is correct to take for granted that just because it's a docker container, you are running on ubuntu.

tiborsimko:
Ad 1, yes, the old logic was bad due to [extra dollar](https://github.com/inveniosoftware/invenio/pull/3664#issuecomment-243111048), please change them everywhere, just adding `!` for every `-v` test should be enough.

Ad 2, the canonical Python docker images that we are using are built on Debian/Ubuntu... it's been like that since ages.  So I think it's relatively safe to rely on that for simplicity's sake. If Python docker image creators switch to another OS one day, we'll see it very soon anyway...

PR #3684, opened 2016-08-02
installation: fix of nginx host header

JavierDelgadoFernandez:
* Updates the ``Host`` header value in order to use ``$http_host`` variable which also contains the port that the browser is sending.

Signed-off-by: Javier Delgado <javier.delgado.fernandez@cern.ch>

### DIFF
--- a/nginx/invenio3.conf
+++ b/nginx/invenio3.conf
@@ -10,7 +10,7 @@ server {
 
     location / {
         proxy_pass http://web:5000;
-        proxy_set_header Host $host;
+        proxy_set_header Host $http_host;
         proxy_set_header X-Real-IP $remote_addr;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }


JavierDelgadoFernandez:
The difference between ``$host`` and ``$http_host`` can be found here: http://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx#answer-15414811

jirikuncar:
@tiborsimko @lnielsen @egabancho @jmartinm did you experience this issue too?

lnielsen:
> did you experience this issue too?

No, but I also only have Zenodo running on default ports (80/443) in both docker and production.

jirikuncar:
@JavierDelgadoFernandez please re-format commit message since line 3 is too long (133 > 72).

JavierDelgadoFernandez:
Sorry. It is done ;).

tiborsimko:
Ditto as @lnielsen said, only default ports are used for the Invenio 3 demo site and for the COD caching layer, so we should be good.  (Another use case to double check would be the multiple domains pointing to the same node, but we don't really have that either.)

PR #3689, opened 2016-08-24
global: installation and docs fixes

alizeepace:


.travis.yml
```diff
@@ -40,9 +40,12 @@ env:
...
 python:
   - "2.7"
-  - "3.3"
-  - "3.4"
   - "3.5"
+  - "pypy"
+
+matrix:
+  allow_failures:
+    - python: pypy
```
tiborsimko: `pypy` should be there only after the Invenio 3 demo site  runs on pypy... Please delete here and in `setup.py` later, squash and repush.

alizeepace: Ok. Should I still remove version 3.3 and 3.4?

tiborsimko: Yes, that's OK.

Issue #3691, opened 2016-09-07
installation: support for Ubuntu 16.04 LTS


diegodelemos:
The current Invenio 3.0 installation instructions support Ubuntu 14.04 LTS and CentOS7. It would be good to add support for Ubuntu 16.04 LTS as well.

- [ ] change `scripts/*`
- [ ] use `ubuntu14` but add `ubuntu16` helper functions as necessary
- [ ] amend and check documentation
- [ ] check Vagrant way of installation
- [ ] check Docker way of installation

mpasternak:
Hi. I'm interested in this branch, I'm actually doing some tests. From what I saw, you haven't created the PR yet, any particular reasons for that, @diegodelemos ?

diegodelemos:
There is https://github.com/inveniosoftware/invenio/pull/3692 but it was abandoned.

BAA5809:
Hi, I am starting to work on a bigger repository and I am checking Invenio for it.

I have ubuntu16 and I was thinking maybe if I change the installation scripts and extend them to ubuntu16 it could just work? I did in fact only change the 14 to 16 and most of the things got installed. However I get errors for ngix.

Question is: How is the progress for this feature. Do you see major difficulties for ubuntu16) I am not an expert about the differences between distributions. Is it worth trying further or shall I simply install ubuntu14?
Subquestion: I have ubuntu14 on a 32bit machine, I guess this is not recommended as well, am I right?
Thanks!

mpasternak:
@BAA5809 , check out @diegodelemos branch https://github.com/diegodelemos/invenio/tree/3691-installation-on-ubuntu16

As for your subquestion, you described old, unsupported Ubuntu version on an old, obsolete CPU architecture.

PR #3692, opened 2016-09-09
installation: support for Ubuntu 16.04

diegodelemos:
Docker build and xenial64 test with Vagrant still to be done.

### DIFF
--- a/scripts/provision-elasticsearch.sh
+++ b/scripts/provision-elasticsearch.sh
@@ -44,7 +44,7 @@ check_environment_variables () {
 # quit on unbound symbols:
 set -o nounset
 
-provision_elasticsearch_ubuntu14 () {
+provision_elasticsearch_ubuntu () {
 
     # sphinxdoc-install-elasticsearch-ubuntu14-begin
     # install curl:
@@ -63,7 +63,7 @@ provision_elasticsearch_ubuntu14 () {
     # install Elasticsearch server:
     sudo DEBIAN_FRONTEND=noninteractive apt-get -y install \
          elasticsearch \
-         openjdk-7-jre
+         default-jre
 
     # allow network connections:
     if ! sudo grep -q "network.host: ${INVENIO_ELASTICSEARCH_HOST}" \
@@ -165,10 +165,8 @@ main () {
 
     # call appropriate provisioning functions:
     if [ "$os_distribution" = "Ubuntu" ]; then
-        if [ "$os_release" = "14" ]; then
-            check_environment_variables
-            provision_elasticsearch_ubuntu14
-            install_plugins
+        if [ "$os_release" = "14" -o "$os_release" = "16" ]; then
+            provision_elasticsearch_ubuntu
         else
             echo "[ERROR] Sorry, unsupported release ${os_release}."
             exit 1
--- a/scripts/provision-postgresql.sh
+++ b/scripts/provision-postgresql.sh
@@ -55,7 +55,7 @@ fi
 # quit on unbound symbols:
 set -o nounset
 
-provision_postgresql_ubuntu14 () {
+provision_postgresql_ubuntu () {
 
     # sphinxdoc-install-postgresql-ubuntu14-begin
     # update list of available packages:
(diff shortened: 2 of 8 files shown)


diegodelemos:
Docker has been run successfully.

There was a problem when launching `docker-compose build` on my Ubuntu 16.04. The image could not reach the Debian repositories.  In order to fix it I went through the following steps:
    1- You go to the file `/etc/systemd/system/multi-user.target.wants/docker.service` and the line:
    `ExecStart=/usr/bin/dockerd -H fd://`
    Needs to be changed to:
    `ExecStart=/usr/bin/dockerd -H fd:// --dns 137.138.16.5 --dns 137.138.17.5 --dns 8.8.8.8 --dns 8.8.4.4`
    2- Then the following commands need to be run:
    `sudo systemctl restart docker`
    `sudo systemctl daemon-reload`
    `sudo systemctl restart docker`

diegodelemos:
Docker check complete.  Found an error while launching, in order to fix it I went through the following steps (provided by @JavierDelgadoFernandez):

1. Edit the file `/etc/systemd/system/multi-user.target.wants/docker.service` changing the ExecStart command as follows:
    * `ExecStart=/usr/bin/dockerd -H fd:// --dns 137.138.16.5 --dns 137.138.17.5 --dns 8.8.8.8 --dns 8.8.4.4`
2. Then run the following comands:
    * `sudo systemctl restart docker`
    * `sudo systemctl daemon-reload`
    * `sudo systemctl restart docker`

scripts/provision-elasticsearch.sh
```diff
@@ -54,7 +55,7 @@ provision_elasticsearch_ubuntu14 () {
     # install Elasticsearch server:
     sudo DEBIAN_FRONTEND=noninteractive apt-get -y install \
          elasticsearch \
-         openjdk-7-jre
+         $openjdk_version
```
tiborsimko: This would be incomprehensible in the generated documentation... Can you please amend to have explicit version detection?

```
# detect Elasticsearch version
apt-cache search | grep | and| stuff
```

(or checkout `default-jre`?)

scripts/provision-postgresql.sh
```diff
@@ -65,25 +65,27 @@ provision_postgresql_ubuntu14 () {
     sudo DEBIAN_FRONTEND=noninteractive apt-get -y install \
          postgresql
 
+    postgres_version=$(psql --version | cut -d" " -f3)
```
tiborsimko: Please add  a header like:

```
# detect PostgreSQL version:
```

and let's name variable `postgresql_version`.

tiborsimko:
> There was a problem when launching docker-compose build on my Ubuntu 16.04. The image could not reach the Debian repositories. In order to fix it I went through the following steps

FWIW an alternative quick fix is to work around `dnsmasq`:

```console
$ sudo vim /etc/NetworkManager/NetworkManager.conf
$ sudo grep dnsmas /etc/NetworkManager/NetworkManager.conf
#dns=dnsmasq
$ sudo /etc/init.d/network-manager restart
```

diegodelemos:
I think it should be ok now, I will squash the three commits if everything is correct.

mpasternak:
Is it stated anywhere in the docs, that the ``invenio`` user has to be allowed to ``sudo``, at least for the time of installation? 

Also, even if I'm not doing ``provision-web.sh`` as root, I still get a magnitude of those:

```
gyp WARN EACCES user "root" does not have permission to access the dev dir "/usr/lib/node_modules/node-sass/.node-gyp/4.8.7"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/lib/node_modules/node-sass/.node-gyp"
gyp verb tmpdir == cwd automatically will remove dev files after to save disk space
gyp verb command install [ '4.8.7' ]
gyp verb install input version string "4.8.7"
gyp verb install installing version: 4.8.7
gyp verb install --ensure was passed, so won't reinstall if already installed
gyp verb install version not already installed, continuing with install 4.8.7
gyp verb ensuring nodedir is created /usr/lib/node_modules/node-sass/.node-gyp/4.8.7
gyp WARN EACCES user "root" does not have permission to access the dev dir "/usr/lib/node_modules/node-sass/.node-gyp/4.8.7"
gyp WARN EACCES attempting to reinstall using temporary dev dir "/usr/lib/node_modules/node-sass/.node-gyp"
gyp verb tmpdir == cwd automatically will remove dev files after to save disk space
```

That's some ``npm`` fault, described in another issue (https://github.com/nodejs/node-gyp/issues/454 - explanation is there: https://github.com/nodejs/node-gyp/issues/454#issuecomment-315691803 )

mpasternak:
Seems I've fixed the gyp issue, https://github.com/diegodelemos/invenio/pull/1

scripts/provision-rabbitmq.sh
```diff
@@ -37,6 +37,14 @@ provision_rabbitmq_ubuntu14 () {
     # install RabbitMQ server:
     sudo DEBIAN_FRONTEND=noninteractive apt-get -y install \
          rabbitmq-server
+
+    # allow remote connections for guest user
+    sudo sudo chmod 777 -R /etc/rabbitmq
```
jirikuncar: I am not sure if this is a good idea.

I would go with something like:
```bash
echo '[{rabbit, [{loopback_users, []}]}].' | sudo tee -a /etc/rabbitmq/rabbitmq.config
```

mpasternak: @jirikuncar except that one (good catch BTW), do you see any direct problems with this PR that would prevent you from merging it? 

as @diegodelemos wrote in his comment, https://github.com/inveniosoftware/invenio/issues/3691#issuecomment-355896393 , from what I understand, he considers this PR/branch to be abandoned. From my perspective, it just needs a few simple fixes and it could be committed, so Ubuntu 16.04 users could benefit from easy installation of Invenio. 

@jirikuncar , what are your thoughts on that?

jirikuncar: I don't see any other blocker.

PR #3721, opened 2016-11-18
bibupload: use CFG_OAI_ID_FIELD for deduping

### LINKED ISSUE #2812: Also use CFG_OAI_ID_FIELD for deduping

tiborsimko:
* FIX In case of replicating datasets between instances by means of OAI
  harvesting it is plausible to check for external OAI-IDs not only in the field
  specified by CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG but also in CFG_OAI_ID_FIELD.
  (closes #2812) (PR #2816)

* Note: find_record_from_oaiid() always returned the first record with a given
  OAI-ID assuming that it is unique. However, due to manual intervention one may
  have accidentially produced dupes here. Thus check, if more than one record
  with an OAI-ID exists. If so, check if all records except one are deleted. In
  this case return this surviving record, as cataloguers have resolved dupes
  manually. In case we still have dupes resort to the old behaviour for sake of
  compatibility, but at least throw a warning. This is somewhat of a TODO.
  Probably one should refuse merge here and wait for manual curation (to be
  discussed).

### DIFF
--- a/modules/bibupload/lib/bibupload.py
+++ b/modules/bibupload/lib/bibupload.py
@@ -650,6 +650,13 @@ def find_records_from_extoaiid(extoaiid, extoaisrc=None):
                     if extoaisrc is None:
                         write_message('WARNING: Found recid %s for extoaiid="%s" that specify a provenance (%s), while input record does not have a provenance.' % (id_bibrec, extoaiid, this_extoaisrc), stream=sys.stderr)
 
+        if len(ret) == 0:
+            # no oaiid in CFG_BIBUPLOAD_EXTERNAL_OAIID_TAG. Check if we have one
+            # in the locally used CFG_OAI_ID_FIELD that matches. This can happen
+            # if oai harvesting is used to duplicate records.
+            recid = find_record_from_oaiid(extoaiid)
+            if recid is not None:
+                ret.add(recid)
         return ret
     except Error, error:
         write_message("   Error during find_records_from_extoaiid(): %s "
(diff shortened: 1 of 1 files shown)


egabancho:
LGTM

PR #3724, opened 2016-11-22
Merge branch 'maint-1.1' into maint-1.2

tiborsimko:
@aw-bib Please note that your OAI matching fix for `maint-1.1` may need some adaptation in `maint-1.21 due to the presence of `CFG_BIBUPLOAD_MATCH_DELETED_RECORDS` that was absent in `maint-1.1`. Please check how I merged the feature and, if you have an Invenio 1.2 installation, please verify the behaviour on your concrete use case, as there are no integration tests for this feature...

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -6,6 +6,160 @@ releases.  For more information about the current release, please
 consult RELEASE-NOTES.  For more information about changes, please
 consult ChangeLog.
 
+Invenio v1.1.7 -- released 2016-11-20
+--------------------------------------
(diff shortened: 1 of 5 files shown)


egabancho:
LGTM

aw-bib:
@tiborsimko I have to admit, I do not understand what I should check for. Also I fear we did not manage to get 1.2 running properly in our setups so we stopped exploring that upgrade path.

PR #3516, opened 2015-09-18
crossrefutils: urlencode encoding

kaplun:
- Correctly urlencodes the query sent to CrossRef in order to
  retrieve information about a given DOI.
  (closes #3503)

Reported-by: Chokri Ben Romdhane chokribr@gmail.com
Signed-off-by: Samuele Kaplun samuele.kaplun@cern.ch

PR #3725, opened 2016-11-22
PR #2758 #2824 #3516

tiborsimko:
- 6f928c1ea221de82cc820dd6b73a2107d4e3f92d crossrefutils: urlencode encoding
- 96f4027153dbd4cfdddada1b5b4ca56738b342ff BibFormat: outdated_caches() fix
- 2b598ecbb369fde1a55ccd7e647bc1ddf1fee410 BibIndex: infinite loop in beautify_range_list

### DIFF
--- a/modules/bibformat/lib/bibreformat.py
+++ b/modules/bibformat/lib/bibreformat.py
@@ -270,8 +270,7 @@ def outdated_caches(fmt, last_updated, chunk_size=5000):
     sql = """SELECT br.id
              FROM bibrec AS br
              INNER JOIN bibfmt AS bf ON bf.id_bibrec = br.id
-             WHERE br.modification_date >= %s
-             AND bf.format = %s
+             WHERE bf.format = %s
              AND bf.last_updated < br.modification_date
              AND br.id BETWEEN %s AND %s"""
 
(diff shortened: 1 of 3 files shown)


modules/bibformat/lib/bibreformat.py
```diff
@@ -280,7 +279,7 @@ def outdated_caches(fmt, last_updated, chunk_size=5000):
     max_id = run_sql("SELECT max(id) FROM bibrec")[0][0] or 0
     for start in xrange(1, max_id + 1, chunk_size):
         end = start + chunk_size
-        recids += intbitset(run_sql(sql, (last_updated_str, fmt, start, end)))
```
egabancho: Since the `last_updated` is not used anymore, maybe it could be a nice idea to remove it from the parameters.

egabancho:
LGTM

modules/bibformat/lib/bibreformat.py
tiborsimko: Yes, indeed! Nobody uses `outdated_caches()` as an API... Perhaps we can amend this (and remove unnecessary date statements) in a next commit?

PR #3726, opened 2016-11-23
installation: plugins for Elasticsearch

jirikuncar:


scripts/provision-elasticsearch.sh
```diff
@@ -140,30 +148,36 @@ main () {
         os_distribution=$(cut -d ' ' -f 1 /etc/redhat-release)
         os_release=$(grep -oE '[0-9]+\.' /etc/redhat-release | cut -d. -f1 | head -1)
     else
-        os_distribution="UNDETECTED"
-        os_release="UNDETECTED"
+        os_distribution="${1:-UNDETECTED}"
```
tiborsimko: This is a bit weak, people can accidentally pass unwanted arguments.  Elsewhere, we test for Docker presence like:

```
   if [ -f /.dockerinit -o -f /.dockerenv ]; then
        # running inside Docker
```

It would be good to do the same philosophy everywhere.
```

File: docker-compose.yml [hunk 2]
```diff
@@ -70,7 +70,8 @@ redis:
 
 elasticsearch:
   restart: "always"
-  image: elasticsearch:2
+  build: .
+  dockerfile: Dockerfile.elasticsearch
```
tiborsimko: For nginx, we use `nginx/Dockerfile`. Perhaps we can harmonise this?

scripts/instance.cfg
```diff
@@ -0,0 +1,41 @@
+# Database
```
jirikuncar: @tiborsimko please let me know what do you think about this config template.

docs/installation/installation-detailed.rst
```diff
@@ -22,8 +22,7 @@ Detailed installation guide
 
 .. admonition:: CAVEAT LECTOR
 
-   Invenio v3.0 alpha is a bleeding-edge developer preview version that is
-   scheduled for public release in Q3/2016.
+   Invenio v3.0 alpha is a bleeding-edge developer preview version.
```
tiborsimko: If we are amending and/or removing Q3/206, then there are a few more occurrences:

```console
$ git grep 'Q3/2016'
docs/architecture/system-overview.rst:   scheduled for public release in Q3/2016.
docs/installation/installation-detailed.rst:   scheduled for public release in Q3/2016.
docs/introduction/releases.rst:set of standalone independent Python packages. *(ETA: Q3/2016)*
```

I'd rather replace Q3/2016 by Q1/2017, as it is good to leave some kind of public ETA indication... even if we have to postpone the date.

scripts/create-instance.sh
```diff
@@ -135,55 +130,8 @@ fi
...
-echo "${RECORDS_UI_CONF}" >> ../../var/${INVENIO_WEB_INSTANCE}-instance/${INVENIO_WEB_INSTANCE}.cfg
-JSONSCHEMAS_CONF=`cat <<EOF
-JSONSCHEMAS_ENDPOINT='/schema'
-JSONSCHEMAS_HOST='${INVENIO_WEB_HOST}'
-EOF
-`
-echo "${JSONSCHEMAS_CONF}" >> ../../var/${INVENIO_WEB_INSTANCE}-instance/${INVENIO_WEB_INSTANCE}.cfg
-echo "OAISERVER_RECORD_INDEX='marc21'" >> ../../var/${INVENIO_WEB_INSTANCE}-instance/${INVENIO_WEB_INSTANCE}.cfg
-echo "OAISERVER_ID_PREFIX='oai:${INVENIO_WEB_INSTANCE}:recid/'" >> ../../var/${INVENIO_WEB_INSTANCE}-instance/${INVENIO_WEB_INSTANCE}.cfg
+pip install jinja2-cli
```
tiborsimko: Wouldn't it be better to "hide" the necessity of explicit installing of `jinja2-cli` by treating it as a pre-requisite dependency?

jirikuncar: I would not hide it yet. I am experimenting also with [`anytemplate`](https://github.com/ssato/python-anytemplate/) as an alternative.

docs/installation/installation-detailed.rst
jirikuncar: I have removed the deadline from all places since we have missed all passed deadlines so far. Honestly, I don't think it is good idea to give ETA if we don't have clearly defined roadmap with allocated resources.

--
TODO list: https://github.com/issues?utf8=%E2%9C%93&q=org%3Ainveniosoftware+is%3Aopen+milestone%3AI3B-API-Docs

scripts/create-instance.sh
```diff
@@ -108,13 +108,8 @@ set -o nounset
 
 if [[ "$@" != *"--devel"* ]]; then
 # sphinxdoc-install-invenio-full-begin
-# FIXME the next pip commands are needed only for invenio<3.0.0a3
-pip install invenio-db[postgresql] --pre
```
tiborsimko: Just testing the updated branch, and got into:

```console
$ docker-compose run --rm web ./scripts/populate-instance.sh
[...]
  File "/home/invenio/.virtualenvs/invenio3/lib/python2.7/site-packages/flask_sqlalchemy/__init__.py", line 542, in get_engine
    self._sa.apply_driver_hacks(self._app, info, options)
  File "/home/invenio/.virtualenvs/invenio3/lib/python2.7/site-packages/invenio_db/shared.py", line 63, in apply_driver_hacks
    from psycopg2.extensions import adapt, register_adapter
ImportError: No module named psycopg2.extensions
```

Looks like the explicit `pip install invenio-db[postgresql] --pre` instruction steps may still be needed?

docs/installation/installation-detailed.rst
tiborsimko: OK. We do have a [roadmap](https://github.com/inveniosoftware/invenio/wiki/Invenio-3.0-Roadmap), but  since the work is taken up mostly by "volunteering", ETAs are hard indeed.

scripts/provision-elasticsearch.sh
```diff
@@ -147,23 +159,29 @@ main () {
     # call appropriate provisioning functions:
     if [ "$os_distribution" = "Ubuntu" ]; then
         if [ "$os_release" = "14" ]; then
+            check_environment_variables
             provision_elasticsearch_ubuntu14
+            sudo install_plugins
```
tiborsimko: This does not work because `install_plugins` is a shell function. The `sudo` must be used inside it.  
You can use the same technique as e.g. `scripts/provision-web.sh`:

```
# runs as root or needs sudo?
if [[ "$EUID" -ne 0 ]]; then
    sudo='sudo'
else
    sudo=''
fi

[...]

install_plugins () {
     $sudo /usr/share/elasticsearch/bin/plugin install -b mapper-attachments
}
```

jirikuncar: I had to adapt it a bit since we don't want `sudo` in Docker.

tiborsimko:
LGTM

Everything is OK on both Docker and Vagrant.

PR #3733, opened 2016-12-16
BibIndex: native fulltext index textification

### LINKED ISSUE #2576: BibIndex: New records not fulltext indexed with native indexer
The native indexer is not able to index records "on the fly". This is due to a missing call to extract_text 

https://github.com/inveniosoftware/invenio/blob/master/modules/bibindex/lib/tokenizers/BibIndexFulltextTokenizer.py#L123

### LINKED ISSUE #3690: bibdocfile :  function extract_text not defined referenced in get_text
By looking to get_text in bidocfile  code I think that we need to to add extract_text def
https://github.com/inveniosoftware/invenio/blob/maint-1.2/modules/bibdocfile/lib/bibdocfile.py#L1562

### LINKED ISSUE #3708: Bibindex: new record Index fulltext   is  working only   after  bibdocfile --textify  --recid
The Fulltext native indexer in the invenio 1.2 version is working only after invoking bibdocfile --textify  --recid  without the index does not working.

I think that this issue could be an answer for those ones:

https://github.com/inveniosoftware/invenio/issues/3690
https://github.com/inveniosoftware/invenio/issues/2576

chokribr:
* FIX Enables native fulltext indexing by explicitly extracting text when using the BibDoc
  interface. (closes #3708) (closes #3690), (closes #2576)

### DIFF
--- a/modules/bibindex/lib/tokenizers/BibIndexFulltextTokenizer.py
+++ b/modules/bibindex/lib/tokenizers/BibIndexFulltextTokenizer.py
@@ -121,6 +121,8 @@ class BibIndexFulltextTokenizer(BibIndexDefaultTokenizer):
                 else:
                     text = ""
                     if hasattr(bibdoc, "get_text"):
+                     #Textify Doc Before getting Text
+                        bibdoc.extract_text()
                         text = bibdoc.get_text()
                     return self.tokenize_for_words_default(text)
             else:


chokribr:
@kaplun @tiborsimko  did this commit OK?

chokribr:
Dear @lnielsen i think that this commit is also critical since the native fulltext index will not work without.

lnielsen:
@chokribr We are not doing any new releases of the v1.x branch, so even if the PR was integrated, it won't make it into a release. I would like that we could still be able to support 1.x for some time but we simply don't have the manpower to do that. Others could pick up v1 and continue the support, but I don't see anyone else in the community that's currently able or willing to do that either.  I thank you very much for the effort in reporting and fixing the issue, and we leave the PR around, in case other v1 users hit the same problem.

PR #3735, opened 2016-12-19
travis: catalog compilation removal

jirikuncar:


tiborsimko:
Shall we remove also all the catalog-related directves from `setup.cfg`?

PR #3736, opened 2016-12-19
installation: more dependency pinning

tiborsimko:
* FIX Adds more upper version limits for dependencies such as Celery, FS,
  Flask-Collect. This fixes recently observed Invenio 2.0 installation issues.


tiborsimko:
FYI it was necessary to also pin `bower` and friends; `uglifyjs` wasn't enough...

PR #3754, opened 2017-02-07
WIP: Fix multiline firerole handling in managedocfiles

sbujam:
According to the docs, fireroles may include multiple rules, one per line separated by '\n'.
For example: "firerole: deny until '2011-01-01'
    allow any"
(note the ^M or \n before 'allow any')

When revising bibdocs with such restrictions in Document File Manager, (even when one does not change the 'Restrictions' dropdown) the firerole is 'cropped' in the first \n character, changing the actual rule!
With this patch, newlines are 'escaped' before written to the bibdocactions.log, so that the firerole is properly parsed and reapplied when needed.


sbujam:
notes: 
- This is my first PR in the invenio repo, so please accept my apologies if you find the code/style/documentation not up to your standards. I will try to do my best to fix it, should you find any issues with it.
- I was a bit lazy and made the changes in-line in my forked repo in GitHub and then created the PR. Having said that, the code (with different comments) seems to be working in my production site.
- I have not filed an issue in advance, but on second thought I might do now for historical reasons

sbujam:
Closing and properly creating a new PR

PR #3757, opened 2017-02-08
BibDocFile: proper handling of multiline fireroles when revising bibdocs

sbujam:
According to the docs, fireroles may include multiple rules, one per line separated by '\n'.
For example: "firerole: deny until '2011-01-01'
allow any" (note the ^M or \n before 'allow any')

When revising bibdocs with such restrictions in Document File Manager, (even when one does not change the 'Restrictions' dropdown) the firerole is 'cropped' in the first \n character, changing the actual rule!
With this patch, newlines are 'escaped' before written to the bibdocactions.log, so that the firerole is properly parsed and reapplied when needed.

### DIFF
--- a/modules/bibdocfile/lib/bibdocfile_managedocfiles.py
+++ b/modules/bibdocfile/lib/bibdocfile_managedocfiles.py
@@ -1298,7 +1298,8 @@ def log_action(log_dir, action, bibdoc_name, file_path, rename,
 
     Newlines are also reserved, and are escaped from the input values
     (necessary for the 'comment' field, which is the only one allowing
-    newlines from the browser)
+    newlines from the browser and for the 'file_restriction' field, which
+    allows multiline rules)
 
     Each line starts with the time of the action in the following
     format: '2008-06-20 08:02:04 --> '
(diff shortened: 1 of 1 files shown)


sbujam:
notes:
- pylint produces several warnings, but this is because in general the code does not fully abide by the coding standards.
- The patch is based on the maint-1.2 branch (but could/should be applied to other legacy branches)
- This is my first PR in the invenio repo, so please accept my apologies if you find the code/style/documentation not up to your standards. I will try to do my best to fix it, should you find any issues with it. 
- I have not filed an issue in advance. If needed, I could do that for historical reasons

sbujam:
@ludmilamarian Done suggestions 1 and 2 [the easy ones :) ]

Creating a test is a bit more tricky (I could try, but I don't feel confident enough meddling in to that area)

Regarding your last point, I think it applies 'as-is' (with an offset) to maint-1.1. For maint-1.0 the patch must be applied to another file (modules/websubmit/lib/websubmit_managedocfiles.py) instead. 
What do you think is the best approach to deal with it?

PR #3780, opened 2017-03-16
developersguide: update of the tutorial

drjova:
Uses the 3-datamodel tutorial as datamodel, this way the users can see the created records

Tutorial preview: https://github.com/drjova/iugw2017/tree/align-with-data-model/5-develop/tutorial


drjova:
related with https://github.com/inveniosoftware/iugw2017/pull/17

Issue #3802, opened 2017-04-24
Add a docker config for developpers


remileduc:
What?
------

In order to ease the integration of newcomers, we should have a fully working docker ready to develop.

The idea is to copy what have been done in https://github.com/cernopendata/opendata.cern.ch/blob/master/docker-compose-dev.yml with a specific docker-compose for developpers, with the necessary documentation.

This docker should:
- [x] run a demo site very easily (so it will be very close to the current `docker-compose.yml`
- [x] run invenio in debug mode (auto refresh...)
- [x] mount a dev folder so we can `pip install -e` any modules we want inside the containers, and thus work on a given module
- [ ] integrate a celery worker
- [x] install the debug toolbar
- [ ] add files for records
    - [ ] integrate invenio-records-files
    - [ ] add some example files to the demo records

Any other suggestions?
see https://github.com/inveniosoftware/invenio/pull/3803

How?
-----

The idea is to have the following development folder structure:

```yml
- dev
    - invenio (a clone of this repo)
    - modules (folder mounted into the containers)
        - invenio-records
        - invenio-i-like-you
        - ...
    - zenodo
    - ...
```

We have a `dev` folder containing a clone of `invenio`, `zenodo`, or all invenio instances you are working on...

It also contains a specific `modules` folder (you will be able to change the name in the `docker-compose-dev.yml`) that will be mounted in the containers. Thus, you can put here whatever you want to be accessible inside the containers.

Generally, you will have the clones of the modules you are working on here (or symlinks to them, or git-new-workdir, or whatever you want)

Then, you can `pip install -Ue` inside your docker to work on the modules you want.

remileduc:
added the `add files for records` section

lnielsen:
Now integrated with the cookiecutter templates

PR #3803, opened 2017-04-24
installation: add docker setup for developpers

remileduc:
- New docker-compose-dev.yml for developpers
- see #3802 

The documentation (at the top of the file) can be copied in antother RST file, but I'm not sure which one... I can add it later

This PR adds 2 files:
- Dockerfile-dev
- docker-compose-dev.yml

These can be used to create a docker dev environment.
We now use the version 3 of the YML of docker-compose...

We use the file `requirements-extra.txt` to install additional modules useful for debugging. In the future, we can just install `invenio[extra]`, see the comment in the `.sh` file.

This is for demo, it is not intended to be merged as is. The goal is to have an easy dev environment. Don't hesitate to provide feedbacks!

### DIFF
--- a/scripts/create-instance.sh
+++ b/scripts/create-instance.sh
@@ -89,6 +89,9 @@ if [ "${INVENIO_WORKER_HOST}" = "" ]; then
     echo "[ERROR] Example: export INVENIO_WORKER_HOST=192.168.50.15"
     exit 1
 fi
+if [ -z "$DEBUG" ]; then
+    DEBUG=0
+fi
 
 # load virtualenvrapper:
 # shellcheck source=/dev/null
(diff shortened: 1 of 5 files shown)


scripts/create-instance.sh
```diff
@@ -106,12 +109,18 @@ cdvirtualenv
...
+#    pip install invenio[ils,extra]
+#fi
+# For the moment, we use an extra requirements file:
 # sphinxdoc-install-invenio-full-begin
 pip install invenio-app-ils[postgresql,elasticsearch2]
 # sphinxdoc-install-invenio-full-end
-else
-    pip install -r "$scriptpathname/../requirements-devel.txt"
+if [ "$FLASK_DEBUG" == 1 ]; then
+    pip install -r "$scriptpathname/../requirements-extra.txt"
```
remileduc: this doesn't work because ENV variable in docker-compose.yml are set for run time, and this script is called at build time...
if you have an idea to make it work for both dev and non-dev...

PR #3808, opened 2017-05-29
OAIRepository: get_earliest_datestamp fix

kaplun:
* FIX Fixes get_earliest_datestamp() definition to conform to OAI-PMH
  standard: "a UTCdatetime that is the guaranteed lower limit of all
  datestamps recording changes, modifications, or deletions in the
  repository".

### DIFF
--- a/modules/oairepository/lib/oai_repository_server.py
+++ b/modules/oairepository/lib/oai_repository_server.py
@@ -247,7 +247,7 @@ def get_modification_date(recid):
 def get_earliest_datestamp():
     """Get earliest datestamp in the database"""
     out = ""
-    res = run_sql("SELECT DATE_FORMAT(MIN(creation_date),'%Y-%m-%d %H:%i:%s') FROM bibrec", n=1)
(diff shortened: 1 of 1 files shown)


kaplun:
@ludmilamarian this should fix what you reported me.

But it actually depends anyway on what we have recorded on our services as modification_date. E.g. in INSPIRE we have artificially set an initial `modification_date` when records have been imported from SPIRES (preserving the SPIRES modification date, although SPIRES didn't have any OAI-PMH server).

ludmilamarian:
Thanks @kaplun, this looks good!

kaplun:
cc: @tiborsimko for merging

lnielsen:
I can see test fails because of problem fixed in #3807, but unfortunately tests in #3807 also fails for some other reason.

kaplun:
Yeah, this fix should not cause any additional test to fail because there is no test testing such functionality :)

kaplun:
Note that I can't merge it myself. So up to you guys and gals :dancer:

PR #3818, opened 2017-07-20
bibcirculation: set manual loan period

### LINKED ISSUE #3816: bibcirculation: Not possible to set a due date manually
There seem to be several issues in the set due date functionality, ie the manual overwrite of due dates by library staff.

1. From the patrons page:
patron
     - _select action_ / _Change due date_ causes a 500. This error results from an (at least now) invalid SQL statement in get_loan_infos(), where an unused table is referenced and thus leeds to a `None` result.
2. From the items page:
item
    - _select action_ / _Change due date_ causes a 500. This error results from a wrong parameter passed to the backend (barcode instead of loan_id).
3. Finally, once the above is done, changing the due date in `tmpl_change_due_date_step1()` has no effect, as the input field defined in [l.7936](https://github.com/inveniosoftware/invenio/blob/d4654c31b75c6445eeb6daf04cd8b8755ce70e5a/modules/bibcirculation/lib/bibcirculation_templates.py#L7936) has a wrong `name`. It has to be `due_date` not `period_from`.

The second and third bug can be solved in `bibcirculation_templates`, e.g. by introducing named parameters and the proper name respectively, when building the necessary HTML.

The first error seems to require intervention in `bibcirculation_dblayer.py`, [l.1916](https://github.com/inveniosoftware/invenio/blob/d4654c31b75c6445eeb6daf04cd8b8755ce70e5a/modules/bibcirculation/lib/bibcirculation_dblayer.py#L1916) where `, crcLOANREQUEST lr` requires removal. This fix will then solve a subsequent error that would appear in case two if parameters were passed on properly.

isIdenticalTo: join2/join2#935

aw-bib:
* Fix: get_loan_infos() referenced an unused table leading to an
  invalid SQL statement and subsequently an application error.

* Fix tmpl_change_due_date_step1() used wrong variable for passing the
  date to the backend effectively setting due dates to 0000-00-00
  regardless the users input.

* Fix tmpl_get_item_loans_details() to set proper values in table row
  build up by introducing named parameters.

* Fix tmpl_borrower_loans_details() to set proper values in table row
  build up by introducing named parameters.

Closes #3816

### DIFF
--- a/modules/bibcirculation/lib/bibcirculation_dblayer.py
+++ b/modules/bibcirculation/lib/bibcirculation_dblayer.py
@@ -1913,7 +1913,7 @@ def get_loan_infos(loan_id):
                              l.status,
                              it.loan_period,
                              it.status
-                        from crcLOAN l, crcITEM it, crcLOANREQUEST lr
+                        from crcLOAN l, crcITEM it
                         where l.barcode=it.barcode and
                               l.id=%s""",
                    (loan_id, ))
--- a/modules/bibcirculation/lib/bibcirculation_templates.py
+++ b/modules/bibcirculation/lib/bibcirculation_templates.py
@@ -3973,37 +3973,42 @@ class Template:
 
                 out += """
             <tr>
-              <td>%s</td>
-              <td>%s</td>
-              <td>%s</td>
-              <td>%s</td>
-              <td>%s</td>
-              <td>%s - %s</td>
-              <td>%s</td>
-              <td>%s</td>
-              <td>%s</td>
+              <td>%(title_link)s</td>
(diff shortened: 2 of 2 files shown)


aw-bib:
Could it be that Travis fails at a completely unrelated point? I get he complains about

```
mv: cannot stat `/tmp/mediaelement/johndyer-mediaelement-*/build/*': No such file or directory

make: *** [install-mediaelement] Error 1
```

Similarly, QuantifiedCode complains also about non-touched parts? (I did _not_ fix up either of the modules to current coding standards everywhere, but just the parts that were required to fix the issue.)

PR #3824, opened 2017-08-11
Added testing section with e2e test to the documentation.

ntarocco:


docs/developersguide/testing-practices.rst
```diff
@@ -0,0 +1,72 @@
+Testing practices
+=================
+
+It is a good practice to add test cases to each Invenio module. Tests are
+normally organized in:
+
+1. Unit functional testing: test cases run against the Flask app defined in the
+   module, ofter as example app. They can be found in ``tests`` or
```
switowski: often :)

docs/developersguide/testing-practices.rst
```diff
@@ -0,0 +1,72 @@
+Testing practices
+=================
+
+It is a good practice to add test cases to each Invenio module. Tests are
+normally organized in:
+
+1. Unit functional testing: test cases run against the Flask app defined in the
```
switowski: This sounds a bit confusing (but I missed the end of sprint so maybe it was discussed IRL), but what are `unit functional tests`? Shouldn't there be a separation between `unit` tests (where you test some functions from a module) and `functional` tests (where you test the whole module as a `black box`) ?

docs/developersguide/testing-practices.rst
```diff
@@ -0,0 +1,72 @@
...
+-----------------------------
+
+To run end-to-end tests, the requirements are:
+
+- Invenio installed, with modules that are covered by the tests
+- an Elastic Search running instance
+- a Redis running instance
+- a web browser driver
+
+The database is instead created (if not existing) as SQLite instance when tests
```
switowski: How about rephrasing it as: `SQLite database will be created (if it doesn't exist) when tests are launched.` ?

docs/developersguide/testing-practices.rst
```diff
@@ -0,0 +1,72 @@
...
+
+End-to-end functional testing
+-----------------------------
+
+To run end-to-end tests, the requirements are:
+
+- Invenio installed, with modules that are covered by the tests
+- an Elastic Search running instance
+- a Redis running instance
+- a web browser driver
```
switowski: When I was reading this bullet point, I started thinking, `how do I install the web driver` and I see it's explained 2 lines below, so how about we merge those two lines:
- a web browser driver (check `Selenium WebDriver`_ documentation to see how to install it)

docs/developersguide/testing-practices.rst
ntarocco: Good point. The thing is that, IMHO, the tests I see in the `unit` folder, are not 100% unit test, where you test one specific function/method input and outputs. What I see there, are functional tests where you run a Flask app and you test a feature, but without UI.
Am I wrong?

So, again IMHO, those are not unit tests.
Any opinion @switowski and @lnielsen?

switowski: I agree, depending on which module, tests are a bit of a mess (for example [dojson](https://github.com/inveniosoftware/dojson/blob/master/tests/test_utils.py)
has tests that looks like `unit tests`, but most of the modules usually have more like `functional tests`).
How about we say: `unit and functional` tests for the first bullet and `end-to-end` tests for the second bullet? I'm also perfectly fine with leaving this line as it is, if you think there will be no confusion about it :)

ntarocco: done, thank you for your comments!

PR #3845, opened 2017-12-02
install PostgreSQL 9.4

Atisom:
change PostgreSQL 9.3 to PostgreSQL 9.4


lnielsen:
Thanks for the PR. We are redoing how the Invenio package is working, which means that these scripts won't be there anymore. I'll hence close the PR, but we will address the issue with Postgres version in the docker images.

PR #3873, opened 2018-06-13
docs: document cookiecutter install + instance output

fenekku:
- Updated the copyright notice
- Added a section on installing cookiecutter outside of 
  the project's virtualenv since it is also used for invenio 
  modules and data records. It is a tool worth having 
  everywhere and will simplify documentation for the
  Module Tutorial.
- Displayed the output of the cookiecutter call so that
  various names are explicit to the reader.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,7 +1,7 @@
 ..
     This file is part of Invenio.
     Copyright (C) 2015-2018 CERN.
-    Copyright (C) 2018 Galter Health Sciences Library.
+    Copyright (C) 2018 Northwestern University, Feinberg School of Medicine, Galter Health Sciences Library.
 
     Invenio is free software; you can redistribute it and/or modify it
     under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -36,7 +36,19 @@ your instance.
...
-First, let's create a `virtualenv <https://virtualenv.pypa.io/en/stable/installation/>`_
+
+Before we begin, you want to make sure to have Cookiecutter installed. Invenio
+leverages this tool to generate the starting boilerplate for different
+components, so it will be useful to have in general.
+
+.. code-block:: shell
+
+  # Globally install or upgrade it
+  $ sudo pip install --upgrade cookiecutter
```
diegodelemos: I would only show the procedure to install it as user package or even install  it inside the virtual env, otherwise it looks good to me :)

fenekku: I removed the instructions to install globally and mentioned how `cookiecutter` can alternatively be installed in the virtualenv. :+1:

PR #3874, opened 2018-06-14
Module tutorials scaffold

fenekku:
This pull request continues where the last one left off. I am updating the Module Tutorial in this one with what I had to do to make the webpage interactions work. I am not too sure if this is what was intended or if the backend is consistent with how things should be. Correct me and don't hesitate to explain things!

### DIFF
--- a/docs/tutorial-module/develop.rst
+++ b/docs/tutorial-module/develop.rst
@@ -303,13 +303,16 @@ Demo time
 ---------
 
 Let's now see our Invenio module in action when integrated with our Invenio instance.
-We will install this module in our Invenio instance and then run the ``server``
-script. First go to the root folder of the ``invenio-unicorn`` module, then
+First we activate our instance's virtualenv and we install the new Invenio-Unicorn module:
 
 .. code-block:: console
 
     $ workon my-repository-venv
     (my-repository-venv)$ pip install --editable .[all]
+
+Then, we go to the instance folder, `my-repository`, and start the ``server``
+script:
+
     (my-repository-venv)$ cd ../my-repository
     (my-repository-venv)$ ./scripts/server
 
--- a/docs/tutorial-module/develop.rst
+++ b/docs/tutorial-module/develop.rst
@@ -12,11 +12,12 @@ The goal of this tutorial is to add data to Invenio v3. We'll create a
 form that inserts the data in the database. Also we will touch different
 part of the development process such as:
 
-- How to create a new view
 - How to create a form
+- How to create a new view
 - How to add a utility function
 - How to add new templates
 - How to use Jinja2
+- How to define your own JSON Schema
 
 Flask extensions
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


docs/tutorial-module/develop.rst
```diff
@@ -254,31 +247,73 @@ On the ``utils.py`` module will create a helper function that creates a record.
...
+    }
+
+
 Demo time
 ---------
 
-Let's start our server again.
+Let's now see our Invenio module in action when integrated with our Invenio instance.
+We will install this module in our Invenio instance and then run the ``server``
+script. First go to the root folder of the ``invenio-unicorn`` module, then
```
diegodelemos: Shall we split this into two parts to make it more explicit? What do you think about the following:

>First we activate our instance virtualenv and we install the new `Invenio-Unicorn` module:
>```console
>$ workon my-repository-venv
>(my-repository-venv)$ pip install --editable .[all]
>```
>Then, we go to the instance folder, `my-repository`, and start the server:
>```console
>(my-repository-venv)$ ./scripts/server
>```
You can rephrase but I think splitting it would make it more clear!

docs/tutorial-module/run-n-test.rst
```diff
@@ -17,16 +17,23 @@ Before that, we need to **stop** any running Invenio instance.
...
-Install the module is very easy, you just need to go to the root directory of
-the module and run the command:
+First, create a virtualenv for the module:
 
 .. code-block:: bash
 
-    pip install -e .[all]
+    $ mkvirtualenv my-module-venv
+
+Installing the module is very easy, you just need to go to the root directory
```
diegodelemos: Maybe a shorter version: `Installing the module is very easy, you just need to go to its root directory and run:` what do you think?

docs/tutorial-module/run-n-test.rst
```diff
@@ -17,16 +17,23 @@ Before that, we need to **stop** any running Invenio instance.
...
+
+.. code-block:: bash
+
+    (my-module-venv)$ cd invenio-unicorn/
+    (my-module-venv)$ pip install --editable .[all]
 
 Some explanations about the command:
 
-- the ``-e`` option is used for development. It means that if you change the
+- the ``--editable`` option is used for development. It means that if you change the
```
diegodelemos: 👍

docs/tutorial-module/run-n-test.rst
```diff
@@ -99,16 +106,27 @@ Build the documentation
...
-section above) and run:
+section above). Make sure you are back at the root directory
+of the module and run:
 
 .. code-block:: console
 
-    $ python setup.py build_sphinx
+    (my-module-venv)$ python setup.py build_sphinx
+
+Similar to how you generate the documentation for the Invenio instance, you
```
diegodelemos: I think that providing only one way of building docs manually (`python setup.py build_sphinx`) is enough since we use `setuptools` everywhere :)

fenekku: I *think* the `all_files` option was slowing me down at the time (regenerating everything), but mostly `python setup.py build_sphinx` was not documented as the way to generate the documentation in the `quickstart.rst`. I fixed it to just mention the `build_sphinx` approach in both the instance and the module.

docs/tutorial-module/run-n-test.rst
fenekku: Short and sweet, I like it.

docs/tutorial-module/develop.rst
fenekku: Yup, separating explanations in smaller chunks à-la Windows wizards works.

PR #3898, opened 2018-10-31
release: v3.0.1rc1

lnielsen:
@inveniosoftware/architects Please have a look.


docs/releases/v3.0.1.rst
```diff
@@ -0,0 +1,68 @@
...
+- **Invenio-App:** Fixed compatibility issue with latest release of
+  Flask-Talisman.
+
+- **Invenio-DB:** Fixed a compatibility issue with newly released
+  SQLAlchemy-Continiuum.
+
+- **Invenio-Search:** Fixed a compatibility issue with
+  elasticsearch-dsl>=6.2.0.
+
+- **Pytest-Invenio:** Fixed a problem which caused Invenio modules to be loaded
```
slint: Maybe grouping these might look a bit better and be easier to read? (e.g. like Django does on a per-"module" level: https://docs.djangoproject.com/en/2.1/releases/2.1/#minor-features)

lnielsen: Actually they keep it flat for patch releases: https://docs.djangoproject.com/en/2.1/releases/2.1.1/

My thinking was just that in most cases there's one or two bullets per module.

setup.py
```diff
@@ -40,18 +40,18 @@
...
         'invenio-userprofiles>=1.0.1,<1.1.0',
     ],
     'metadata': [
-        'invenio-indexer>=1.0.0,<1.1.0',
+        'invenio-indexer>=1.0.1,<1.1.0',
         'invenio-jsonschemas>=1.0.0,<1.1.0',
         'invenio-oaiserver>=1.0.0,<1.1.0',
         'invenio-pidstore>=1.0.0,<1.1.0',
-        'invenio-records-rest>=1.1.0,<1.3.0',
+        'invenio-records-rest>=1.1.1,<1.2.0',
```
lnielsen: Note, here I'm staying with ``<1.2.0`` and not ``<v1.3.0`` which was merged in master

egabancho: I think `1.1.1` was never released, travis was failing, what about `1.1.2` as minimum?

docs/releases/v3.0.1.rst
```diff
@@ -0,0 +1,68 @@
...
+  despite the related test fixtures not being used, thus causing import errors
+  if the module was not installed.
+
+- **Pytest-Invenio:** Fixed a compatibility issue with pytest v3.8.x.
```
fenekku: To be more precise, it was starting with pytest 3.8.1 (https://github.com/pytest-dev/pytest/issues/3941) (3.8.0 was fine) :nerd_face: 

```suggestion
- **Pytest-Invenio:** Fixed a compatibility issue with pytest v3.8.1+.
```

Issue #3883, opened 2018-07-23
docs: new section "Infrastructure architecture"


lnielsen:
- Overview
- Request handling: LB, Web server, App server, WSGi, UI vs APP
- Storing records: JSON and pids in database.
- Search and indexing: ES/MQ, direct/bulk indexing, usage stats
- Background processing: queue, low prio, indexing, cronjobs
- Caching and temporary storage: sessions, celery results, rendered pages
- Storing files: file systems
- Running with Docker: Kubernets, OpenShift,
- Running with machines: requirements

PR #3909, opened 2018-12-11
docs: docker

topless:
* Initial version of documentation for docker
* Added diagram picturing the build process
* Added docker commands to build containers
* Fixed lint errors in conf.py
* Addresses #3883


slint:
I have to agree with @ppanero, it's not clear that the `Dockerfile.base` has to be built + tagged before `Dockerfile`. I wrote some stuff about it on https://github.com/pypa/pipenv/issues/3285#issuecomment-445464634

lnielsen:
Closing PR because it's been stale for a long time.

PR #3992, opened 2020-02-11
dependencies: pin werkzeug (broken imports)

ppanero:
Pin werkzeug to greater than 0.15 and lower than 1 to fix the broken imports temporarily.


lnielsen:
Closing in favour of just fast-forward merging maint-3.2 as there's no difference on master and maint-3.2.

Issue #3995, opened 2020-03-01
docs: remove references to use docker in production


ntarocco:
Rephrase the paragraphs that suggest to check the docker-compose.full.yml for production and remove the `production` word to make sure that it is clear we don't suggest to use docker in production.

E.g. here:
- [docker-compose.full.yml](https://invenio.readthedocs.io/en/latest/tutorials/build-a-repository.html#docker-and-docker-compose)
- bootcamp sentences

PR #3996, opened 2020-03-03
docs: remove production suggestion

ppanero:
``` console
$ git grep production
docs/community/code-of-conduct.rst:      is to find them early, before they make it into production.
docs/deployment/securing-your-instance.rst:  production)
docs/general/bundles.rst:    This bundle is in beta. The modules are being used in production systems
docs/general/bundles.rst:    This bundle is in alpha. The modules are being used in production systems
docs/general/bundles.rst:    These modules are in alpha. The modules are being used in production
docs/general/bundles.rst:which are all being used in production systems, but which are likely subject
docs/migrating.rst:v2.1 production environment, the one we recommend is using `Virtualenv
docs/releases/v0.rst:Invenio v0.x code base was developed and used in production instances
docs/releases/v1.rst:Invenio v1.x code base is suitable for stable production. It uses legacy
docs/releases/v2.rst:    The 2.x code base is not suitable for production systems, and will not
docs/releases/v3.0.0.rst:already been relied upon in large-scale production systems for more than 1.5
docs/releases/v3.1.0.rst:usable in production environments like OpenShift.
docs/tutorials/bootcamp.rst:Invenio in production
docs/tutorials/build-a-module.rst:  production environment, this option shouldn't be used.
docs/tutorials/build-a-repository.rst:      e.g. a production or testing environment will remove all existing data.
docs/tutorials/build-a-repository.rst:  when you deploy your instance on a production environment, you can reproduce
docs/tutorials/build-a-repository.rst:  This file contains a full-fledged definition of a production-like application
docs/tutorials/build-a-repository.rst:      with actual nodes/machines or configuring a production-level container
```
The only one that needed change was:

``docs/tutorials/build-a-repository.rst:  when you deploy your instance on a production environment, you can reproduce``

Requires https://github.com/inveniosoftware/training/pull/61

Closes: https://github.com/inveniosoftware/invenio/issues/3995

Didn't find any other reference to "production" that seemed misleading in https://github.com/inveniosoftware/training

### DIFF
--- a/docs/tutorials/build-a-repository.rst
+++ b/docs/tutorials/build-a-repository.rst
@@ -167,7 +167,7 @@ configuring the following files:
 
 **docker-compose.full.yml**
 
-  This file contains a full-fledged definition of a production-like application
+  This file contains a full-fledged definition of a more scalable application
   infrastructure. It has all of the ``docker-compose.yml`` file's containers
   defined, and additionally:
 
@@ -188,14 +188,9 @@ configuring the following files:
   described below.
 
   .. warning::
-
-      While one might be tempted to deploy this as a fully functional Invenio
-      instance, it is not meant to be a turn-key solution, since it hasn't been
-      tested for this purpose. This is rather meant to be an inspiration (in
-      terms of configuration, networking and general principles) for
-      configuring your own setup either by replacing the container services
-      with actual nodes/machines or configuring a production-level container
-      orchestration system like Kubernetes, OpenShift, etc.
+      This file is **not** intended to be used for production, neither as
+      reference for a production infrastructure. It is just an example of a
+      more complete application deployment.
 
 **Dockerfile.base**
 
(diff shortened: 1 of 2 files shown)


docs/tutorials/build-a-repository.rst
```diff
@@ -167,7 +167,7 @@ configuring the following files:
 
 **docker-compose.full.yml**
 
-  This file contains a full-fledged definition of a production-like application
+  This file contains a full-fledged definition of a more scalable application
```
ntarocco: WDYT if we are more explicit?

```
This file contains a full-fledged definition of an application infrastructure. It has all of the ``docker-compose.yml`` file's containers defined, and additionally:
...

.. note::  This file is **not** intended to be used for production, neither as reference for a production infrastructure. It is just an example of a more complete application deployment.

```

ntarocco: @lnielsen do you have better words?

lnielsen: Agree with this comment.

ppanero: Done. Could you recheck? :)

PR #4000, opened 2020-03-19
contributing: review guidelines

ppanero:
Review guidelines as used by RDM team members.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -167,3 +167,47 @@ To speed up the integration process, it helps if on GitHub you `allow
 maintainers to edit your pull request
 <https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/>`_
 so they can fix small issues autonomously.
+
+.. rubric: Review guidelines
+
+A pull request will probably be reviewed by someone in a different location,
+maybe even a different time zone. Therefore, avoiding blockings and
+misunderstandings is key. The following ideas could help ease and speed
+up the review process:
+
+- Be clear when something is ready to be reviewed (see `WIP` above).
+- If a PR requires changes, say so explicitly, but politely (use emojis!).
+- When something is ready to be merged say it explicitly. If the reviewer does
+  not have permission to merge, (s)he should ping the corresponding person.
+- When reviewing, rate your comments. This would help to understand if
+  a comment must be addressed (e.g. it has substantial implications) or it is
+  something the reviewer thought of but is a matter of "take it or leave it". A
+  proposed scale is:
+  1. Comment/Doubt/Question: exactly that. A doubt, a question or a comment.
+  2. Minor: a change that the reviewer thinks might need change. However, it
+  is not blocking, it is up to the developer to choose if and how to change
+  it. It can be merged!
+  3. Moderate/Normal (Default): a change that requires further discussion
+  (e.g. breaking changes). It cannot be merged, unless explicitly stated by
+  the reviewer (e.g. choose a solution proposed by the reviewer and implement
+  it). Depending on the nature of the change a new review
+  might be needed, use common sense.
+  4. Major: a change that needs further discussion, probably a chat. Even the
+  opinion of an architect. It has high implications. It cannot be merged.
+  5. Shelved: a suggestion that will be treated later on as part of a
+  different issue. It is a good practice to reference the issue. Note that
+  any of the previous (comment, minor, moderate and major) can be shelved if
+  agreed by the PR creator and the reviewer.
+
+Here is an example:
+
+  minor: I think this variable name should be plural since it refers to a list.
+
+If no level is given, assume it is of "moderate" level within reason (might
+clearly be minor or major). GitHub suggestions for instance are typically
+minor, so no need to tag them all the time.
+
+Following these guidelines builds trust over time. As a reviewer, you know
+your comments will be heard and as a submitter, you know someone has taken the
+time to go over your work and help you solve the problem you were tackling. In
+both cases, it's done in a friendly and efficient manner.
--- a/pytest.ini
+++ b/pytest.ini
@@ -8,3 +8,4 @@
 
 [pytest]
 addopts = --pep8 --ignore=docs --cov=invenio --cov-report=term-missing
+filterwarnings = ignore::pytest.PytestDeprecationWarning


CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+
+A pull request will probably be reviewed by someone in a different location,
+maybe even a different time zone. Therefore, avoiding blockings and
+misunderstandings is key. The following ideas could help ease and speed
+up the review process:
+
+- Be clear when something is ready to reviewed (see `WIP` above).
+- If a PR requires changes, say so explicitly, but politely (use emojis!).
+- When something is ready to be merge say it explicitly. If the reviewer does
+not have permission to merge, it should ping the corresponding person.
```
fenekku:
```suggestion
not have permission to merge, they should ping the corresponding person.
```
:wink: 
This is the new gender neutral pronoun... which is actually quite interesting in French because no pronouns like it existed and it had to be invented in the last few years! ( "Ile" I think is the recommended one in French)

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+- When reviewing, rate your comments. This would help the other person know If
+a comment must be addressed (e.g. it has substantial implications) or it is
+something the reviewer thought of but is a matter of "take it or leave it". A
+proposed scale is:
+  1. Comment/Shelved: Something worth mentioning, a doubt or question (comment).
+  If it will be treated later on as part of a different issue it is "shelved".
```
fenekku:
```suggestion
  If it will be treated later or as part of a different issue, it is "shelved".
```

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+  In case of shelving reference the issue.
+  2. Minor: Something that the reviewer thinks might need change. However, it
+  is not blocking, it is up to the developer to choose if and how to change
+  it. It can be merged!
+  3. Moderate: Something that requires further discussion (e.g. breaking
```
fenekku:
```suggestion
  3. Moderate/Normal (Default): Something that requires further discussion (e.g. breaking
```

The way I've approached it is to assume this is the level if no level or other indication is given (from the comments I read and the ones I give). I will add a little section about this below.

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+  changes). It cannot be merged, unless explicitly said (e.g. choose a
+  solution proposed by the reviewer and implement it).
+  4. Major: Needs discussion, probably a video call. Even the opinion of an
+  architect. It has high implications. It cannot be merged.
```
fenekku:
```suggestion
  architect. It has high implications. It cannot be merged.
  
Here is an example:
  
    minor: I think this variable name should be plural since it refers to a list.
  
If no level is given, assume it is of "moderate" level within reason (might clearly be minor or major). GitHub suggestions for instance are typically minor, so no need to tag them all the time.

Following these guidelines builds trust over time. As a reviewer, you know your comments will be heard and as a submitter, you know someone has taken the time to go over your work and help you solve the problem you were tackling. In both cases, it's done in a friendly and efficient manner.  
```

CONTRIBUTING.rst
ppanero: Oh nice, so for future reference, is "they" not "it" even if the intention is for a user (singular, not plural)

CONTRIBUTING.rst
ppanero: I actually meant:

- It will be treated later
- Or as part of a different issue.

But is not a good practice to do things without an issue. So the correction applies :) hehe

CONTRIBUTING.rst
ppanero: Oh! I was actually assuming minor 😅 so yeah, better to put a default 👌

CONTRIBUTING.rst
fenekku: Yes. In any case, "it" is for a singular *thing* in English, so we would have used "he or she" in the past (why the winking face :smile:  ), but now "they" is more appropriate... which is also interesting because using "they" like here used to be a grammatical mistake. "They" was solely considered plural, so using it to refer to a singular person was erroneous (but it was common slang, so it was a typical way to lose points in English classes!).

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
 so they can fix small issues autonomously.
+
+.. rubric: Review guidelines
+
+A pull request will probably be reviewed by someone in a different location,
+maybe even a different time zone. Therefore, avoiding blockings and
+misunderstandings is key. The following ideas could help ease and speed
+up the review process:
+
+- Be clear when something is ready to reviewed (see `WIP` above).
```
ntarocco:
```suggestion
- Be clear when something is ready to be reviewed (see `WIP` above).
```

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+- If a PR requires changes, say so explicitly, but politely (use emojis!).
+- When something is ready to be merge say it explicitly. If the reviewer does
```
ntarocco:
```suggestion
- When something is ready to be merged say it explicitly. If the reviewer does
```

CONTRIBUTING.rst
ntarocco: Minor: I would honestly go for `she/he` or `(s)he`, they sounds weird :)

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+not have permission to merge, it should ping the corresponding person.
+- When reviewing, rate your comments. This would help the other person know If
```
ntarocco:
```suggestion
- When reviewing, rate your comments. This would help to understand if
```

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+a comment must be addressed (e.g. it has substantial implications) or it is
+something the reviewer thought of but is a matter of "take it or leave it". A
+proposed scale is:
+  1. Comment/Shelved: Something worth mentioning, a doubt or question (comment).
+  If it will be treated later on as part of a different issue it is "shelved".
+  In case of shelving reference the issue.
```
ntarocco: Moderate: why not separating `comment` and `shelved` then? If I understand correctly, `comment` means potentially no action, but `shelved` means "fix it later and create an issue for that". If this is the case, I would split the 2 in different points because they require different actions.

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+  2. Minor: Something that the reviewer thinks might need change. However, it
+  is not blocking, it is up to the developer to choose if and how to change
+  it. It can be merged!
+  3. Moderate: Something that requires further discussion (e.g. breaking
+  changes). It cannot be merged, unless explicitly said (e.g. choose a
```
ntarocco:
```suggestion
  changes). It cannot be merged, unless explicitly stated by the reviewer (e.g. choose a
```
It is not clear to me said by who ;)

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+  solution proposed by the reviewer and implement it).
```
ntarocco: Moderate: I am not sure this point is clear. It looks like that the reviewer proposed several different solutions, the creator of the PR then chooses one and then merge? After having implemented the chosen solution, I guess another review is required. Maybe I did not understand correctly.

CONTRIBUTING.rst
```diff
@@ -167,3 +167,30 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+  4. Major: Needs discussion, probably a video call. Even the opinion of an
```
ntarocco:
```suggestion
  4. Major: Needs more discussions, probably a chat. Even the opinion of an
```
or similar :)

Minor: what do you mean with the sentence "Even the opinion of an architect"? That you should involve one?

CONTRIBUTING.rst
fenekku: The "choose a solution" is an example. Commonly, "moderate" means don't merge right away, engage with the review comment and see where that leads: sometimes a change is needed, sometimes the reviewer missed something and no changes is needed. In any case, a back and forth will occur, but the back and forth is usually short. For example:

- reviewer: I thought we wanted to require this field?
- submitter: `<checks up on the question>` Ah you're right! Let me fix that. <pushes changes> `<alerts reviewer>`
- reviewer: `<takes another look>` Thanks, all good! `<approves if no other outstanding issues>`

CONTRIBUTING.rst
ppanero: True, I have separated it and set in `5.`. As it was was actually not accurate, since we are shelving also other types of comments when agreed (being more flexible to progress faster).

CONTRIBUTING.rst
ppanero: Exactly. Maybe this is a "loose end" we might have done sometimes. There are changes that are trivial (e.g. add `requires=True` to a marshmallow schema) but it has an impact so the comment is "moderate". I guess in those cases what we have done is:
1- Comment the `moderate` change
2- State that it can be merged
3- PR creator does the change (trivial) and can merge

Ofc, in this case we are assuming PR creator/reviewer are both maintainers... Maybe we shouldnt?

CONTRIBUTING.rst
ppanero: Indeed, e.g. a change in `pidstore` base code or `invenio-records-rest` views.

ppanero:
Thanks @fenekku and @ntarocco. I have addressed/answered most of the comments.

> Comment: I would define the naming and stick to it, e.g. the "PR creator" (or submitter), the "reviewer(s)", the "module maintainer" and replace e.g. "the corresponding person." with the "module maintainer" and any other occurrence to improve clarity.

I like the idea. Will tackle it a bit later :)

CONTRIBUTING.rst
fenekku: minor: So... This can be a fraught topic and I am not necessarily the best champion for that cause... but "they" is now a [grammatically correct](https://owl.purdue.edu/owl/general_writing/grammar/pronouns/gendered_pronouns_and_singular_they.html) way to denote both "he or she" *and* non-binary subjects. Agreed, it does sound strange to use "they" when you are used to being prevented from saying it :)! If members of our community identify as such, it is not a big burden on us to use that pronoun to have that extra layer of inclusion. For me, it's just easier to use "they" than alternatives ;).

In any case, we have been approaching this in good faith and no matter the decision the choice was not meant to exclude. You can merge when you want.

CONTRIBUTING.rst
```diff
@@ -167,3 +167,46 @@ To speed up the integration process, it helps if on GitHub you `allow
...
+
+- Be clear when something is ready to be reviewed (see `WIP` above).
+- If a PR requires changes, say so explicitly, but politely (use emojis!).
+- When something is ready to be merged say it explicitly. If the reviewer does
+not have permission to merge, (s)he should ping the corresponding person.
+- When reviewing, rate your comments. This would help to understand if
+a comment must be addressed (e.g. it has substantial implications) or it is
+something the reviewer thought of but is a matter of "take it or leave it". A
+proposed scale is:
+  1. Comment: a change worth mentioning, a doubt or question (comment).
```
Glignos, line 186: IMHO the categories 'Comment' and 'Minor' are currently very similar and could prove slightly confusing to pick between them.
The reasoning behind this would be that: the resulting action is the same since in both cases a change is proposed that is not necessary to be applied for the PR to be merged.
The solution for this could be to remove the option of suggesting a change through a Comment.
```suggestion
  1. Comment:  a doubt or question (comment).
```
In this case, this would mean that actually this is a 'Question' category as it serves solely to provide more information to the reviewer.
So then maybe it could look like:
```suggestion
  1. Question: a doubt or question (comment).
```

ntarocco:
We should really merge this: I was looking for this guide in the doc and I couldn't find it :)
My proposal: let's try to merge it having everyone more or less happy with the content, we can iterate on this later on eventually! :)

ppanero:
Finished fixing the content according to comments. Ready to merge on my side 🚀

run-tests.sh
```diff
@@ -8,7 +8,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 pydocstyle invenio && \
-isort -rc -c -df **/*.py && \
+isort -c --df **/*.py && \
```
fenekku, line 11:
```suggestion
isort --check-only --diff **/*.py && \
```

ppanero:
@inveniosoftware/architects I have rebased on latest master. If this is good to go (has 3 approval, 2 of architects), can we merge it?

ntarocco:
> @inveniosoftware/architects I have rebased on latest master. If this is good to go (has 3 approval, 2 of architects), can we merge it?

There is an error:
```
Warning, treated as error:

CONTRIBUTING.rst:181: WARNING: Bullet list ends without a blank line; unexpected unindent.
```
I can fix it in the afternoon if you don't do it before, then we merge it...

PR #4074, opened 2020-12-10
docs: Updates system setup

### LINKED ISSUE #4060: docs: write a system setup guide for linux
Write a system setup guide for Linux (see https://invenio.readthedocs.io/en/master/getting-started/development-environment.html#linux).Use the macOS example on the same page as a template.

I would think we would likely need to split the guide into 
- Ubuntu/Debian
- CentOS

Create a VM with the system (CentOS or Ubuntu/Debian) to test with so that we don't forget important things for newcomers.

jrcastro2:
* closes #4060

### DIFF
--- a/docs/getting-started/development-environment.rst
+++ b/docs/getting-started/development-environment.rst
@@ -175,10 +175,199 @@ See https://docs.docker.com/docker-for-mac/#resources.
 A typical sign of needed more resources, is that services are not running or
 images are having problems building.
 
-Linux
-~~~~~
+Ubuntu
+~~~~~~
 
-Want to write it? Contact us on the chat!
+System setup guide for Ubuntu.
+
+**Discord**
+
+The easiest way to install Discord is by using snap.
+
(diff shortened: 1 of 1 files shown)


docs/getting-started/development-environment.rst
```diff
@@ -58,6 +64,22 @@ using the command:
 
     $ brew install <packages>
 
+If you are on linux you can use the native package manager instead of Homebrew.
```
lnielsen, line 67: Question: Is the package names not different between brew, yum and apt?

lnielsen (commented):
Minor: The intention was to have separate guides for macOS/Linux so that the guides are super easy to follow. Also, the Linux guide should include getting the developer tools like Discord, VS Code...

jrcastro2:
Updated: Created a guide for Ubuntu

ines-cruz:
closing in favor of #https://github.com/inveniosoftware/invenio/pull/4084

